PR #31, opened 2020-12-02
Replace all dots in name

tenshiak:
Right now it replaces just the first one

### DIFF
--- a/src/jackett.js
+++ b/src/jackett.js
@@ -16,7 +16,7 @@ function reformatTitleForSearching(name) {
 		: movieMatch
 		? movieMatch[0]
 		: name;
-	return fullMatch.replace(".", " ");
+	return fullMatch.replace(/\./g, " ");
 }
 
 function fullJackettUrl(jackettServerUrl, params) {


src/jackett.js
```diff
@@ -16,7 +16,7 @@ function reformatTitleForSearching(name) {
 		: movieMatch
 		? movieMatch[0]
 		: name;
-	return fullMatch.replace(".", " ");
+	return fullMatch.replace(/\./g, " ");
```
mmgoodnow, line 19: Good catch! Have you tested this? I'm not sure the `replace` method will do that. 
```suggestion
	return fullMatch.replaceAll(".", " ");
```

tenshiak: I tested my version, yes. Not sure about replaceAll, I'm not a JS dev. Looks like replaceAll is a prototype function.

mmgoodnow: They're both prototype functions :) If it works, I'm satisfied.

mmgoodnow (approved):

mmgoodnow:
I'll merge and release this later today. Thanks for the contribution!

PR #32, opened 2020-12-03
Allow torrents with less files

tenshiak:
Currently cross-seed will allow only identical files, down to the byte, but very often sites reupload the same torrents without nfo file, or without sample. My changes allow this, because these torrents still match the original. The change checks if every file of new torrent exists in the original torrent. I tested it with 200 entries, found few matches that cross-seed didn't catch before, zero false positives.

### DIFF
--- a/src/decide.js
+++ b/src/decide.js
@@ -5,23 +5,18 @@ const logger = require("./logger");
 const { partial } = require("./utils");
 
 function compareFileTrees(a, b) {
-	if (a.length !== b.length) return false;
-	const sorter = (m, n) => (m.path < n.path ? -1 : m.path > n.path ? 1 : 0);
-	const sortedA = a.slice().sort(sorter);
-	const sortedB = b.slice().sort(sorter);
-
 	const cmp = (elOfA, elOfB) => {
 		const pathsAreEqual = elOfB.path === elOfA.path;
 		const lengthsAreEqual = elOfB.length === elOfA.length;
 		return pathsAreEqual && lengthsAreEqual;
 	};
-	return sortedA.every((elOfA, i) => cmp(elOfA, sortedB[i]));
+	return a.every(elOfA => b.some(elOfB => cmp(elOfA, elOfB) ));
 }
 
 function sizeDoesMatch(result, ogInfo) {
 	const { length } = ogInfo;
-	const lowerBound = length - 0.01 * length;
-	const upperBound = length + 0.01 * length;
+	const lowerBound = length - 0.02 * length;
+	const upperBound = length + 0.02 * length;
 	return result.Size >= lowerBound && result.Size <= upperBound;
 }
 
@@ -52,10 +47,6 @@ async function assessResultHelper(result, ogInfo, hashesToExclude) {
 	// if you got rate limited or some other failure
 	if (!info) return info;
 
-	if (info.length !== ogInfo.length) {
-		logReason("it has a different size");
-		return null;
-	}
 	if (hashesToExclude.includes(info.infoHash)) {
 		logReason("the info hash matches a torrent you already have");
 		return null;


src/decide.js
```diff
@@ -52,10 +47,6 @@ async function assessResultHelper(result, ogInfo, hashesToExclude) {
 	// if you got rate limited or some other failure
 	if (!info) return info;
 
-	if (info.length !== ogInfo.length) {
-		logReason("it has a different size");
-		return null;
-	}
```
mmgoodnow, line 58: Could you explain why we don't need this part anymore? One thing I want to watch out for here is re-subtitled files where they took the file and added subtitles for a certain language (say, Romanian). The info.length property also only applies to single-file torrents so I'm not sure that this would affect your target in this PR.

src/decide.js
```diff
@@ -5,23 +5,18 @@ const logger = require("./logger");
...
 
 function compareFileTrees(a, b) {
-	if (a.length !== b.length) return false;
-	const sorter = (m, n) => (m.path < n.path ? -1 : m.path > n.path ? 1 : 0);
-	const sortedA = a.slice().sort(sorter);
-	const sortedB = b.slice().sort(sorter);
-
 	const cmp = (elOfA, elOfB) => {
 		const pathsAreEqual = elOfB.path === elOfA.path;
 		const lengthsAreEqual = elOfB.length === elOfA.length;
```
mmgoodnow, line 10: I'm wondering now if we should do the same fuzzy checking on these file lengths. I realize it was like that before this PR, but since you're in the mood to weed out all the false negatives, this might be one of the sources of false negatives. Could you figure out if the `length` properties on individual files in multi-file torrents vary with piece size / align to piece size boundaries? If so then maybe we'll want to do something like a sizeDoesMatch call on them.

mmgoodnow (commented):
Thanks again for the contribution! I really appreciate it :)

src/decide.js
tkozybski: It's checking the whole torrent content size, so if the new torrent has less files, it would never match.

tenshiak: Better check the size per-file in the compareFileTrees. This one is redundant I think. It's checking the whole torrent content size, so if the new torrent has less files, it would never match. Goes against this PR.

src/decide.js
tenshiak: I wouldn't touch the file size (length) checking, because It would fail to detect any tampering with the file. Do you mean path lengths? Perhaps it could be taken out, but I worry comparing just the file sizes is not enough. I don't know about the piece size you ask for, but when I did cross-seed manually the chunk size didn't matter for the end result so I wouldn't rely on that (i had a torrent with 1MB chunk and 16MB chunk and it cross-seeded just fine).

mmgoodnow: Good point, let's keep that as-is then.

src/decide.js
mmgoodnow: Here's the relevant portion of the BT spec:

> There is also a key length or a key files, but not both or neither. If length is present then the download represents a single file, otherwise it represents a set of files which go in a directory structure.

> In the single file case, length maps to the length of the file in bytes.

> For the purposes of the other keys, the multi-file case is treated as only having a single file by concatenating the files in the order they appear in the files list. The files list is the value files maps to, and is a list of dictionaries containing the following keys:

> length - The length of the file, in bytes.

> path - A list of UTF-8 encoded strings corresponding to subdirectory names, the last of which is the actual file name (a zero length list is an error case).

> In the single file case, the name key is the name of a file, in the muliple file case, it's the name of a directory.

(From http://www.bittorrent.org/beps/bep_0003.html)

length will only be set on single-file torrents so we will still want to check it, I think.

mmgoodnow:
I'm curious, how many extra matches did it find?

src/decide.js
tenshiak: I checked that on multi-file torrent, it had length set to sum of size of every file. I think length from parse-torrent is not the same as length from the protocol. It's just size of the torrent.

tenshiak:
> 
> 
> I'm curious, how many extra matches did it find?

In terms of percents, I'd say 10 to 15? It depends on what you seed, It has best effects on single episodes and I don't have a lot of those.

src/decide.js
mmgoodnow: Ah yes, you're right. https://github.com/webtorrent/parse-torrent/blob/master/index.js#L182-L193

Sounds good then!

mmgoodnow (approved):

mmgoodnow:
Thanks! I'll test, merge, and release later today.

PR #33, opened 2020-12-03
Regex updates

tenshiak:
Some shows may have over 100 episodes in a season, current implementation won't handle that. Also season regex won't catch situations like S01-10, it'll see only S01. For examples see:
https://regex101.com/r/HCc47S/2
https://regex101.com/r/hkMIfw/1


mmgoodnow (approved):
Multi-season bundles are almost always disallowed on trackers so I doubt the season range modification will appreciably change the efficacy of this tool but I guess it can't hurt.

You'll probably want to remove these lines:
https://github.com/mmgoodnow/cross-seed/blob/446fca1fec093f0de3e0254901a7ccb3dda0db24/src/preFilter.js#L36-L41

Issue #46, opened 2021-03-03
False positives


Drazzilb08:
So I've ran into a issue with a couple of torrents that just will not recheck in my client, even though every aspect of the two torrents line up (or at least from what I can see).

 Instead of banging my head against a table to try to figure every torrent that just won't force check. 
Would there be a possible way to black list a certain release as opposed to just removing every time the script cycles through?

mmgoodnow:
I might be able to work something out, but it's also very important to me that cross-seed produces no false-positives. Is there any chance you could send me the offending torrents for inspection? If you use *nix you can edit out your passkey fairly easily with vim.

I understand if you don't want to do that but it would really be invaluable for tracking down this bug.

mmgoodnow:
Additionally, seeing a mediainfo dump for any particular files that fail the recheck would be very helpful as well.

Here are a couple other questions:
- What's the scale of this problem? Is it just one torrent, two, ten?
- If it's multiple, are they all on the same tracker?
- maybe they have embedded subtitles of a different language?
- How many files fail to recheck? Which ones?

Drazzilb08:
> * What's the scale of this problem? Is it just one torrent, two, ten?: The scale of the problem is quite small (which is a testament to the script). I don't remember how many. there were, I've removed them before I thought about making a GHI out of it.

> * If it's multiple, are they all on the same tracker? I'll get back to you on this one.

> * maybe they have embedded subtitles of a different language? I'll have to check for you.

> * How many files fail to recheck? Which ones?

At the end of all of it I'll let another full cycle run through and get back to you with more details.

Drazzilb08:
Okay so I let it go for a while (over night). So far I have 3 (out of my massive amount thats quite impressive for such a low number). However, one thing they have in common are:
1. From the same tracker: BHD
2. Season packs
3. 2 out of the 3 are from the same release group BTN of the same show. (other one is DON for a different show)
4. DON will check up to 36%
....
So I figured out what the main difference between the two BTN torrents are. 
In the folder structure the sub-folder of the original was `_` <-- just that
In he folder structure of the cross seed it has the release name
Cross-seed
Property.Brothers.S04.720p.WEB-DL.AAC2.0.h.264-BTN
|__ Episodes

Original 
Property.Brothers.S04.720p.WEB-DL.AAC2.0.h.264-BTN
|__ _
     |__Episodes

...
For the DON release 
Here is a paste of the media info
[...]

PR #51, opened 2021-03-16
Write custom path checking using original parse-torrent library

mmgoodnow:
This is the permanent fix for #46. It was fixed temporarily in https://github.com/mmgoodnow/cross-seed/commit/8385937bb39b57ed347f902037790b446a2be8d4 but that involved forking the parse-torrent package and referencing that fork in cross-seed's package.json. I was able to implement a custom solution using the original version of parse-torrent which is preferable because now I don't have to maintain the fork.

@Drazzlib08 I'm going to merge this in and cut a new release; please let me know if that bug comes back or if you see anything else that's funky.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
 				"bencode": "^2.0.1",
 				"chalk": "^4.1.0",
 				"commander": "^7.0.0",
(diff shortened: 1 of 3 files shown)


Drazzilb08:
The "quick fix" that you made previously has been holding strong. I haven't noticed anything strange in my client. I will be sure to keep you informed if I come across anything in the future.

Issue #84, opened 2021-06-11
Configurable fuzzy matching


newadventure079:
I noticed pull request https://github.com/mmgoodnow/cross-seed/pull/32 which upped the boundary for matching a torrent.  I think it should be taken a step further and allow that matching amount to be a user input (the default value would be left to how it is now and only changed if the user wanted to use the new input option).  This would allow for values, such as `.1 * length` or even `.2 * length` for torrents that include samples, nfos, subtitles and other stuff.  This would be useful for people who dont mind having a few extra torrent matches that aren't 100% and also want to seed ones they have that are 80% completed and will download the missing 20% and then begin seeding

Relevant code:
```
const lowerBound = length - 0.02 * length;
const upperBound = length + 0.02 * length;
```

mmgoodnow:
I like that idea. I'll add it to the backlog, but pull requests are always welcome!

ClandestineClout:
Definitely a good idea +1

mmgoodnow:
I've merged @jasonhazel's PR. No guarantees from me on it working well.

PR #157, opened 2021-12-02
Feature/fuzzy size match config

jasonhazel:
Attempts to add #84 

I haven't actually used cross-seed, but a buddy is and was asking about this.  Seeing that there was a feature request for it already and it seemed simple enough, I figured I would give it a stab.

Hopefully I'm not too far off. This is my first time working in TypeScript so all feedback is welcome.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -4,7 +4,7 @@ import { Option, program } from "commander";
 import { createRequire } from "module";
 import { inspect } from "util";
 import { generateConfig, getFileConfig } from "./configuration.js";
-import { Action, FUZZY_SIZE_THRESHOLD } from "./constants.js";
+import { Action } from "./constants.js";
 import { dropDatabase } from "./db.js";
 import { diffCmd } from "./diff.js";
 import { CrossSeedError } from "./errors.js";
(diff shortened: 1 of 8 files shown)


package-lock.json
```diff
@@ -1,10593 +0,0 @@
-{
```
mmgoodnow, line 1: Please remove your changes to package-lock.json

src/cmd.ts
```diff
@@ -109,6 +109,19 @@ async function run() {
 				"--notification-webhook-url <url>",
 				"cross-seed will send POST requests to this url with a JSON payload of { title, body }",
 				fileConfig.notificationWebhookUrl
+			)
+			.addOption(
+				new Option(
+					"-f, --fuzzy-size <decimal>",
```
mmgoodnow, line 115: Commander requires consistent naming, so an option called `--bobs-your-uncle` in the commander config would then become `bobsYourUncle` in the runtimeConfig. In order for the runtimeConfig to correctly pick up this option, it should be named `--fuzzy-size-threshold`. I also would prefer not to have a short flag called `-f` because `-f` sort of has its own meaning. 
```suggestion
					"--fuzzy-size-threshold <decimal>",
```

src/decide.ts
```diff
@@ -63,9 +64,11 @@ function compareFileTrees(candidate: Metafile, searchee: Searchee): boolean {
 }
 
 function sizeDoesMatch(resultSize, searchee) {
+	const fileConfig = getRuntimeConfig();
```
mmgoodnow, line 67: fileConfig and runtimeConfig are two separate things. Please rename this to runtimeConfig

src/cmd.ts
```diff
@@ -109,6 +109,19 @@ async function run() {
...
+					"The size difference allowed to be considered a match."
+				)
+					.default(
+						fallback(
+							fileConfig.fuzzySizeThreshold,
+							FUZZY_SIZE_THRESHOLD
```
mmgoodnow, line 121: can we just inline this constant? It's only used here

mmgoodnow (changes requested):
Thanks for the contribution! Much appreciated. 

How much have you tested this? I think you might need to also relax the prefiltering stage for this to actually work, but I'm not sure.

Otherwise, things look mostly good, just check out the comments please!

PR #197, opened 2022-02-06
Search by season and ep

### LINKED ISSUE #191: Search by season/ep through torznab
Hi. Neat project here.

I recently tried to run it with a Prowlarr torznab indexer to THE big tv tracker. But it could not find anything because it searched with `title S0X` and for some reason they don't support that. Searching manually, with that same torznab url and in Prowlarr without the season in the title do yield results.

I haven't looked at how the release name parsing is done, but if you could make it parse a title without season in it, it would work great for that site, and possibly others too.

mmgoodnow:
Closes #191

### DIFF
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -1,141 +1,247 @@
 import { zip } from "lodash-es";
 import fetch from "node-fetch";
 import xml2js from "xml2js";
+import { EP_REGEX, SEASON_REGEX } from "./constants.js";
 import { CrossSeedError } from "./errors.js";
 import { Label, logger } from "./logger.js";
 import { SearchResult } from "./pipeline.js";
 import { getRuntimeConfig, NonceOptions } from "./runtimeConfig.js";
-import { reformatTitleForSearching } from "./utils.js";
+import {
+	cleanseSeparators,
+	getTag,
+	MediaType,
+	reformatTitleForSearching,
+} from "./utils.js";
 
 interface TorznabParams {
-	t: "caps" | "search";
+	t: "caps" | "search" | "tvsearch" | "movie";
 	q?: string;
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -23,28 +30,32 @@ export function wait(n: number): Promise<void> {
...
+		: MediaType.OTHER;
 }
 
 export type Result<T> = T | Error;
 
 export const ok = <T>(r: Result<T>): r is T => !(r instanceof Error);
 
+export function cleanseSeparators(str: string): string {
+	return str
+		.replace(/[._()[\]]/g, " ")
```
bakerboy448, line 49: missing an escape for the [ here?

should be `/[._()\[\]/g` no?

src/torznab.ts
```diff
@@ -1,141 +1,247 @@
...
+		} else if (mediaType === MediaType.SEASON && caps.tvSearch) {
+			const match = name.match(SEASON_REGEX);
+			return {
+				t: "tvsearch",
+				q: encode(cleanseSeparators(match.groups.title)),
+				season: extractNumber(match.groups.season),
+			};
+		} else {
+			return {
+				t: "search",
```
bakerboy448, line 190: Why is movie searched not used?

src/utils.ts
mmgoodnow: `[` doesn't need to be escaped inside a `[]` but `]` does

src/torznab.ts
mmgoodnow: I started out with movie enabled but I had a season pack (of a miniseries) which got miscategorized as a movie. Since we're using the same arguments(just q) I figured I'd just fall back to t=search.

PR #199, opened 2022-02-20
Fix qbit contentLayout not being respected

### LINKED ISSUE #74: Incorrect save paths with qBittorrent injection

mmgoodnow:
fixes #74

@Byter09 and @ajhorvath20 - mind testing this out?

- Docker: Please pull from the tag `mg-fix-qbittorrent-content-layout`
- CLI: please clone the repo, checkout the `mg-fix-qbittorrent-content-layout` branch, run `npm ci`, `npm link`, and then `npm link cross-seed`.


Byter09:
Since reporting this I have gone back to just accepting torrents as they are (disabled the feature). I can activate it again and give this a shot. I guess I'll have to find a single-file torrent that is available on multiple of my trackers to test it out.

Byter09:
Am having trouble finding fitting torrents and I don't have that much time right now. Is this time critical? Otherwise I'll just let it run like this for like a week and see what happens with new incoming torrents from Radarr/Sonarr/Lidarr and report back, should one of them be a single-file one.

mmgoodnow:
> Is this time critical? Otherwise I'll just let it run like this for like a week and see what happens with new incoming torrents from Radarr/Sonarr/Lidarr and report back, should one of them be a single-file one (and of course allow for cross-seeding).

That works for me! It's a 9 month old issue so not very time critical lol

Byter09:
Alright. It downloaded something earlier and then cross-seed added a second torrent. It _looks_ weird because the content tabs aren't the same (the cross-seeding torrent only has the bare file in its list while the sonarr-added one includes the folder), but because the second added one has a different storage location it works out.

So yes. It's working, but a bit janky cause it's not immediately clear that it used the same file.

In case I'm describing this badly, here a simplification:
- original torrent
  - folder: `/torrents/`
  - contents shown as: `NameOfTorrent/NameOfTorrent.mkv`
- cross-seed added torrent
  - folder: `/torrents/NameOfTorrent/`
  - contents shown as: `NameOfTorrent.mkv`

So, as you can see, it _works_ but it on first glance it looks weird in the qBittorrent UI :D

mmgoodnow:
Yeah, I understand. Unfortunately that was the only way I was able to get it to work. 

I think there might be some sort of bug in qBittorrent that only respects your contentLayout if you're adding and starting a fresh torrent at 0%. Of course it's entirely possible I'm doing it wrong but I also did some tests directly from qBittorrent's add torrent dialog and I was getting the same results. 

I hope this will be sufficient for what we need it to do.

Byter09:
This is fine as is. It allows for me to use folders only, and that's exactly what I want. So thank you very much for putting time into this super minor issue 🥰

Byter09:
Alright, a few more torrents came in, most of them single file ones. Everything seems to work. Can't complain :D

(Just in case you needed more reports, idk)

mmgoodnow:
That's good enough for me! I'll merge this in later today

muddro1:
I think this broke crossseed for me. After updating, injected would start errors in the wrong folder.

mmgoodnow:
Please give more information. Qbittorrent version? Content layout you use? Which what kind of torrents? All or just some? What is the error?

muddro1:
> Please give more information. Qbittorrent version? Content layout you use? Which what kind of torrents? All or just some? What is the error?

Apologies. I actually think the issue is rather #74. I can post over there if you like. But essentially, using qbit 4.3.9, and any cross seed that gets injected goes to the default folder rather than the location of the torrent it is crossseeding for.

mmgoodnow:
This was supposed to fix #74 so it's likely it did cause your problem if it started recently. Please share more information/steps to reproduce

muddro1:
Yes it started with 3.3.1. I've since reverted to 3.3.0. Like I said, I use the injection method into qbit 4.3.9. It has been working fine where the injected will have the right path. Since updating to 3.3.1, anything that got injected went into the Default folder, so it would immediately error out and say missing files. To fix it, I would need to change category/path so its looking in the right folder. I'll go back to 3.3.1 and see if there is a difference between torrents that are single file or contained within a folder, since I don't recall that level f granularity. Not sure what other information I can produce.

mmgoodnow:
Yeah I'm wondering if this is a problem with auto tmm

muddro1:
Could be autotmm. But yes, basically what byter said. My setup is something like this: arr add torrent to incomplete category which has path /download/incomplete and after import moves to complete category at /download/complete. qbittorrent requires a default dl location, so for that I just set it to /download as it is basically never used.
When cross-seed injects, its injecting into /download, not /download/complete, where the original is saved and exists.

Byter09:
A possible solution would be to just copy as many properties of the original torrent to the new one as possible, right? Like straight up set the same category, path, everything. Add an additional tag called "cross-seed" like it's already doing and job done. Or am I imagining this too simple? :D

mmgoodnow:
There's seemingly a lot of logic that goes into calculating the actual content path of the torrent when you call the /torrents/add endpoint. You can pass `save_path`, but it's not the actual save path if you also set the category and have autoTMM turned on. You can pass `contentLayout` but it seems like that doesn't get respected unless the torrent is new (not totally sure, I never really figured that one out). The `contentLayout` of the existing torrent also has to be inferred from the difference between `save_path` and `content_path` - there don't seem to be any endpoints that expose its actual value.

If either of you is familiar with the qBittorrent api and also javascript, I'd love a little bit of help.

mmgoodnow:
It seems like we have several different boolean variables here - 

| `contentLayout` | `autoTMM` | has category | is single file torrent |
| --- | --- | --- | --- |
| false | false | false | false |
| false | false | false | true |
| false | false | true | false |
| false | false | true | true |
| false | true | false | false |
| false | true | false | true |
| false | true | true | false |
| false | true | true | true |
| true | false | false | false |
| true | false | false | true |
| true | false | true | false |
| true | false | true | true |
| true | true | false | false |
| true | true | false | true |
| true | true | true | false |
| true | true | true | true |

Byter09:
Wow :0 sorry I can't be of greater help. But if you ever need a tester for something please don't hesitate to ping me :D

mmgoodnow:
Actually I'll reopen #74 and have the convo over there

PR #205, opened 2022-03-11
Fix qBittorrent injection edge cases

### LINKED ISSUE #74: Incorrect save paths with qBittorrent injection

### LINKED ISSUE #144: FR - Enable Automatic Torrent Management
Is it possible to have cross-seed add torrents with Automatic Torrent Management.

Usecase: Hypothetically there could be a torrent that is tagged BTN was downloaded first and put into the incomplete-tv category. It does not get imported in sonarr yet because the show is still showing up as TBA. Meanwhile cross-seed runs and finds a torrent and adds it to qbit at the incomplete-tv category. Sonarr finally imports the torrent and the original torrent gets moved to a new category completed-tv and the save path gets changed. The cross-seed torrent does not change to the new path when changing the category to the completed-tv

Reason:  So if the cross-seeded torrent doesn't have auto torrent mgmt enabled and I manually change the category to completed-tv after the original moved. I would have to not only change the category but also the save path as well because if auto torrent mgmt is disabled then the category could be changed to completed-tv while the save path could be pointing to the incomplete directory

Alt Solution: Omit categories #95

mmgoodnow:
Hopefully fixes #74, fixes #144 

Please test. Docker tag `qbittorrent` or clone the repo, check out the `qbittorrent` branch, `npm link && npm link cross-seed`

### DIFF
--- a/.idea/dictionaries/mmgoodnow.xml
+++ b/.idea/dictionaries/mmgoodnow.xml
@@ -3,8 +3,11 @@
     <words>
       <w>bencode</w>
       <w>caronc</w>
+      <w>dlspeed</w>
+      <w>infohash</w>
       <w>jackett</w>
       <w>jsified</w>
+      <w>leechs</w>
       <w>qbittorrent</w>
       <w>rtorrent</w>
       <w>savepath</w>
@@ -13,6 +16,7 @@
       <w>seedable</w>
       <w>torznab</w>
       <w>tvsearch</w>
+      <w>upspeed</w>
       <w>xmlrpc</w>
     </words>
   </dictionary>
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,8 @@
 				"bencode": "^2.0.1",
 				"chalk": "^5.0.0",
 				"commander": "^8.3.0",
+				"fetch-blob": "^3.1.4",
+				"formdata-polyfill": "^4.0.10",
 				"lodash-es": "^4.17.21",
 				"lowdb": "^3.0.0",
 				"node-fetch": "^3.2.0",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,8 @@
 		"bencode": "^2.0.1",
 		"chalk": "^5.0.0",
 		"commander": "^8.3.0",
+		"fetch-blob": "^3.1.4",
+		"formdata-polyfill": "^4.0.10",
 		"lodash-es": "^4.17.21",
 		"lowdb": "^3.0.0",
 		"node-fetch": "^3.2.0",
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -1,18 +1,73 @@
+import { fileFrom } from "fetch-blob/from.js";
 import { FormData } from "formdata-polyfill/esm.min.js";
+import { unlink, writeFile } from "fs/promises";
 import fetch, { BodyInit, Response } from "node-fetch";
-import { File } from "fetch-blob/from.js";
+import { tmpdir } from "os";
 import parseTorrent, { Metafile } from "parse-torrent";
-import { dirname } from "path";
+import { dirname, join } from "path";
 import { InjectionResult } from "../constants.js";
 import { CrossSeedError } from "../errors.js";
 import { Label, logger } from "../logger.js";
-import { getRuntimeConfig, NonceOptions } from "../runtimeConfig.js";
+import { getRuntimeConfig } from "../runtimeConfig.js";
 import { Searchee } from "../searchee.js";
+import { isSingleFileTorrent } from "../torrent.js";
 import { TorrentClient } from "./TorrentClient.js";
 
 const X_WWW_FORM_URLENCODED = {
 	"Content-Type": "application/x-www-form-urlencoded",
 };
+
+interface SearchResult {
+	added_on: number;
+	amount_left: number;
+	auto_tmm: boolean;
+	availability: number;
+	category: string;
+	completed: number;
+	completion_on: number;
+	content_path: string;
+	dl_limit: number;
+	dlspeed: number;
+	download_path: string;
+	downloaded: number;
+	downloaded_session: number;
+	eta: number;
+	f_l_piece_prio: boolean;
+	force_start: boolean;
+	hash: string;
+	infohash_v1: string;
+	infohash_v2: string;
+	last_activity: number;
+	magnet_uri: string;
+	max_ratio: number;
+	max_seeding_time: number;
+	name: string;
+	num_complete: number;
+	num_incomplete: number;
+	num_leechs: number;
+	num_seeds: number;
+	priority: number;
+	progress: number;
+	ratio: number;
+	ratio_limit: number;
+	save_path: string;
+	seeding_time: number;
+	seeding_time_limit: number;
+	seen_complete: number;
+	seq_dl: boolean;
+	size: number;
+	state: string;
+	super_seeding: boolean;
+	tags: string;
+	time_active: number;
+	total_size: number;
+	tracker: string;
+	trackers_count: number;
+	up_limit: number;
+	uploaded: number;
+	uploaded_session: number;
+	upspeed: number;
+}
 
 export default class QBittorrent implements TorrentClient {
 	url: URL;
@@ -122,6 +177,9 @@ export default class QBittorrent implements TorrentClient {
 	async getTorrentConfiguration(searchee: Searchee): Promise<{
 		save_path: string;
 		isComplete: boolean;
+		autoTMM: boolean;
+		category: string;
+		content_path: string;
 	}> {
 		const responseText = await this.request(
 			"/torrents/info",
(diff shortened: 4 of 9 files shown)


Byter09:
I'll deploy it on my server and report any issues. Thanks for all the effort you're putting into this. I really appreciate it.

muddro1:
Didn't run it for too long. When one had injected, which was a single file, it created it in a folder, so ended up re-downloading.

mmgoodnow:
@muddro1 a couple questions 

- what content layout was the source torrent added with originally? (Subfolder, NoSubfolder, Original) 
- did you have auto torrent management on?
- would you be willing to send me the torrent file in the discord after removing all announce urls?

muddro1:
> @muddro1 a couple questions
> 
>     * what content layout was the source torrent added with originally? (Subfolder, NoSubfolder, Original) - Not 100% sure. Do you mean the one prior to adding the cross-seed? If that one, then that was nosubfolder (is there a way for me to check to confirm, I am just basing this on it dling xxxxx.mkv directly with no folder.
> 
>     * did you have auto torrent management on? - I generally have it on. In 3.3.0 the crossseeds would be added with ATM turned off. I think after that version they have been turned on by your app.
> 
>     * would you be willing to send me the torrent file in the discord after removing all announce urls? -Sure if you can show me how.

Byter09:
Just to leave my comment on this as well, single file torrents are now exactly displayed like their "original", down to the folder view and category. Very nice. The only thing differentiating them is the tag.

@muddro1 If you had previous torrents without the subfolder and you're now using it together with this version of cross-seed, when it adds cross-seeds for those previous torrents, qBittorrent will put them in a folder and there's no way around that, so it'll download again. I circumvent that by simply not checking torrents older than 3 days. But yeah, that just tells me its working, cause it happened for one of mine as well and thats essentially my fault as I changed the setting back to subfolder creation.

mmgoodnow:
> If you had previous torrents without the subfolder and you're now using it together with this version of cross-seed, when it adds cross-seeds for those previous torrents, qBittorrent will put them in a folder and there's no way around that, so it'll download again. 

Not sure I follow, but this does seem like a cross-seed problem. Could you elaborate with a more specific example?

Byter09:
Oh don't worry this is doesn't seem like a bug to me.

Essentially, before I switched back to subfolder creation I had torrents as single files to at least get cross-seed working, ya know.

So lets say the file is here: `torrents/file.mkv`

Now, because I re-enabled subfolder creation, the cross-seeding torrent corresponding to that original one will have a path of `torrents/file/file.mkv`. But that's to be expected. I should've moved the original to a subfolder first. This does not seem like a cross-seed problem to me, cause newly added torrents _will_ be in a subfolder now and cross-seed will have a correct path when injecting cross-seeds for those.

mmgoodnow:
> But that's to be expected. 

It's not. `cross-seed` doesn't check your global content layout setting, it attempts to read the current content layout of the source torrent in the client. Could you also answer these questions for the torrents you noticed failed?

- what content layout was the source torrent added with originally? (Subfolder, NoSubfolder, Original) 
- did you have auto torrent management on?
- would you be willing to send me the torrent file in the discord after removing all announce urls?

Byter09:
Oh interesting!
- original
- no
- sure, am already on the discord, just ping me haha

mmgoodnow:
> Not 100% sure. Do you mean the one prior to adding the cross-seed? If that one, then that was nosubfolder (is there a way for me to check to confirm, I am just basing this on it dling xxxxx.mkv directly with no folder.

That could be NoSubfolder or Original. How did the torrent get added to qBittorrent in the first place? If it was automatic, you could check your Sonarr/Radarr (or whatever else added it) settings to see if there's a content layout option. Otherwise it would probably just be your default (which you can see in Settings)

Re: removing announce urls, one way to do it is right from inside your client. You can remove a tracker URL by right clicking on the tracker in the trackers list: 
image

After that you can grab/copy the torrent file from the session directory, send it to me, then add the announce URL back.

Byter09:
In addition to the storage location issue, I just noticed that the cross-seed injected torrent also has no category. The ones before and after have it correctly set. Must be some really weird edge case :thinking:

mmgoodnow:
> In addition to the storage location issue, I just noticed that the cross-seed injected torrent also has no category. The ones before and after have it correctly set. Must be some really weird edge case.

The same exact torrent both has no category and also has missing files?

I'm wondering if it picked up a cross-seeded torrent from a previous run as the "source torrent" - this would be the reason it wouldn't have the category. 

I'm wondering if the torrent it had identified as the source torrent was _already_ in the "missing files" state. Might that have been the case?

What happens if you remove all torrents that cross-seed found and injected for the offending torret, and then run it again?

Byter09:
It definitely must've picked up the original, assuming it is correctly ignoring torrents with the "cross-seed" label.

The original torrent was definitely not in a missing files state.

Will test removing torrents and running it again in a few hours. Btw, with everything now set up using subfolders, new torrents being added and cross seed injecting more, everything seems to work flawlessly. This might just be a super rare edge case when your original has a different storage method (original) compared to what qBittorrent is set to now (subfolder). Is this actually something cross-seed should be able to handle now? Cause I don't think this will happen too often in the wild 🤔

mmgoodnow:
> It definitely must've picked up the original, assuming it is correctly ignoring torrents with the "cross-seed" label.
> 
> The original torrent was definitely not in a missing files state.

When cross-seed runs, we first read all the torrent files in your torrent directory, and then deduplicate them based on their names (cross-seed makes the assumption that any two torrents with the same name came from an earlier cross-seed run). 
Because of this, for a given group of torrents that all have the same name, the representative torrent is not necessarily the "original" torrent, but the first torrent in the group (which is likely the one with the "earliest" info hash, alphabetically). 

For the purposes of strictly finding cross-seeds, this is sufficient because all torrents in the group have identical file trees, which is how we do matching. Cross-seed only communicates with your torrent client when it finds a match and needs to inject it. 

Because of this, for the "missing files" bug to fully go away (to the best of my knowledge thus far), you'll need to delete any torrents in the client which are in a "missing files" state, and for autoTMM to fully start working, you'll need to either delete *all* torrents that cross-seed previously injected and re-run it, or assign categories to all of the torrents cross-seed previously injected.

mmgoodnow:
> The original torrent was definitely not in a missing files state.

@Byter09 do you mean to say that there may have been some cross-seeds of the original torrent that _were_ in a missing files state?

Byter09:
As described in a comment above, the torrent cross-seed injected was in a missing files state till I adjusted the location path again.

> For the purposes of strictly finding cross-seeds, this is sufficient because all torrents in the group have identical file trees, which is how we do matching. Cross-seed only communicates with your torrent client when it finds a match and needs to inject it.
>
> Because of this, for the "missing files" bug to fully go away (to the best of my knowledge thus far), you'll need to delete any torrents in the client which are in a "missing files" state, and for autoTMM to fully start working, you'll need to either delete all torrents that cross-seed previously injected and re-run it, or assign categories to all of the torrents cross-seed previously injected.

And for that exact reason I think its fine and not really a bug. This branch/image works perfectly so far. It's only that one torrent that it injected which ended up in a missing-files state, and I manually fixed that. So I _really_ do not think this is something to worry about.

Kind of busy this coming week, so I can't really do any testing, but again, everything is working so far, with autoTTM off and subfolders on.

KyleSanderson:
This is still broken - it's still matching single units with folders. Data is the exact same, it's the files themselves that are different. .mkv came first.

image

mmgoodnow:
@KyleSanderson can you be more specific? I don't understand what you mean.

KyleSanderson:
> @KyleSanderson can you be more specific? I don't understand what you mean.

As part of BEP3 - a torrent file contains a name key that may or not may be honoured by the client.

In the case of a single torrent file, this contains the name of the file.
In the case of a multi-file torrent file, the name key is the suggested directory to house these multiple files.

The infohash will be the same, but the actual layout of the contents of the file will be different based on this differentiation. When cross-seed is injecting files, the infohash is the exact same, but some files come in the multi-file format, and others come in the single file format. This results in the client, qBittorrent, being unable to find the existing file, and instead triggers a brand new download action, resulting in two+ copies of the exact same file.

Lastly, this is using the "Original" Content Layout field on qBittorrent.

mmgoodnow:
I'm pretty sure that the `files` array (for a multi file torrent) and `length` (single) are inside the info dictionary, meaning that you couldn't have a multi file torrent with the same info hash as a single file torrent. Ultimately though that probably doesn't have any impact on the bug.

Can you examine these torrent files and verify that cross-seed indeed matched a multi-file torrent to a single-file torrent?

KyleSanderson:
> I'm pretty sure that the `files` array (for a multi file torrent) and `length` (single) are inside the info dictionary, meaning that you couldn't have a multi file torrent with the same info hash as a single file torrent. Ultimately though that probably doesn't have any impact on the bug.
> 
> Can you examine these torrent files and verify that cross-seed indeed matched a multi-file torrent to a single-file torrent?

Yes - this is guaranteed.

image

mmgoodnow:
Guaranteed by what?

KyleSanderson:
> Guaranteed by what?

There is no other way that these files are added other than cross-seed into this client.

mmgoodnow:
Just because they downloaded data doesn't mean that cross-seed is finding false positives though.

KyleSanderson:
> Just because they downloaded data doesn't mean that cross-seed is finding false positives though. I think it means they're getting injected into the wrong place.

The source torrents are constructed fundamentally differently. It's the same underlying data, but the torrent types that it's not doing the placements (or migrations) properly stem from these single file torrents to multi-files, with a single directory. It's a single destination file, but under the files dict so it creates a dir as per BEP3 which is where cross-seed is falling short.

mmgoodnow:
@KyleSanderson has identified a bug in this PR specifically with torrents which have autoTMM on but no category. @muddro1 do you use that configuration?

mmgoodnow:
Okay @KyleSanderson that bug should be fixed. Instead of using this heuristic to determine the content layout:
```js
const shouldManuallyEnforceContentLayout =
    isSingleFileTorrent(newTorrent) &&
    dirname(content_path) !== save_path;
```

I added another API call to qBittorrent (which I had been hoping to avoid):
```js
async isSubfolderContentLayout(searchee: Searchee): Promise<boolean> {
	// hit /torrents/files endpoint and get json back as "files" variable
	return files.length === 1 && name !== path.posix.basename(name);
}

const shouldManuallyEnforceContentLayout =
    isSingleFileTorrent(newTorrent) &&
    (await this.isSubfolderContentLayout(searchee));
```

muddro1:
I use a category, but it sounds like what I was experiencing:

Original torrent added by arr, in Original layout. It was a single file so it was in the designated folder (per category) just the file (not contained within a folder of its own name). When injection happened it was placed in a subfolder of its own name, and obviously didn't match so started downloading all over again. I believe in that version cross seeds injected with atm on.

mmgoodnow:
Okay, I think qbit injection has been broken on master for long enough lol, and I have a relatively high degree of confidence in this PR at this point. So I'm gonna merge this in. Hopefully any remaining bugs are in low-volume configurations.

mmgoodnow:
v3.3.3

KyleSanderson:
> Okay @KyleSanderson that bug should be fixed. Instead of using this heuristic to determine the content layout:
> 
> ```js
> const shouldManuallyEnforceContentLayout =
>     isSingleFileTorrent(newTorrent) &&
>     dirname(content_path) !== save_path;
> ```
> 
> I added another API call to qBittorrent (which I had been hoping to avoid):
> 
> ```js
> async isSubfolderContentLayout(searchee: Searchee): Promise<boolean> {
> 	// hit /torrents/files endpoint and get json back as "files" variable
> 	return files.length === 1 && name !== path.posix.basename(name);
> }
> 
> const shouldManuallyEnforceContentLayout =
>     isSingleFileTorrent(newTorrent) &&
>     (await this.isSubfolderContentLayout(searchee));
> ```

Still not correct on 3.4.0. The source content is a single file, I'm now forcing a directory for layout in qbittorrent. Getting "Missing Files" from the cross-seed inject (and even new downloads triggering).

muddro1:
Working as it should for me

mmgoodnow:
@KyleSanderson  do you use windows or posix? Also can you try the command `cross-seed tree` on the torrent?

Issue #53, opened 2021-03-21
Cross-seed based on data files


mmgoodnow:
After #52 we'll have an abstraction over searchees. This means that we'll be able to easily create searchees out of entries in your data dirs. 

Plan:
- add a --data-dir command line option that can be used multiple times
- concatenate those with the torrents found in the session dir
- be careful to deduplicate these

ajhorvath20:
This one will be slick. Nice one

ajhorvath20:
I'm keeping an eye on this 👀

newadventure079:
Just an FYI, this is how https://github.com/BC44/Cross-Seed-AutoDL operates

mmgoodnow:
You do have torrent files inside your qbit config, just gotta hunt them down.

Byter09:
I found that folder. Luckily the docker container maps it to the outside so I got access too. Thanks. Hope that wasn't too off topic.

mmgoodnow:
@byter09 No worries! I am gonna mark as off topic though.

PR #212, opened 2022-03-22
Add announce API

### LINKED ISSUE #204: IRC monitoring support
Just throwing this out there. Perhaps some integration could be added with one of the below projects, since they are becoming quite a popular way to fetch new torrents.

https://gitlab.com/cloudb0x/trackarr
https://github.com/autobrr/autobrr

I did notice from https://github.com/mmgoodnow/cross-seed/issues/13 that RSS is now planned. Either of these options will make a great addition.

mmgoodnow:
Closes #204 

```http
POST /api/announce
Content-Type: application/json
{
  "name": "string", // torrent name
  "guid": "string", // ideally the torrent's comments/listing url, would be for caching purposes
  "link": "string", // download link
  "size": "number", // in bytes
  "tracker": "string" // this is for logging purposes
}
```

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -4,7 +4,7 @@ import path from "path";
 import { appDir } from "./configuration.js";
 import { Decision, TORRENT_CACHE_FOLDER } from "./constants.js";
 import db, { DecisionEntry } from "./db.js";
-import { SearchResult } from "./pipeline.js";
+import { Candidate } from "./pipeline.js";
 import { Label, logger } from "./logger.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 import { Searchee } from "./searchee.js";
@@ -76,12 +76,12 @@ function sizeDoesMatch(resultSize, searchee) {
 	return resultSize >= lowerBound && resultSize <= upperBound;
 }
 
-async function assessResultHelper(
-	{ link, size }: SearchResult,
+async function assessCandidateHelper(
+	{ link, size }: Candidate,
 	searchee: Searchee,
 	hashesToExclude: string[]
 ): Promise<ResultAssessment> {
-	if (!sizeDoesMatch(size, searchee)) {
+	if (size != null && !sizeDoesMatch(size, searchee)) {
 		return { decision: Decision.SIZE_MISMATCH };
 	}
 
@@ -126,12 +126,12 @@ function cacheTorrentFile(meta: Metafile): void {
 }
 
 async function assessAndSaveResults(
-	result: SearchResult,
+	result: Candidate,
 	searchee: Searchee,
 	Guid: string,
 	infoHashesToExclude: string[]
 ) {
-	const assessment = await assessResultHelper(
+	const assessment = await assessCandidateHelper(
 		result,
 		searchee,
 		infoHashesToExclude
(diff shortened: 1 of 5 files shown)


mmgoodnow:
@LudvigLundgren @KyleSanderson you might be interested in this.

Let's say you have 10 torznab trackers configured.

The benefits of `/api/announce` over `/api/webhook` are as follows:

| Path | Searches | Snatches |
| --- | --- | --- |
| `/api/announce` | 0 | 0–1 (compares name and size, if passed, before snatching) |
| `/api/webhook` | 10 | 0–(number of search results) (compares size before snatching, also caches previous rejections) |

ludviglundgren:
I haven't taken the time to answer on the other issue yet, but lets do some of it here.

So with this you want to be able to get a POST early on a new announce, and if it's a match (as in torrent already in client..?) you'll respond with 204, correct?

So lets say there's a method for pre-filters or something along those lines that can be HTTP based, which would hook into this, and depending on answer either run the filter or not. I need to think a bit more about the flow since it probably still have to go through the filters in autobrr to see if those match.

But in the first case you want enough data like torrenturl to be able to directly download it and be done there, or should it just be taken into consideration and continue like a match, since you only get the name from the start.

Essentially:
1. Announce comes in
2. Send POST to /announce with release name to see if match
3. Responds with 204 if match, does it start the download there and then?

mmgoodnow:
It responds with 204 if the request is received successfully, before actually determining if it's a match or not. If it is, then it injects or saves the cross-seed.

The weird behavior of the response coming back before the actual work is done is mostly just because that's how the/api/webhook endpoint works; I figured it'd be better to close the connection than keep it open for potentially a very long time (some jackett queries can go 30 seconds or more).

1. Announce comes in
2. POST /api/announce with parameters
3. Does some light validation on the request body, responds with 204 or 400
4. Indexes torrentDir
5. check for torrent in dir with same name as announced
6. run match/decision logic
7. inject or save if match

Given that we're not making any jackett/torznab queries, and the rest of the process should be very quick, we _could_ potentially make the endpoint behave synchronously—that is, waiting to respond with an actually relevant status code once the matching and injecting has finished. Do you see value in that?

ludviglundgren:
Making the `announce` call synchronous and waiting for the response, even if it takes some seconds would be best for the case of autobrr usage. 

That way, if it isn't a cross-seed match it can still continue with filtering because it might be new content from the origin tracker.

I need to decide on where this fits in autobrr and how it will work, but having this in place will make that easier.

In the matching/decision logic for announce, does it check only files loaded in the client or will it check for possibly orphaned files on disk as well?

ludviglundgren:
Btw, what unit is size in the announce body? Bytes?

You might also want to add a field for custom headers because some trackers use cookies instead of url params for auth keys.

Could be a basic array of strings like `KEY=value` or an array of objects with `{ "key": "value" }`.

mmgoodnow:
> Btw, what unit is size in the announce body? Bytes?

Yes, bytes. 

> That way, if it isn't a cross-seed match it can still continue with filtering because it might be new content from the origin tracker.

What's the benefit of filtering based on cross-seed's response? Would it be better than just sending all announces to all consumers? Presumably these consumers have their own domain-specific, and very robust, filtering logic (thinking sonar and radar mostly).

> In the matching/decision logic for announce, does it check only files loaded in the client or will it check for possibly orphaned files on disk as well?

It only checks files loaded in the client. Support for orphaned files on disk is planned in #53 but probably won't happen any time soon.

ludviglundgren:
> What's the benefit of filtering based on cross-seed's response? Would it be better than just sending all announces to all consumers? Presumably these consumers have their own domain-specific, and very robust, filtering logic (thinking sonarr and radarr mostly).

If it's a match of existing content then we know that and don't continue with filters and actions that might add it to different category etc.

I don't want autobrr to rely on external tools like the arrs. Integrations are fine, because it's up to the user. 

And not everything can go into an arr. Music is a bit meh with Lidarr. Games and software have no proper arr. elearning/ebooks maybe do. There's other types of content than just movies and tv that people want automatically downloaded, and cross-seeded if possible.

mmgoodnow:
But what would happen if you don't have cross-seed integrated then?

ludviglundgren:
It would simply work like it does today and listen to announces, filter them and on match run actions (qbit, arrs etc) :slightly_smiling_face: 

Integrating with cross-seed would be a nice to have since more than a handful users already is using it but with webhook or scan on cron etc.

mmgoodnow:
In order to not let perfect be the enemy of good, I'm going to forgo custom headers for now.

PR #243, opened 2022-05-15
Respect includeEpisodes even when searchAll is true

edgeworn:
This PR changes to respecting the includeEpisodes flag even when searchAll is true.

I believe that this would be a useful change for two reasons
1. It exclusively adds functionality without removing it. It is still possible to regain the old functionality by using both searchAll and includeEpisodes. But there is a new option, searchAll without includeEpisodes, which was impossible before
2. Using searchAll without includeEpisodes is very useful. Including episodes takes far too long when you automate downloading new content, but without searchAll many legitimately cross-seedable torrents are skipped by the filter. For example, out of about 10,000 torrents searched, roughly on the order of 10% of them were not episodes but were filtered out when searchAll was false, even though matches were found when searchAll was set to true. Among these were some of the largest torrents in the library (ones over 1TB), which are arguable the most beneficial to cross-seed.

Related discussion: https://github.com/mmgoodnow/cross-seed/discussions/91

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Options:
   -T, --torznab <urls...>               Torznab urls with apikey included (separated by spaces)
   -i, --torrent-dir <dir>               Directory with torrent files
   -s, --output-dir <dir>                Directory to save results in
-  -a, --search-all                      Search for all torrents regardless of their contents (default: false)
+  --include-non-videos                    Include torrents which contain non-video files (default: false)
   -e, --include-episodes                Include single-episode torrents in the search (default: false)
   --fuzzy-size-threshold <decimal>      The size difference allowed to be considered a match. (default: 0.02)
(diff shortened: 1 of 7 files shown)


mmgoodnow:
I like this idea. I'm working on a v4 because #242 has some breaking changes; would be a good opportunity to stick this breaking change in as well.

mmgoodnow (changes requested):
Would you mind updating the option descriptions in config.template.js, config.template.docker.js, and cmd.ts?

edgeworn:
Updated the documentation
Let me know if you would like me to change the verbiage or formatting or anything
Thanks!

mmgoodnow:
I'm going to rename the new flag to `nonVideos`/`--non-videos` because I think that's more in spirit of the new behavior.

mmgoodnow (approved):

PR #257, opened 2022-05-29
Fix sonarr category issue

### LINKED ISSUE #254: Feature Request: Qbittorrent Category
Currently cross-seed adds torrents with a special tag, and then the previously used/existing category of the torrent file. **However, it would be great to have the option to set a specific _category_ for injected torrents.** 

Seems like the most obvious use case for this would be for seedbox clients, as users' files are often synced to and from seedbox. Categories allow users to set file location, so this would be a great feature for seedbox users that rely on this sort of workflow to and from remote devices.

mmgoodnow:
Closes #254

Behavior: 
- Sets up a category with the same savePath as the source torrent, but appends `.cross-seed` to the original category name.
- If the source torrent doesn't have a category, the injected torrent will also be uncategorized.

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -263,6 +263,7 @@ export default class QBittorrent implements TorrentClient {
 		newTorrent: Metafile,
 		searchee: Searchee
 	): Promise<InjectionResult> {
+		const { duplicateCategories } = getRuntimeConfig();
 		if (await this.isInfoHashInClient(newTorrent.infoHash)) {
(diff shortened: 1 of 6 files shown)


KyleSanderson:
Any update on this?

Byter09:
I didn't try this out as I thought after this long time it's about to be merged. Is this practically done or does it need more testing? If so I'd switch my instance over to this branch.

mmgoodnow:
It's been pretty tested at this point though. Jonza is using it in prod I believe

PR #292, opened 2022-07-23
Send 204 if no match for announce POST

zze0s:
Per discussion on Discord.

This changes the behavior for the `/announce` endpoint. 

Before: Send 200 OK on post with data if a match.
Now: Send 200 OK with data if match, and 204 No content if no match.

Why: to be able to easily use [autobrr](https://github.com/autobrr/autobrr) external filter to run cross-seed check early on.


mmgoodnow (approved):
Thanks for updating the docs!

PR #300, opened 2022-08-09
Search for torrents with data instead of .torrent files

### LINKED ISSUE #53: Cross-seed based on data files

akesher123:
closes #53 

At the moment this is only a minimum viable product to experiment with this working.

Limitations: 

- Cannot simultaneously search with .torrent files and data, only one or the other.
- Cannot inject yet, only saving works
- Must be run on the top level directory of torrent downloads. Torrents in that directory cannot be nested
- Can only run on one directory at a time

Changes planned:
-  Fix injecting
-  Add ability to search multiple directories

Maybe changes:
- Have option to just search on individual video files, ignoring folders. This would cost matches but could allow matches within an organized directory such as the one produced by radarr/sonarr
- Run with .torrent files and data in the same run. This is probably a good idea and useful for those that have their torrent and usenet downloads separate.
- Maybe some sort of hardlinking solution for better matching in the first maybe-change. Ideally this could extract an .mkv or other video file from an organized directory, make a hardlink to it elsewhere where it can match a torrent to 99% and download .nfo or samples to complete seeding. This is definitely larger and scope and would be tough to accomplish.


src/searchee.ts
```diff
@@ -19,6 +20,39 @@ export interface Searchee {
...
+		{
+			path: path.relative(path.join(dirPath, ".."), file),
+        	name: path.basename(file),
+			length : fs.statSync(file).size
+		})
+	)
+	return torrentFiles
+  }
+
+function getFilePathsFromPath(dirPath, arrayOfFiles) {
```
mmgoodnow, line 40: nit: can we move this function above `getFilesFromPath`?

src/searchee.ts
```diff
@@ -19,6 +20,39 @@ export interface Searchee {
 	length: number;
 }
 
+function getFilesFromPath(dirPath): File[] {
```
mmgoodnow, line 23: naming suggestion: `getFilesFromDataRoot(root: string)`

`dirPath` is slightly misleading because we're not sure if it's a directory or not yet

src/searchee.ts
```diff
@@ -19,6 +20,39 @@ export interface Searchee {
...
+	if (fs.statSync(dirPath).isDirectory()) {
+		var files: string[] = getFilePathsFromPath(dirPath, []);
+	} else {
+		var files: string[] = [dirPath];
+	}
```
mmgoodnow, line 28: can we absorb this conditional inside the helper `getFilePathsFromPath`?

src/searchee.ts
```diff
@@ -68,3 +102,28 @@ export async function createSearcheeFromTorrentFile(
...
 }
+
+export async function createSearcheeFromPath(
+	filepath: string
+): Promise<Result<Searchee>> {
+		//const rawPathSegments: Buffer[] = filepath["path.utf-8"] || filepath;
+		//const pathSegments = rawPathSegments.map((s) => s.toString());
+
+		const fileName : string = path.basename(filepath);
+		var fileList : File[] = getFilesFromPath(filepath);
```
mmgoodnow, line 113:
```suggestion
		const fileList : File[] = getFilesFromPath(filepath);
```

src/searchee.ts
```diff
@@ -68,3 +102,28 @@ export async function createSearcheeFromTorrentFile(
...
+		var totalLength = 0;
+		fileList.forEach(file => totalLength += file.length);
```
mmgoodnow, line 115:
```suggestion
		const totalLength = fileList.reduce<number>((runningTotal, file) => runningTotal + file.length, 0);
```

mmgoodnow (commented):
thanks so much for doing this! I left a few comments, mostly just some naming nits. 

Please delete `dist.zip` when you're ready.

mmgoodnow:
> Maybe some sort of hardlinking solution for better matching in the first maybe-change. Ideally this could extract an .mkv or other video file from an organized directory, make a hardlink to it elsewhere where it can match a torrent to 99% and download .nfo or samples to complete seeding. This is definitely larger and scope and would be tough to accomplish.

you can safely call this out of scope lol

src/cmd.ts
```diff
@@ -63,6 +63,11 @@ function createCommandWithSharedOptions(name, description) {
 			"Torznab urls with apikey included (separated by spaces)",
 			fallback(fileConfig.torznab)
 		)
+		.option(
+			"-d, --data-dir <dir>",
```
mmgoodnow, line 67: you can refer to the torznab syntax right above this to support an array of datadirs as space-delimited option arguments.

mmgoodnow:
Also if you want to try tackling Daemon mode, I'd start by looking at implementation/usages of `indexTorrentDir`

mmgoodnow:
> Must be run on the top level directory of torrent downloads. Torrents in that directory cannot be nested

I think we should keep this limitation as-is. I don't think there's really any way to easily figure out if a given file/directory at arbitrary nesting level X is a root or not, and people don't really store their downloads that way either.

mmgoodnow:
might be able to use https://stackoverflow.com/a/54439981 to reduce the number of disk reads

src/searchee.ts
akesher123: I could try to absorb it, but it might not look exactly like this. It has to be properly recursive since there's no way to know how far the torrent directory tree goes. And to be honest, I'd rather not refactor it. It's a little ugly and inelegant right now but I had issues when I initially tried to do it this way. Maybe my recursion skills have fallen off...

src/cmd.ts
akesher123: Will do in the next round of post-review coding.

src/pipeline.ts
```diff
@@ -185,16 +187,28 @@ export async function checkNewCandidateMatch(
...
-	const { torrents } = getRuntimeConfig();
+async function findSearchableTorrents(useData : boolean) {
+	const { torrents, dataDirs } = getRuntimeConfig();
 	let parsedTorrents: Searchee[];
-	if (Array.isArray(torrents)) {
-		const searcheeResults = await Promise.all(
-			torrents.map(createSearcheeFromTorrentFile)
-		);
+	if (useData) {
+		const fullPaths = [];
```
mmgoodnow, line 194: can we extract this new code to a helper in a different file (maybe searchee.ts), a la loadTorrentDirLight? I like to keep new code to a minimum in pipeline.js because I think it's already pretty messy

src/clients/QBittorrent.ts
```diff
@@ -248,6 +257,10 @@ export default class QBittorrent implements TorrentClient {
 	}
 
 	async isSubfolderContentLayout(searchee: Searchee): Promise<boolean> {
+		const { dataDirs } = getRuntimeConfig();
+		if (dataDirs.length > 0) {
+			return statSync(searchee.path).isDirectory();
+		}
```
mmgoodnow, line 263: subfolder content layout only really takes effect for single-file torrents, but I'm inclined to say that it'll be easiest to just ignore content layouts entirely in v1 and only support `Original`. That would mean that we can just return false in this situation.

src/clients/QBittorrent.ts
```diff
@@ -224,6 +225,14 @@ export default class QBittorrent implements TorrentClient {
 		autoTMM: boolean;
 		category: string;
 	}> {
+		const { dataDirs } = getRuntimeConfig();
+		if (dataDirs.length > 0) {
+			const save_path: string = join(searchee.path, "/..");
+			const isComplete: boolean = true;
+			const autoTMM: boolean = false;
+			const category: string = " ";
+			return {save_path, isComplete, autoTMM, category}
```
mmgoodnow, line 234: nit: can we inline these variables into the return object? also, it might be clearest to just omit the category property entirely in this case (and make it optional in the type) since it doesn't get used.

src/clients/QBittorrent.ts
```diff
@@ -224,6 +225,14 @@ export default class QBittorrent implements TorrentClient {
 		autoTMM: boolean;
 		category: string;
 	}> {
+		const { dataDirs } = getRuntimeConfig();
+		if (dataDirs.length > 0) {
```
mmgoodnow, line 229: this conditional can be searchee-based to make it more future proof
```suggestion
		if (searchee.path) {
```

src/clients/QBittorrent.ts
```diff
@@ -224,6 +225,14 @@ export default class QBittorrent implements TorrentClient {
...
+			const save_path: string = join(searchee.path, "/..");
```
mmgoodnow, line 230: does `path.dirname(searchee.path)` work here instead of `path.join(searchee.path, '/..')`?

src/clients/QBittorrent.ts
```diff
@@ -274,10 +287,12 @@ export default class QBittorrent implements TorrentClient {
 		try {
 			const { save_path, isComplete, autoTMM, category } =
 				await this.getTorrentConfiguration(searchee);
-
-			const newCategoryName = duplicateCategories
+			
+			const newCategoryName = dataDirs.length == 0 ?
```
mmgoodnow, line 291:
```suggestion
			const newCategoryName = searchee.infoHash ? 
```

src/searchee.ts
```diff
@@ -68,3 +102,27 @@ export async function createSearcheeFromTorrentFile(
 		return e;
 	}
 }
+
+export async function createSearcheeFromPath(
+	filepath: string
+): Promise<Result<Searchee>> {
+		//const rawPathSegments: Buffer[] = filepath["path.utf-8"] || filepath;
+		//const pathSegments = rawPathSegments.map((s) => s.toString());
```
mmgoodnow, line 110: nit: commented code

src/searchee.ts
```diff
@@ -68,3 +102,27 @@ export async function createSearcheeFromTorrentFile(
...
+			files:  fileList,
+			path: filepath,
+			name: fileName,
+			length: totalLength,
+		};
+	// } catch (e) {
+	// 	logger.error(`Failed to parse ${basename(filepath)}`);
+	// 	logger.debug(e);
+	// 	return e;
+	// }
```
mmgoodnow, line 126: nit: commented code

mmgoodnow (commented):
I noticed you're using a lot of forEach patterns like the following:
`const x = []; arr.forEach(() => x.push(etc));`

These are completely valid code, but the conventions in this project use `.map` and `.reduce` instead - I find them to be a little more hygienic with variables. If you're unfamiliar with using .map/.reduce, I can do those conversions.

src/pipeline.ts
```diff
@@ -185,16 +187,28 @@ export async function checkNewCandidateMatch(
...
+		dataDirs.forEach(dataDir => {
+			if (fs.statSync(dataDir).isDirectory()) {
+				fs.readdirSync(dataDir).forEach(file => fullPaths.push(path.join(dataDir, file)));
+			} else {
+				fullPaths.push(dataDir);
+		}})
```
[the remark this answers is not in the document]
akesher123: This is scanning at the top level directory, every item one level down should be an individual torrent. Each folder is 1 torrent, and each file outside those folders should also be 1 torrent. An mkv file just sitting in the top level torrent downloads directory is probably a torrent by itself.

akesher123:
Yes, I would rather you go through and do the map edits. I'm not familiar with the syntax and I have a high probability of messing something up with it.

akesher123:
Other limitations I've come up with:

Regarding partial torrents, as I understand it qbit will allocate the storage space required at the start of the download, so there's no way to determine whether a download is complete or not. Right now it assumes true. I think it might be wise to at least give the option to recheck torrents or start paused so there's a safeguard against partial downloads. It would be pretty easy to filter and resumed the paused ones, anyway.

Also for usenet downloads: my Sabnzbd puts everything into folders. I think this is the default behavior and I don't even know if it can be changed. This will likely cause issues with usenet->torrent matching. I think my next move in terms of a feature will be looking inside of folders for matchable files. This would give better usenet matching and more importantly from my perspective it can allow users to scan the organized libraries that sonarr/radarr produce.

mmgoodnow:
> my Sabnzbd puts everything into folders
what does the folder structure look like?

akesher123:
> > my Sabnzbd puts everything into folders
> > what does the folder structure look like?

/Usenet Downloads
|___ /download 1
|     |____item1.mkv
|     |____item2.mkv
|
|___ /download 2
|     |____item1.mkv
|
|___/ download 3
      |____ item1.mkv
      
No matter what, even if you're only downloading a single item, it will put it in a parent dir. I'm actually not sure if nested dirs are possible or common, someone else would have to comment on that. For some reason my sab is deleting my old downloads anyway lol

src/pipeline.ts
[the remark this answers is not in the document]
akesher123: Actually you're right, this one doesn't make sense. I had my directory levels wrong in my head.

src/pipeline.ts
```diff
@@ -186,17 +191,27 @@ export async function checkNewCandidateMatch(
...
 	if (Array.isArray(torrents)) {
 		const searcheeResults = await Promise.all(
-			torrents.map(createSearcheeFromTorrentFile)
+			torrents.map(createSearcheeFromTorrentFile) //also create searchee from path
 		);
 		parsedTorrents = searcheeResults.filter(ok);
+	} else if (dataDirs.length > 0) {
+		var fullPaths: string[] = [];
+		dataDirs.forEach(dataDir => {
+            const allPaths = getFilePathsFromPath(dataDir, [], 0, 4);
```
mmgoodnow, line 204: i'm worried about the 4 here. I feel like nearly all of the matches are going to be in the 0 depth or the 1 depth - most releases aren't really nested, right?

```
datadir/
  cheers.s01e01/
    cheers.s01e01.mkv
  mash.s01/
    mash.s01e01.mkv
    mash.s01e02.mkv
    mash.s01e03.mkv
    Sample/
      mash.s01e01.sample.mkv
```

for this tree, I personally would only want searchees for the first two. I could easily see an argument for the first 6, but definitely not the last 2. 
1. `mash.s01`
2. `cheers.s01e01.mkv`
3. `mash.s01e01.mkv`
4. `mash.s01e02.mkv`
5. `mash.s01e03.mkv`
6. `cheers.s01e01`
7. `Sample`
8. `mash.s01e01.sample.mkv`

I also would still like to apply some favoritism to single-file torrents if we can - I've gone on record before as saying that I want to use my power, however small, to encourage single-file torrents and discourage multi-file torrents with extra crap in them (nfos, samples, etc). This links back up to my earlier question about `MATCH_EXCEPT_PARENT_DIR` - maybe we can support the first 2 by default, and the first 5 on risky mode if the last one (6) doesn't really have a good hit rate in practice. If it does hit regularly, then maybe not (edit: i guess I sort of forgot about packs, although I vaguely remember hearing that maybe packs aren't a thing in usenet?. 

You'll notice, also, that my priority list doesn't really relate linearly to `maxDepth` - the first one is a 0 depth, and the second one is a 1 depth. I think we can probably apply heuristics to be smart about this, but that can probably come later.

src/pipeline.ts
```diff
@@ -186,17 +191,27 @@ export async function checkNewCandidateMatch(
...
+            fullPaths = fullPaths.concat(allPaths.filter(file => 
+                fs.statSync(file).isDirectory() || 
+                ([".mkv", ".avi", ".mp4", ".ts", ".flac", ".mp3"].includes(path.extname(file))  // try to avoid searching for a RARed pieces
+                && fs.statSync(file).size > 100000000))); //100 MB to start)
```
mmgoodnow, line 208:
```suggestion
                && fs.statSync(file).size > 100000000))); // 100 MB to start
```

mmgoodnow (commented):
This PR is coming along really well! Are you still planning on working on Daemon mode? I would like to support that for v1.

src/constants.ts
```diff
@@ -28,6 +32,7 @@ export type ActionResult = InjectionResult | SaveResult;
 
 export enum Decision {
 	MATCH = "MATCH",
+	MATCH_EXCEPT_PARENT_DIR = "MATCH_EXCEPT_PARENT_DIR",
```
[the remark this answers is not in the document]
akesher123: Very common, at least with my trackers. Running on my library it's about 60/40 single file, multi-file with one file. It's not that much of an issue in terms of stability (so far), but it needs more testing.

src/pipeline.ts
akesher123: 4 is probably a little deep yes, I think my deepest match was 3 (individual episodes in season pack). 

This also calls into discussion what safe/risky should actually be. I think of them to be used for matching, not searching. For searching, the only thing to be concerned about is API hits, and IMO anyone concerned about api hits probably shouldn't be using a tool that can generate hundreds of queries and thousands of grabs on a mid-sized library.  I think just setting the max depth to 3 is a reasonable enough heuristic to get a high match rate.

In a way, single file torrents are still favored since they're more likely to be a perfect match. TBH I'm not sure if I've enabled renaming of single file torrents, but I need to do that if I haven't. I totally agree on the philosophy of single file torrents. To be fair, this still doesn't match torrents with .nfos and samples and whatnot, only single file in a directory. Of course, I think this practice should be ditched as well, but according to my results it's still a major use-case. Granted, I'm still mainly cross-seeding to places like ipt and tl, and not upper tier trackers since I'm not in them. I'm not sure how common these practices are in a tracker with better organization like BHD.

src/clients/QBittorrent.ts
```diff
@@ -340,4 +373,53 @@ export default class QBittorrent implements TorrentClient {
 			return InjectionResult.FAILURE;
 		}
 	}
+	async rename(
```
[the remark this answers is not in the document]
mmgoodnow: I feel like the differences could be abstracted away inside some sort of `calculateDataDir` type method while keeping the top-level API of `TorrentClient` simple.

src/pipeline.ts
mmgoodnow: > I think my deepest match was 3 (individual episodes in season pack).

ooc, what makes that 3 levels deep?

> For searching, the only thing to be concerned about is API hits, and IMO anyone concerned about api hits probably shouldn't be using a tool that can generate hundreds of queries and thousands of grabs on a mid-sized library

With all due respect, imo this is a lazy answer. As the developers of this API-happy tool, we especially have a duty to minimize our API footprint as much as we can, because the effects of our decisions here multiply by the size of the userbase. If that comes at the cost of a small number of matches, then it's worth it IMO.

akesher123: >ooc, what makes that 3 levels deep?
top level -> season -> per episode folder -> episode.mkv

While an uncommon structure, it unfortunately exists. 

>With all due respect, imo this is a lazy answer. As the developers of this API-happy tool, we especially have a duty to minimize our API footprint as much as we can, because the effects of our decisions here multiply by the size of the userbase. If that comes at the cost of a small number of matches, then it's worth it IMO.

Good points, I agree. 

I think it comes down to there being two safe/risky concepts: safe/risky in matching, which is what the current flag dictates, and safe/risky in api hits. The best way to limit risky api hits is by searching less of the files you come across and not going as deep in directories. Like I said, I believe my deepest matches were 3 levels deep. IMO these probably aren't worth it. I do think it makes sense to go 2 deep (single file inside folder inside parent dir) because that's how radarr libraries are laid out, which I think is a major use case. 

That said, I don't think we can simultaneously have favoritism for single file torrents while also having good results on a torrent download folder. Running on only 1 layer deep is essentially what it does now but on data instead of torrents. If the download was a folder, then it's searched as a folder. To prefer matching single-file torrents, we have to drill down to the file itself, which is frequently on layer 2.

Right now, it will search a folder that contains only 1 torrent file. I think this can easily be gotten rid of to reduce API hits by half with no losses (when running on a radarr library). We can also kick up the minimum size a bit to reduce further (2gb perhaps?), and limit the depth to 2 or 3 (I need to experiment on what this number means relative to the search dir) to reduce further. I would say this is sufficient to cover the bases without the need for a safe/risky heuristic setting for searching. With those changes, running on a data dir will probably not produce that many more API hits than if that equivalent dir were all torrents in a client.

Anyway, with your concerns about MATCH_EXCEPT_PARENT_DIR, like I said on my trackers it's extremely common,

mmgoodnow: Very well reasoned, this all sounds good. Sorry I haven't had a lot of time to give this PR attention lately; I'm in a bit of a cross-seed hibernation cycle (I do most of my development in the winter). I do still plan on working on daemon mode support for this, but it may be a little while.

PR #321, opened 2022-09-08
Fix zero-length-path bug on the other side of the comparison

### LINKED ISSUE #320: cross-seeded directory structure includes top-level directory named "_"
As per discord discussion https://discord.com/channels/880949701845872672/880949702349164556/1017187328734789692, cross-seed injected a torrent but included a top level folder, "_", containing the files causing the client to indicate missing files

```
root@foxden:~# docker exec -it cross-seed cross-seed --version
Configuration file already exists.
4.3.8
```

mmgoodnow:
fixes #320


mmgoodnow:
@thereclusivefox would you mind testing this branch out? expected behavior is that cross-seed finds and injects the false positive torrent on `master` or `latest`, but does not find or inject the false positive on `zero-length-path-bug`

TheReclusiveFox:
@mmgoodnow What's the best/easiest way to check out the branch? I'm using docker in my current setup and looks like CI failed to push new image

mmgoodnow:
Looks like they were transient failures. I reran the builds and they worked, so you should be all set to pull from the branch tag.

TheReclusiveFox:
So it looks like the same behaviour is occurring using the image from this branch. The injected torrent has the extra `_` parent directory

mmgoodnow:
Hmm, can you try running cross-seed clear-cache and then try again? Make sure to remove the injected torrent as well.

TheReclusiveFox:
How long does `cross-seed clear-cache` take? I let it run for over 2 hours but it never exited -- it also didn't show any logging when running the command except:
```
root@foxden:~# docker exec -it cross-seed cross-seed clear-cache
Configuration file already exists.
```

mmgoodnow:
hm, that's probably a bug. it should be near-instantaneous.

PR #358, opened 2023-01-15
Support hardlinks on data based matching

akesher123:
It's a little rough around the edges in code but this is currently working. I'm mostly just publishing the PR for visibility reasons, I still have cleanup to do.

For many people this will also depend on a change in qbm. Discussion here: https://github.com/StuffAnThings/qbit_manage/issues/208

Known issues:
- [ ] Crashes if hardlinkDir not already existing
- [ ] Needs to clean up hardlinkDir after run


src/cmd.ts
```diff
@@ -73,6 +73,10 @@ function createCommandWithSharedOptions(name, description) {
 			"Safe will only download torrents with perfect matches. Risky will allow for renames and more matches, but might cause false positives",
 			fallback(fileConfig.dataMode, "safe")
 		)
+		.option(
+			"-h, --hardlink-dir <dir>",
```
mmgoodnow, line 77:
```suggestion
			"--link-dirs <dirs...>",
```

maybe i'm getting ahead of myself here, but given that hardlinks don't work across mount points, i could see us needing at most one link dir per data dir, if each data dir is on a different volume. it doesn't feel great but it could be an implicit mapping where order matters, and then it'd be at least supported. but maybe that's a feature for later. 

also, we can always add short params later but taking them away is hard so let's keep it long for now

src/decide.ts
```diff
@@ -121,25 +121,61 @@ async function assessCandidateHelper(
...
+						mkdirSync(path.join(hardlinkDir, candidateParentDir));
+					}
+					correctedHardlinkDir = path.join(hardlinkDir, candidateParentDir);
+				}
+				symlinkFile(
+					path.dirname(searchee.path),
+					correctedHardlinkDir,
+					path.basename(searchee.path),
+					path.basename(candidateMeta.files[0].path)
+				);
```
mmgoodnow, line 151: i think this code is less related to decisions and more related to action/injection. maybe the top of performAction would be a good place for it? and we could add an argument to `inject` specifying the parent dir

mmgoodnow (commented):
just a quick pass here, haven't fully looked at the linking code. but the concept feels pretty solid

akesher123:
> do you think we need to namespace entries in the link dir?

What do you mean?

src/cmd.ts
akesher123: +1 on long name, however initially I think supporting one dir is preferable. Getting into mapping is weird especially if you don't have an even number of source and dest dirs. I suppose there could be a validation that there is either one dest dir, or as many as there are sources, but imo that should be more formalized. Too easy for users to create errors even with saying there's an implicit hidden mapping and the dir order they provides matters.

This is one of those things that I think would be better suited in a dummy proof UI when that comes down the pipeline.

mmgoodnow: yeah, agreed hard on the web ui. Trying to fit all the configuration data into a format that's specifiable on the command line sucks.

> initially I think supporting one dir is preferable

ok, sounds good.

mmgoodnow:
something like putting everything inside a Subfolder with an infohash in its name. Otherwise I could see things overwriting each other

src/cmd.ts
```diff
@@ -78,6 +78,31 @@ function createCommandWithSharedOptions(name, description) {
...
+			"Skips torrent recheck upon adding to QBittorrent",
+			fallback(fileConfig.skipRecheck, false)
+		)
+		.option(
+			"--max-data-depth <depth>",
+			"Max depth to look for searchees in dataDirs",
+			fallback(fileConfig.maxDataDepth, 2)
+		)
+		.option(
+			"--single-search",
```
akesher123, line 102: Now that I've seen daemon mode is a different path, this option is basically pointless. However keeping this as a note that the function for generating searchees from data needs to have a parameter added to accomplish the same thing for use during daemon mode.

PR #405, opened 2023-04-16
Allow specifying the path to config file via an environment variable

winterqt:
Similar to how the configuration directory is specificable via the `CONFIG_DIR` environment variable, this makes the configuration file path itself configurable, in case it lies outside of the configuration directory.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -75,7 +75,7 @@ export async function getFileConfig(): Promise<FileConfig> {
 	const configPath = path.join(appDir(), "config.js");
 
 	try {
-		return (await import(pathToFileURL(configPath).toString())).default;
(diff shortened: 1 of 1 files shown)


mmgoodnow:
what is your use case for this? I don't want to lose the invariant that the config file is anywhere other than inside the CONFIG_DIR.

winterqt:
This allows usage of systemd's `LoadCredential` feature: https://systemd.io/CREDENTIALS/

mmgoodnow:
i'm still not sure I understand. Do you want to supply a path to the config file as a credential?

mmgoodnow:
Okay, well to be honest i'm not inclined to upstream this as the use-case is a little too platform specific for me to want to maintain support for it, but if it works for you as a fork, I'd recommend that.

PR #417, opened 2023-05-15
Chore: Improvement on config templates

fabricionaweb:
Just small touches that I think can look better

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -4,14 +4,15 @@
 module.exports = {
 	/**
 	 * Pause at least this much in between each search. Higher is safer.
-	 * 	It is not recommended to set this to less than 2 seconds.
+	 * It is not recommended to set this to less than 2 seconds.
 	 */
 	delay: 10,
+
 	/**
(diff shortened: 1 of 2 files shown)


src/config.template.cjs
```diff
@@ -4,14 +4,15 @@
...
 	 */
 	delay: 10,
+
 	/**
-	 * List of Torznab urls.
-	 * The path should end in /api
+	 * List of Torznab URLs.
 	 * For Jackett, click "Copy RSS feed"
-	 * For Prowlarr, click (i) and copy the Torznab Url, then append "?apikey=YOUR_PROWLARR_API_KEY"
+	 * For Prowlarr, click on the indexer name and copy the Torznab Url, then append "?apikey=YOUR_PROWLARR_API_KEY"
```
fabricionaweb, line 14: Prowlarr has changed, it does not have the `(i)` anymore. Instead you need to click at indexer name (first column)

mmgoodnow:
thanks for the cleanup!

PR #418, opened 2023-05-15
Feature: Search limit

### LINKED ISSUE #432: excludeRecentSearch API usage is too bursty
When doing the initial search against torrents or data, everything will have the same search date.  Therefore when excludeRecentSearch is used, all torrents/data will be available for search again at the same time. This will lead to feast vs famine use of API. 

Add a random amount of time based on the excludeRecent duration (for example, if exclude recent is 14 days,  add +/- 7 days to each torrent randomly, to spread them out)

This would also enable more frequent searches (hourly? 2x a day?)  since any particular search will only be hitting a small subset of the files.

fabricionaweb:
The idea here is to limit the number of queries to be made between each running.

The current feature `excludeRecentSearch` can already filter off some results, so combined with a limit feature we can have more flexibility.

Example of my usage:
- My library has 440 items
- `excludeRecentSearch` set to `5 days`
- `searchLimit` set to `100`
- `searchCadence` set to `1 day`

Every day the search runs, it picks only the 100 that have not be done before

The user needs to pay attemption of their values or this can keep items out of the search, for example, if the `excludeRecentSearch` is too short for the limit, it will be start to pick items again...

fixes #432

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -334,6 +334,12 @@ createCommandWithSharedOptions("search", "Search for cross-seeds")
 			"torrent files separated by spaces"
 		).hideHelp()
 	)
+	.addOption(
+		new Option(
+			"--search-limit <number>",
+			"set the limit of searches for the running"
+		)
+	)
 	.action(async (options) => {
 		try {
 			const runtimeConfig = processOptions(options);
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -207,6 +207,12 @@ function createCommandWithSharedOptions(name, description) {
 			"--search-timeout <timeout>",
 			"Timeout for unresponsive searches",
 			fallback(fileConfig.searchTimeout, "30 seconds")
+		)
+		.option(
+			"--search-limit <number>",
+			"The number of searches before stops",
+			parseFloat,
+			fallback(fileConfig.searchLimit, 0)
 		);
 }
 
@@ -333,12 +339,6 @@ createCommandWithSharedOptions("search", "Search for cross-seeds")
 			"--torrents <torrents...>",
 			"torrent files separated by spaces"
 		).hideHelp()
-	)
-	.addOption(
-		new Option(
-			"--search-limit <number>",
-			"set the limit of searches for the running"
-		)
 	)
 	.action(async (options) => {
 		try {
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -211,7 +211,7 @@ function createCommandWithSharedOptions(name, description) {
 		.option(
 			"--search-limit <number>",
 			"The number of searches before stops",
-			parseFloat,
+			parseInt,
 			fallback(fileConfig.searchLimit, 0)
 		);
 }
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -215,4 +215,11 @@ module.exports = {
 	 * null
 	 */
 	searchTimeout: undefined,
+
+	/**
+	 * The number of searches to be done before it stop.
+	 * Combine this with "excludeRecentSearch" and "searchCadence" for better results.
+	 * Default is no limit.
+	 */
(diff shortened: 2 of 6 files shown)


mmgoodnow:
i actually removed a `limit` flag when I added periodic searches, because it felt like it didn't really make sense. However I do see this use case and have wished for something similar. 

what if we had a way for cross-seed to automatically balance?

example:

searchCadence 1 day, excludeRecentSearch is 6 months (~180 days), we're running for the first time. 1000 torrents are eligible to run. 

therefore, we do one 180th of the torrents, plus some number for good measure.

yammes08:
> i actually removed a `limit` flag when I added periodic searches, because it felt like it didn't really make sense. However I do see this use case and have wished for something similar.
> 
> what if we had a way for cross-seed to automatically balance?
> 
> example:
> 
> searchCadence 1 day, excludeRecentSearch is 6 months (~180 days), we're running for the first time. 1000 torrents are eligible to run.
> 
> therefore, we do one 180th of the torrents, plus some number for good measure.
> 
> one problem with this approach would be that as we progress through the 6 months, less and less torrents will be eligible to run. when there are 89 torrents left, a 180th of the eligible torrents would round to 0. so the first time, we'd do like 5, and towards the end we'd be doing less. So it's not fully balanced.
> 
> Another approach: we count the total number of torrents, **ignoring excludeRecentSearch**, then divide that number by 180. Then, we run that number of torrents today (rounding up). this would probably be the most balanced? and it does seem feasible…

I was chatting to fab about this - that last idea sounds like it would work. So, you could just have an `enableSearchLimit` flag, `true` or `false`, and if `true` it does the calculation you describe?

mmgoodnow:
> So, you could just have an enableSearchLimit flag, true or false, and if true it does the calculation you describe?

yeah i think so.

src/cmd.ts
```diff
@@ -207,6 +207,12 @@ function createCommandWithSharedOptions(name, description) {
 			"--search-timeout <timeout>",
 			"Timeout for unresponsive searches",
 			fallback(fileConfig.searchTimeout, "30 seconds")
+		)
+		.option(
+			"--search-limit <number>",
+			"The number of searches before stops",
+			parseFloat,
```
mmgoodnow, line 214:
```suggestion
			parseInt,
```

src/config.template.docker.cjs
```diff
@@ -219,4 +219,11 @@ module.exports = {
 	 * null
 	 */
 	searchTimeout: undefined,
+
+	/**
+	 * The number of searches to be done before stop.
+	 * Combine this with "excludeRecentSearch" and "searchCadence" for better results.
```
mmgoodnow, line 225:
```suggestion
	 * Combine this with "excludeRecentSearch" and "searchCadence" to smooth long-term API usage patterns.
```

src/config.template.docker.cjs
```diff
@@ -219,4 +219,11 @@ module.exports = {
 	 * null
 	 */
 	searchTimeout: undefined,
+
+	/**
+	 * The number of searches to be done before stop.
```
mmgoodnow, line 224:
```suggestion
	 * The number of searches to be done before stop.
```

src/config.template.cjs
```diff
@@ -215,4 +215,11 @@ module.exports = {
 	 * null
 	 */
 	searchTimeout: undefined,
+
+	/**
+	 * The number of searches to be done before it stop.
+	 * Combine this with "excludeRecentSearch" and "searchCadence" for better results.
```
mmgoodnow, line 221:
```suggestion
	 * Combine this with "excludeRecentSearch" and "searchCadence" to smooth long-term API usage patterns.
```

mmgoodnow (approved):

mmgoodnow:
Thank you for the contribution!

PR #422, opened 2023-05-16
Dont wait for delay when theres no indexers to search

fabricionaweb:
Currently, the pipeline is responsable to wait for the config delay. 

But we dont want to wait for it, if theres no search to be done.
I tried to move the delay to another place but it doesnt look good.

The solution I found here was move the filter out of the `searchTorznab` method, than we can call it early and pass it forward via parameters.

I totally agree that It does not look too good to overload the methods argument, but yet I could not present better solution. 

Thats just what I can present today, just trying to help.

### DIFF
--- a/src/indexers.ts
+++ b/src/indexers.ts
@@ -51,7 +51,7 @@ export async function getEnabledIndexers() {
 export async function filterIndexersByTimestamp(
 	name: string,
 	excludeRecentSearch: number,
-	excludeOlder: number,
+	excludeOlder: number
 ) {
 	const enabledIndexers = await getEnabledIndexers();
 
--- a/src/indexers.ts
+++ b/src/indexers.ts
@@ -1,6 +1,6 @@
 import { db } from "./db.js";
 import { Label, logger } from "./logger.js";
-import { humanReadable } from "./utils.js";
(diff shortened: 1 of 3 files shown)


mmgoodnow:
thank you for the contribution! 

fwiw, large numbers of `Searching 0 indexers for …` logs is definitely a bug, and those should really not even be getting through the prefilter stage. That said, it will always have the possibility of happening due to the webhook initiating a search during a long-lived bulk scan. 

I'll take a look at this in the coming week.

mmgoodnow:
hopefully #423 fixes this bug at the source

mmgoodnow:
Okay, given that 423 does not fix it at the source, I'll try to get this merged. I think it's because trackers are getting taken out of the rotation due to rate limiting mid-run, which seems valid and probably unavoidable? So this mitigation is the next best thing we can do.

mmgoodnow:
rather than moving `indexersToUse` into pipeline, my plan is to have `searchTorznab` report how many indexers it searched in its return value. if it searched 0, then we should be safe to skip the delay

mmgoodnow:
closing this in favor of #423

PR #423, opened 2023-05-21
filter indexers the same way in both situations

### LINKED ISSUE #431: delay delays, even if all trackers skipped
Even if no trackers were actually searched, due to timeout filtering, the delay still happens, significantly slowing down scheduled searches.

mmgoodnow:
There's a bug where people are seeing `Searching 0 indexers for …` logs for every torrent in their list. My guess is that this is because the timestamp prefiltering is using different rules for including indexers in the calculations than the individual timestamp calculations in the main loop. This changes the prefiltering to use the same `getEnabledIndexers()` method for including indexers. 

fixes #431

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -56,10 +56,15 @@ interface AssessmentWithTracker {
 	tracker: string;
 }
 
+interface FoundOnOtherSites {
+	searchedIndexers: number;
+	matches: number;
+}
+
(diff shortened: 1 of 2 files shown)


mmgoodnow:
@fabricionaweb interested in taking a look at this PR? I think the second commit does what we want (skips the sleep in the `findMatchesBatch` loop when we searched 0 indexers

PR #434, opened 2023-06-10
chore: tag image with semver compliant container tag

onedr0p:
Hi 👋🏼 

I don't see a reason to have the `version-` prefix on this tag but curious to know if there was a reason for it :)

This changes also makes it easier for automation tools like Renovate that expect semver to automatically sort and filter the tag correctly.

Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/cross-seed
             ${{ secrets.DOCKERHUB_USERNAME }}/cross-seed
           tags: |
-            type=semver,pattern=version-{{version}}
+            type=semver,pattern={{version}}
             type=ref,event=branch
             type=ref,event=pr
       -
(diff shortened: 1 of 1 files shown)


mmgoodnow:
can we keep the old version but add the new version so it's backwards compatible?

onedr0p:
Sure... but what's the need on having both? Having one tag keeps things simple.

mmgoodnow:
Backwards compatibility.

onedr0p:
Yeah... I read that, but I don't understand. How does keeping the old tag format keep backwards compatibility?

mmgoodnow:
People already using `version-` tags can continue to do so without having to change anything.

mmgoodnow:
e.g. when they're updating their container version, and then they change to `version-6` but it doesn't exist. so then they need to debug

onedr0p:
I read the metadata action docs but I'm not seeing a way to have both. Honestly I don't see why this matters but I'll keep looking into it and report back. I just added support for tags which also should help people who just want to pinned to the latest major, minor or full semver tags.

```
1.2.3
1.2
1
```

onedr0p:
Nevermind the latest commit should do it.

mmgoodnow:
thanks for adding the major version tags! that'll be a nice QOL improvement 🙂

PR #435, opened 2023-06-10
(fix) enhance logger cleansing of notification urls

bakerboy448:
cleanse notifiarr url

current console output
`  notificationWebhookUrl: 'https://notifiarr.com/api/v1/notification/crossSeed/[REDACTED]-1234-568-9d05-1fghj1i678uik',`

this changes from /w to 0-9-a-zA-Z_- which covers `-` as well as `/w` 

also aligns regex with the torznab logic for consistency


mmgoodnow:
thanks for the fix!

PR #447, opened 2023-07-03
Search torrentDir before dataDirs

ColinHebert:
Run the search on TorrentDir first as it will look for exact matches as a first order.

Currently cross-seed will process the dataDirs before considering a lookup of the torrentDir.
In practice, searches using the torrent dir allows to find exact matches to cross-seed. It requires fewer operations (no hardlinks) and uses the same category as the original torrent. It is overall "safer".

Because of this, it would be preferable to process the entire content of the torrentDir first, and only then do a more thorough lookup which may or may not get hits.


mmgoodnow (approved):

mmgoodnow:
thanks for the contribution!

Issue #424, opened 2023-05-22
Cross-seed crashes when a file present at the beginning of a search is removed


ColinHebert:
Cross-seed crashes when a file present at the beginning of a search scan with data-based matching no longer exists when the file gets processed:

```
info: [torznab] Searching 11 indexers for TVShow.265-NTb.mkv

node:fs:1583
  handleErrorFromBinding(ctx);
  ^
Error: ENOENT: no such file or directory, stat '/data/torrents/completed/others/TVShow.265-NTb.mkv'
    at statSync (node:fs:1583:3)
    at assessCandidateHelper (file:///usr/src/cross-seed/dist/decide.js:94:10)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async assessAndSaveResults (file:///usr/src/cross-seed/dist/decide.js:110:24)
    at async assessCandidateCaching (file:///usr/src/cross-seed/dist/decide.js:151:22)
    at async assessEach (file:///usr/src/cross-seed/dist/pipeline.js:21:21)
    at async Promise.all (index 36)
    at async findOnOtherSites (file:///usr/src/cross-seed/dist/pipeline.js:42:22)
    at async Promise.all (index 0)
    at async findMatchesBatch (file:///usr/src/cross-seed/dist/pipeline.js:74:28) {

  errno: -2,
  syscall: 'stat',
  code: 'ENOENT',
  path: '/data/torrents/completed/others/TVShow.265-NTb.mkv'
}
```

Seems to come from https://github.com/cross-seed/cross-seed/blob/master/src/decide.ts#L133-L137

I can see two ways of handling it:
- catching the exception (which probably should be done no matter what to avoid a complete crash in case something odd happens)
- checking whether the file/torrent/item to be processed still exists when getting to it when dealing with batches

PR #448, opened 2023-07-03
Try/Catch exception when a single sample search fails

ColinHebert:
Avoid an exception for a single search breaking the entire search batch.

Addresses concerns from #424

When running a batch search, treat each search as a possible failing element, catching the exception and proceeding onto the next item in the batch.
It doesn't fix entirely the issue raised in #424, but it avoid the batch search to crash entirely if during the batch processing an item gets removed.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -147,21 +147,26 @@ async function findMatchesBatch(
 
 	let totalFound = 0;
 	for (const [i, sample] of samples.entries()) {
-		const sleep = new Promise((r) => setTimeout(r, delay * 1000));
+		try {
+			const sleep = new Promise((r) => setTimeout(r, delay * 1000));
 
-		const progress = chalk.blue(`[${i + 1}/${samples.length}]`);
-		const name = stripExtension(sample.name);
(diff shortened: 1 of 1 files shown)


src/pipeline.ts
```diff
@@ -142,15 +142,20 @@ async function findMatchesBatch(
...
+	
+			const progress = chalk.blue(`[${i + 1}/${samples.length}]`);
+			const name = stripExtension(sample.name);
+			logger.info("%s %s %s", progress, chalk.dim("Searching for"), name);
+	
+			const numFoundPromise = findOnOtherSites(sample, hashesToExclude);
+			const [numFound] = await Promise.all([numFoundPromise, sleep]);
+			totalFound += numFound;
+		} catch (e) {
+			logger.error(`error searching for ${name}`);
```
ColinHebert, line 156: Should be `sample.name`

mmgoodnow:
can you rebase this on master?

ColinHebert:
> can you rebase this on master?

Done

mmgoodnow:
thanks!

PR #455, opened 2023-07-16
Optimized image size

hummingbirdy2:
The current Docker image (uncompressed) is around 1GB.
To optimize its size, I took the following actions:
- Use an Alpine image `node:18-alpine` instead of the default Debian.
- Use the Build Stage to avoid compile dependencies in the Production Stage.
- I also took the opportunity to update nodejs to its latest LTS version.

With this PR, is now around 300MB for my arm build.
I lack knowledge of nodejs, there may still be room for improvement.

My first tests seem to show that this works well. But a little more in-depth testing would be welcome.


mmgoodnow:
Thanks for the contribution! I haven't used build and production stages before - what files are we removing? by splitting the stages though?

does the alpine image have curl inside it? that's pretty much the only thing I care about re "included batteries".

Dockerfile
```diff
@@ -1,14 +1,21 @@
...
+FROM node:18-alpine
+WORKDIR /usr/src/cross-seed
+COPY --from=build-stage /usr/src/cross-seed .
 RUN npm link
+ENV CONFIG_DIR=/config
+ENV DOCKER_ENV=true
 EXPOSE 2468
 WORKDIR /config
-ENTRYPOINT ["cross-seed"]
+ENTRYPOINT ["cross-seed"]
```
mmgoodnow, line 21: nit: end all files with LF

```suggestion
ENTRYPOINT ["cross-seed"]

```

mmgoodnow:
One thing that could possibly reduce the size even more - when we do the `npm ci` it installs the devDependencies (namely the typescript compiler) which we probably don't need afterwards. Maybe we can delete the dev dependencies after that and only ship the runtime deps

mmgoodnow:
looks like `npm prune --production` could do it https://docs.npmjs.com/cli/v9/commands/npm-prune

mmgoodnow:
awesome! I'll try to find some time this week to test this out and merge

Issue #365, opened 2023-02-02
Pull searchees from qBittorrent API (BT_Backup gone)


KyleSanderson:
https://github.com/qbittorrent/qBittorrent/discussions/16768

KyleSanderson:
BT_Backup is no longer in-use, and they're using SQLite (torrents.db) with WAL to store session state.

Slikkster:
Since you are using the qbittorrent API to inject the torrents, why not also use it for checking for existing torrents instead of BT_backup? Then it wouldn't matter how qbittorrent stores information on the disk.

dynizzle:
I setup cross-seed on Unraid for the first time today and I think I'm running into this as well. Whenever I add torrents it doesn't even try to do a lookup but by adding the [run external program on completion](https://www.cross-seed.org/docs/basics/daemon#qbittorrent) for new torrents that are added, it works. Doesn't really help for new releases that don't exist on all my trackers yet.

What's weird is I still see new torrents being added in BT_Backup folder so I don't know if that's the issue or not. I'm thinking of just downgrading qBittorrent to <= 4.3 to see if that fixes it.

mmgoodnow:
That's WAD then. It doesn't react to new torrents automatically

dynizzle:
Ah OK. So with Daemon mode, is any way to force a full search of all current torrents without scripting out all the API calls manually?

Slikkster:
Yes, just run the 'cross-seed search' command. Even if the daemon is already running, it will run the search (at least it did for me). If you're using docker, do 'docker exec -it <container> cross-seed search'

mmgoodnow:
please take this conversation to Discussions or Discord.

KyleSanderson:
They've added a new endpoint, `api/v2/torrents/export`, that can export the full torrent file from the client. So what could probably be done on the cross-seed side of things, is scan the list, then export the torrent files it doesn't have based on the infohash and store it in the cache.

KyleSanderson:
Wrote this dumb thing https://github.com/KyleSanderson/sessionerr as a workaround

onedr0p:
This would be great if cross seed was updated to support qb with sqlite. SQLite with WAL has significantly improve startup and loading times for me. I just went thru trying to set up cross-seed but to be hit by this issue, if y'all need someone to help test I'm here. 👍

MrMarble:
Is there any update on this? It has been on ready for a while now

KyleSanderson:
[...]

PR #471, opened 2023-08-29
feat(clients/qbit): pull searchees using qbit api - sqlite/4.6.x compatibility

mmgoodnow:
#365


onedr0p:
I'd like to test this with you @mmgoodnow. It appears like I would just need to set my qbittorrent categories to search?

```js
qbittorrentCategories: [
	"radarr",
    "sonarr"
]
```

mmgoodnow:
to be honest it's not really in a workable state atm

Issue #453, opened 2023-07-08
Search Limit shouldn't include 0 indexer searches


removinggoose:
Hi, I have not been able to properly utilise the search limit as every time a search occurs, it searches for the limit amount (let's say 100) and if the first 100 were already searched for in the past day, it still includes them in the 100 searches. This means that all of the 100 torrents are taken up by items that searched 0 indexers and therefore no actual searches occurred. It would be useful if any item that searches 0 indexers, be excluded from the search limitations.

removinggoose:
any help on this please?

diligent-platypus:
I'm running into the same issue. I've set the `searchLimit=10`, `searchCadence=5min` and `excludeRecentSearch=1w` but it continuously checks the first 10 which, after the first run, are filtered by timestamp.

Any input on this would be appreciated

removinggoose:
> I'm running into the same issue. I've set the `searchLimit=10`, `searchCadence=5min` and `excludeRecentSearch=1w` but it continuously checks the first 10 which, after the first run, are filtered by timestamp.
> 
> Any input on this would be appreciated

Perhaps you could also post your config (minus sensitive info). Then it's already here when/if someone gets around to looking at this.

removinggoose:
any update or info on this please?

mmgoodnow:
No updates

mindset-tk:
Hello. I believe I've fixed this issue with some simple tweaks in the PR I opened earlier today. Thank you for your consideration.

Issue #402, opened 2023-04-16
revisit entries when torznab endpoint is down


KyleSanderson:
```
info: [torznab] Searching 1 indexers for Doppelhaushaelfte.S01.German
warn: Failed to reach http://jackett.jackett:9117/api/v2.0/indexers/core/results/torznab/api
info: [1952/4582] Searching for Tell.Me.Something.S01.1080p.WEB-DL.DDP5.1.H.264-N3RV0US
info: [torznab] Searching 1 indexers for Tell.Me.Something.S01.1080p.WEB-DL.DDP5.1.H.264-N3RV0US
warn: Failed to reach http://jackett.jackett:9117/api/v2.0/indexers/core/results/torznab/api
```

It's hours/days of work to have it scan again. A feature to append the search in the queue as opposed to skipping them when the torznab provider is down would be helpful.

mmgoodnow:
The codebase is not really set up for this, but I'm planning to change the architecture to be event-based at some point, which should make this kind of thing a lot easier

mindset-tk:
fyi @KyleSanderson that pull request should partially resolve your issue. You will have to run search again instead of appending, but on the second search it will only search indexers that haven't been searched within the search timeout you've set - assuming you've set one.

KyleSanderson:
Thank you for letting me know.

PR #479, opened 2023-09-07
Fix search limit config item

mindset-tk:
The SQL query for the filterTimestamps function was malformed and would return the same (too high, low) timestamp every time, since it didn't filter for torrent name, and because max and min were reversed. It would never filter any torrent unless every single torrent had an extant entry in the db for every indexer.

Fixes cross-seed/cross-seed#453

### DIFF
--- a/src/preFilter.ts
+++ b/src/preFilter.ts
@@ -86,7 +86,6 @@ export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
 			last_searched_all: db.raw("coalesce(timestamp.last_searched, 0)"),
 		})
 		.first();
-		
 	const { first_searched_all, last_searched_all } = timestampDataSql;
 	function logReason(reason) {
 		logger.verbose({
--- a/src/preFilter.ts
+++ b/src/preFilter.ts
@@ -63,27 +63,30 @@ export function filterDupes(searchees: Searchee[]): Searchee[] {
 export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
 	const { excludeOlder, excludeRecentSearch } = getRuntimeConfig();
 	const enabledIndexers = await getEnabledIndexers();
-	const timestampDataSql = await db("searchee")
+  const timestampDataSql = await db("searchee")
 		// @ts-expect-error crossJoin supports string
 		.crossJoin("indexer")
 		.leftOuterJoin("timestamp", {
 			"timestamp.indexer_id": "indexer.id",
 			"timestamp.searchee_id": "searchee.id",
 		})
+    .where("searchee.name",
+    searchee.name
+    )
 		.whereIn(
 			"indexer.id",
 			enabledIndexers.map((i) => i.id)
 		)
-		.max({
+		.min({
 			first_searched_all: db.raw(
 				"coalesce(timestamp.first_searched, 9223372036854775807)"
 			),
 		})
-		.min({
+		.max({
 			last_searched_all: db.raw("coalesce(timestamp.last_searched, 0)"),
 		})
 		.first();
-
+    
 	const { first_searched_all, last_searched_all } = timestampDataSql;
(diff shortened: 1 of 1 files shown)


mindset-tk:
Hey, one other thing, I found that leaving line 85 on prefilter.ts as "min" will allow the tool to revisit indexers that were previously filtered by timestamps, see cross-seed/cross-seed/#402 - the code for performing the search is already sensitive to if an indexer was recently searched and will filter at that point.

src/preFilter.ts
```diff
@@ -83,7 +86,7 @@ export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
 			last_searched_all: db.raw("coalesce(timestamp.last_searched, 0)"),
 		})
 		.first();
-
+
```
bakerboy448, line 89: Erroneous white space?

mmgoodnow, line 89:
```suggestion

```

src/preFilter.ts
```diff
@@ -83,7 +86,6 @@ export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
 			last_searched_all: db.raw("coalesce(timestamp.last_searched, 0)"),
 		})
 		.first();
```
mmgoodnow, line 88:
```suggestion
		.first();

```

src/preFilter.ts
mindset-tk: yes, erroneous.

mmgoodnow:
are you really sure that `max` should be `min`?

src/preFilter.ts
```diff
@@ -70,11 +70,14 @@ export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
 			"timestamp.indexer_id": "indexer.id",
 			"timestamp.searchee_id": "searchee.id",
 		})
+		.where("searchee.name",
+		searchee.name
+		)
```
mmgoodnow, line 75: wow this one is embarrassing

src/preFilter.ts
```diff
@@ -70,11 +70,14 @@ export async function filterTimestamps(searchee: Searchee): Promise<boolean> {
...
 		.whereIn(
 			"indexer.id",
 			enabledIndexers.map((i) => i.id)
 		)
-		.max({
+		.min({
```
mmgoodnow, line 80: I don't think this change is right. it would probably make more sense called `first_searched_any` though…

mmgoodnow: nvm I think I'm just getting confused

mmgoodnow: ok yeah I think I understand it now. we really want first_searched_any and last_searched_all, and they are both calculated with `min`

mmgoodnow:
wow this is kind of a huge bugfix. props to you for figuring it out, I spent so much time trying to think about the sql math

mmgoodnow:
would you mind making that rename change and then I'll merge?

src/preFilter.ts
mindset-tk: Ah, yes, that's a good name change. Let me just push that over.

mindset-tk:
> wow this is kind of a huge bugfix. props to you for figuring it out, I spent so much time trying to think about the sql math

No worries! I work with query langs and still get confused all the time. And Knex syntax added an extra layer of complexity to my brain. Happy to help. I updated that var name.

mmgoodnow:
> Hey, one other thing, I found that leaving line 85 on prefilter.ts as "min" will allow the tool to revisit indexers that were previously filtered by timestamps, see https://github.com/cross-seed/cross-seed/issues/402 - the code for performing the search is already sensitive to if an indexer was recently searched and will filter at that point.

That was the original intention but I guess it never worked.

mmgoodnow:
Thanks for the contribution!

PR #481, opened 2023-09-09
Respect `Retry-After` header in Torznab queries

### LINKED ISSUE #375: Respect Retry-After for grabs
https://github.com/cross-seed/cross-seed/blob/7486f190c2c893d4415e3fb75f12623bba896bca/src/torznab.ts#L431-L436

rather than assuming an hour; check and use the retry-after header

bakerboy448:
# ToDo

- [x] cleanup initial chatgpt code 😅
- [ ] Test
- [ ] verbose logging?

# Git Ref
- Fixes #375

### DIFF
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -179,11 +179,10 @@ export async function searchTorznab(
 	const timestampCallout = " (filtered by timestamps)";
 	logger.info({
 		label: Label.TORZNAB,
-		message: `Searching ${indexersToUse.length} indexers for ${name}${
-			indexersToUse.length < enabledIndexers.length
-				? timestampCallout
-				: ""
-		}`,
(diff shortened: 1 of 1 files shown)


onedr0p:
I think the workflow needs to be updated to `pull_request_target` instead of `pull_request` so the tokens carry over.

https://github.com/cross-seed/cross-seed/blob/master/.github/workflows/ci.yml#L9

mmgoodnow:
LGTM, lmk when you're ready for a merge.

bakerboy448:
changes look good.  Probably should be tested, but also seems simple/logical enough to just Yeeet in 😅

onedr0p:
Logic seems sound, unfortunate that we cannot test a container build of this PR. We should be able to make PRs build though without much a hassle.

mmgoodnow:
yolo

PR #482, opened 2023-09-09
Pre filter existing torrents for fuzzy matching using existing regex

akesher123:
Fixes long standing bug-ish behavior from discord. Fuzzy matching would get equal quality matches between multiple episodes from the same season, and would always pick the first alphabetically. This attempts to filter the potential matches with the same regex used for parsing torznab searches.

If the parsing fails (no results), search with all torrents fuzzy anyway.


src/torrent.ts
```diff
@@ -201,8 +201,32 @@ export async function getTorrentByFuzzyName(
...
+			.replaceAll(".", "").replaceAll(" ", "").toLowerCase();
+			if (!dbMatch) return false;
+			return fullMatch === dbMatch});
+	}
+
+	// If none match, proceed with fuzzy name check on all names.
+
 	// @ts-expect-error fuse types are confused
-	const potentialMatches = new Fuse(allNames, {
+	const potentialMatches = new Fuse(filteredNames.length > 0 ? filteredNames : allNames,
```
mmgoodnow, line 228: I read about [uFuzzy](https://github.com/leeoniya) a while back, which has been on my list to replace Fuse with (every fuse search uses a ridiculous amount of CPU and locks the main thread for a while. if you're interested, this could be a pretty nice "while you're here" free performance improvement

mmgoodnow (dismissed):

mmgoodnow:
LGTM pending the helper function extraction

mmgoodnow:
also it looks like you got a typescript error: 

```
13.95 src/torrent.ts(209,4): error TS2550: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.
```

src/torrent.ts
```diff
@@ -201,8 +201,32 @@ export async function getTorrentByFuzzyName(
...
 	const allNames = await db("torrent").select("name", "file_path");
+
+	const episode = name.match(EP_REGEX);
+	const season = name.match(SEASON_REGEX);
+	const movie = name.match(MOVIE_REGEX);
+	const fullMatch = (episode?.[0] ?? season?.[0] ?? movie?.[0])
+		.replaceAll(".", "").replaceAll(" ", "").toLowerCase();
+
+	// Attempt to filter torrents in DB to match incoming torrent before fuzzy check
+	var filteredNames = [];
```
mmgoodnow, line 212: nit: prefer `const` or `let` if mutable
```suggestion
	const filteredNames = [];
```

src/torrent.ts
```diff
@@ -201,8 +201,32 @@ export async function getTorrentByFuzzyName(
...
+
+	if (fullMatch) {
+		filteredNames = allNames.filter((dbName) => {
+			const dbEpisode = dbName.match(EP_REGEX);
+			const dbSeason = dbName.match(SEASON_REGEX);
+			const dbMovie = dbName.match(MOVIE_REGEX);
+			const dbMatch = (dbEpisode?.[0] ?? dbSeason?.[0] ?? dbMovie?.[0])
+			.replaceAll(".", "").replaceAll(" ", "").toLowerCase();
+			if (!dbMatch) return false;
+			return fullMatch === dbMatch});
```
mmgoodnow, line 222: this file needs to be run through prettier

src/torrent.ts
akesher123: I'll take a look

akesher123:
> also it looks like you got a typescript error:

tbh I don't know what's causing this nor how to fix it. `replaceAll` literally works a few lines later so this one is lost on me.

mmgoodnow:
> ```Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.```

I think we just need to tell typescript we're using a later version of javascript than it thinks:

{compilerOptions: {..., lib: "es2021"}}

src/torrent.ts
```diff
@@ -201,8 +201,32 @@ export async function getTorrentByFuzzyName(
 	name: string
 ): Promise<null | Metafile> {
 	const allNames = await db("torrent").select("name", "file_path");
+
+	const episode = name.match(EP_REGEX);
+	const season = name.match(SEASON_REGEX);
+	const movie = name.match(MOVIE_REGEX);
+	const fullMatch = (episode?.[0] ?? season?.[0] ?? movie?.[0])
+		.replaceAll(".", "").replaceAll(" ", "").toLowerCase();
+
```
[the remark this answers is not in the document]
akesher123: oop turns it out's already in one `reformatTitleForSearching(string): string`

src/action.ts
```diff
@@ -44,7 +44,9 @@ export async function performAction(
 			// Candidate is single, nested file
 			if (candidateParentDir != ".") {
 				if (!existsSync(path.join(linkDir, candidateParentDir))) {
-					mkdirSync(path.join(linkDir, candidateParentDir), { recursive: true });
+					mkdirSync(path.join(linkDir, candidateParentDir), {
```
akesher123, line 47: prettier spit this one out, unrelated but might as well merge

package.json
```diff
@@ -48,10 +48,10 @@
 		"@types/xmlrpc": "^1.3.6",
 		"@typescript-eslint/eslint-plugin": "^5.8.1",
 		"@typescript-eslint/parser": "^5.8.1",
-		"eslint": "^8.5.0",
+		"eslint": "^8.49.0",
```
akesher123, line 51: can revert these, also a result of prettier. please advice

mmgoodnow: weird, yeah can we revert package and package-lock?

src/torrent.ts
```diff
@@ -201,12 +206,37 @@ export async function getTorrentByFuzzyName(
 	name: string
 ): Promise<null | Metafile> {
 	const allNames = await db("torrent").select("name", "file_path");
+
+	const fullMatch = reformatTitleForSearching(name).replace(
+		/[^a-z0-9]/gi,
```
mmgoodnow, line 211: is this going to be ineffective on other languages?

package-lock.json
```diff
@@ -5424,4 +5424,4 @@
 			"dev": true
 		}
 	}
-}
+}
```
mmgoodnow, line 5427:
```suggestion
}

```

package.json
```diff
@@ -67,4 +67,4 @@
 	"np": {
 		"yarn": false
 	}
-}
+}
```
mmgoodnow, line 70:
```suggestion
}

```

src/torrent.ts
```diff
@@ -201,15 +206,40 @@ export async function getTorrentByFuzzyName(
...
+
+	// Attempt to filter torrents in DB to match incoming torrent before fuzzy check
+	let filteredNames = [];
+
+	if (fullMatch) {
+		filteredNames = allNames.filter((dbName) => {
+			const dbMatch = reformatTitleForSearching(dbName.item.name).replace(
+				/[^a-z0-9]/gi,
+				""
+			);
```
mmgoodnow, line 223: maybe for the future: I wonder if we would only want to check season and episode number strictly and then leave the name matching to the fuzzy search?

src/torrent.ts
```diff
@@ -201,15 +206,40 @@ export async function getTorrentByFuzzyName(
...
+		/[^a-z0-9]/gi,
+		""
+	);
+
+	// Attempt to filter torrents in DB to match incoming torrent before fuzzy check
+	let filteredNames = [];
+
+	if (fullMatch) {
+		filteredNames = allNames.filter((dbName) => {
+			const dbMatch = reformatTitleForSearching(dbName.item.name).replace(
```
mmgoodnow, line 220: where is this `.item` coming from? `allNames` looks like it has a shape of `{ name: string, file_path: string }[]`

src/torrent.ts
```diff
@@ -201,15 +206,40 @@ export async function getTorrentByFuzzyName(
...
+				""
+			);
+			if (!dbMatch) return false;
+			return fullMatch === dbMatch;
+		});
+	}
+
+	// If none match, proceed with fuzzy name check on all names.
+
+	const uf = new uFuzzy();
```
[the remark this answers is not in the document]
akesher123: like as a static variable? idk how much the instantiation costs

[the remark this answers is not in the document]
mmgoodnow: yeah, like at the top of the file or something. It's probably not very expensive though

src/torrent.ts
akesher123: copied from below, `return parseTorrentFromFilename(firstMatch.item.file_path);`

not sure where it comes from. It's part of the syntax I'm not familiar with, but can change it

mmgoodnow: I think that's part of the Fuse api

src/torrent.ts
akesher123: I think this is reasonable, especially if ufuzzy is more consistent than fuse. We can see how this goes and if it misses a lot of matches maybe dial back the prefiltering

src/torrent.ts
leeoniya: 👋 ping me if you have questions 😉 

https://github.com/leeoniya/uFuzzy/releases/tag/1.0.10

src/torrent.ts
leeoniya: you can drag/drop a json or text list into the demo to compare match quality between Fuse and uFuzzy: https://leeoniya.github.io/uFuzzy/demos/compare.html?libs=uFuzzy,fuzzysort,QuickScore,Fuse&search=super%20ma

src/torrent.ts
```diff
@@ -201,15 +206,37 @@ export async function getTorrentByFuzzyName(
 	name: string
 ): Promise<null | Metafile> {
 	const allNames = await db("torrent").select("name", "file_path");
-	// @ts-expect-error fuse types are confused
```
mmgoodnow, line 204: I think we still need this @ts-expect-error above the Fuse call

mmgoodnow (approved):
LGTM - have you tested it out? might need to hardcode some stuff in order to do so

akesher123:
> LGTM - have you tested it out? might need to hardcode some stuff in order to do so

Yes. While I don't know whether it's more successful or not, I confirmed that it can find matches via RSS

mmgoodnow (approved):

PR #487, opened 2023-09-12
fix(regex): movie and tv (pack and episode) regex

zakkarry:
The following are changes made to improve matching for TV (season pack and episode) as well as movies.

# All Regex's
* Replaced literal space (' ') in expressions with \s
* Changed <title> capture group to match lazily, reducing the total steps per match proportionally to the total string length after the second capture group.

# MOVIE_REGEX
* Added support for '_' in release name.

# SEASON_REGEX
* Added support for '_' in release naming. (Does not use in season numbering range for multi-season packs)
* Capture multiple variations (including poorly named) of multiple season packs properly.
* Fixed matching incorrectly when season has more than one digit

# EP_REGEX
* Added more variations on multi-episode capturing

If you run some test cases and find any outliers or mismatching, please comment and note what release naming you used to demonstrate.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -6,11 +6,11 @@ export const PROGRAM_NAME = packageDotJson.name;
 export const PROGRAM_VERSION = packageDotJson.version;
 export const USER_AGENT = `CrossSeed/${PROGRAM_VERSION}`;
 
-export const EP_REGEX = /^(?<title>.+)[. ](?<season>S\d+)(?<episode>E\d+)/i;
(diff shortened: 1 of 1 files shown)


mmgoodnow:
Congrats on first PR! Can you explain what you mean by capturing lazily in `<title>`?

zakkarry:
I removed that actually before PR'ing but wrote the initial PR prior to posting. I wasn't aware of why it was perhaps greedily capturing. So i decided against changing it.

Simply changing the (?<title>.+) to /^(?<title>.+?) will reduce the steps, and in my test cases I could find no negative results but since it wasn't initially included I just left it out of the commit code, but forgot to edit the comment.

Example in ep_regex matching "Hello_World_S01E01_ntb.mkv":

Using (?<title>.+) takes 58 steps.
Using (?<title>.+?) takes 43 steps.

Match results are the same.

Issue #488, opened 2023-09-14
Add option `excludeSeasonPackEpisodes `


onedr0p:
I would love to disable episodes searches that are part of a season pack. Currently there isn't a way to do this as `includeEpisodes` is only for singular episodes.

zakkarry:
Just merged option `includeSingleEpisodes` which addresses this functionality. Closing this issue.

PR #493, opened 2023-09-23
feat(option): add `includeSingleEpisodes`

zakkarry:
addresses the feature request for individual episodes in season packs being included/excluded in data based searches (https://github.com/cross-seed/cross-seed/issues/488)

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -138,10 +138,16 @@ function createCommandWithSharedOptions(name, description) {
 			fallback(fileConfig.includeEpisodes, false)
 		)
 		.option(
+			"--include-season-pack-episodes",
+			"Include episodes from a season pack in the search",
+			fallback(fileConfig.includeSeasonPackEpisodes, false)
+		)
+		.option(
 			"--no-include-non-videos",
 			"Don't include torrents which contain non-videos"
 		)
 		.option("--no-include-episodes", "Don't include episode torrents")
+		.option("--no-include-season-pack-episodes", "Don't include episodes from a season pack")
 		.requiredOption(
 			"--fuzzy-size-threshold <decimal>",
 			"The size difference allowed to be considered a match.",
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -134,20 +134,20 @@ function createCommandWithSharedOptions(name, description) {
 		)
 		.option(
 			"-e, --include-episodes",
-			"Include single-episode torrents in the search",
+			"Include all episode torrents in the search",
 			fallback(fileConfig.includeEpisodes, false)
 		)
 		.option(
-			"--include-season-pack-episodes",
-			"Include episodes from a season pack in the search",
-			fallback(fileConfig.includeSeasonPackEpisodes, fileConfig.includeEpisodes)
+			"--include-single-episodes",
+			"Include only single episodes torrents in the search",
+			fallback(fileConfig.includeSingleEpisodes, false)
 		)
 		.option(
 			"--no-include-non-videos",
 			"Don't include torrents which contain non-videos"
 		)
 		.option("--no-include-episodes", "Don't include episode torrents")
-		.option("--no-include-season-pack-episodes", "Don't include episodes from a season pack")
+		.option("--no-include-single-episodes", "Don't include episodes from a season pack")
 		.requiredOption(
 			"--fuzzy-size-threshold <decimal>",
 			"The size difference allowed to be considered a match.",
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -309,6 +309,7 @@ createCommandWithSharedOptions("daemon", "Start the cross-seed daemon")
 			setRuntimeConfig(runtimeConfig);
 			initializeLogger();
 			initializePushNotifier();
+			logger.info(`${PROGRAM_NAME} v${PROGRAM_VERSION}`);
 			logger.verbose({
 				label: Label.CONFIGDUMP,
 				message: inspect(runtimeConfig),
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -140,7 +140,7 @@ function createCommandWithSharedOptions(name, description) {
 		.option(
 			"--include-season-pack-episodes",
 			"Include episodes from a season pack in the search",
-			fallback(fileConfig.includeSeasonPackEpisodes, false)
+			fallback(fileConfig.includeSeasonPackEpisodes, fileConfig.includeEpisodes)
 		)
 		.option(
 			"--no-include-non-videos",
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -147,7 +147,6 @@ function createCommandWithSharedOptions(name, description) {
 			"Don't include torrents which contain non-videos"
 		)
 		.option("--no-include-episodes", "Don't include episode torrents")
-		.option("--no-include-single-episodes", "Don't include episodes from a season pack")
 		.requiredOption(
 			"--fuzzy-size-threshold <decimal>",
 			"The size difference allowed to be considered a match.",
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -134,12 +134,12 @@ function createCommandWithSharedOptions(name, description) {
 		)
 		.option(
 			"-e, --include-episodes",
-			"Include all episode torrents in the search",
+			"Include all episode torrents in the search (including from season packs)",
 			fallback(fileConfig.includeEpisodes, false)
 		)
 		.option(
 			"--include-single-episodes",
-			"Include only single episodes torrents in the search",
+			"Include single episode torrents in the search",
 			fallback(fileConfig.includeSingleEpisodes, false)
 		)
 		.option(
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -156,6 +156,8 @@ module.exports = {
 	 * cross-seed will send POST requests to this url
 	 * with a JSON payload of { title, body }.
 	 * Conforms to the caronc/apprise REST API.
+	 * If necessary, supply your username and password inside the url like so:
+	 * "http://username:password@localhost:8000/notify/cross-seed"
 	 */
 	notificationWebhookUrl: undefined,
 
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -76,15 +76,25 @@ module.exports = {
 	outputDir: ".",
 
 	/**
+	 * 
+	 * To search for everything except individual episodes, use (includeEpisodes: false, includeSeasonPackEpisodes: false, includeNonVideos: true)
+	 * To search for all episodes, use (includeEpisodes: true, includeSeasonPackEpisodes: true, includeNonVideos: true)
+	 * To search for season pack episodes, but exclude single episode torrents, use (includeEpisodes: false, includeSeasonPackEpisodes: true, includeNonVideos: true)
+	 *
 	 * Whether to search for single episode torrents
 	 */
 	includeEpisodes: false,
 
 	/**
+	 * Whether to search for single episodes from a season pack during
+	 * data based searches.
+	 */
+	includeSeasonPackEpisodes: false,
+
+	/**
 	 * Include torrents which contain non-video files
-	 * This option does not override includeEpisodes.
-	 * To search for everything except episodes, use (includeEpisodes: false, includeNonVideos: true)
-	 * To search for everything including episodes, use (includeEpisodes: true, includeNonVideos: true)
+	 * This option does not override includeEpisodes or
+	 * includeSeasonPackEpisodes.
 	 */
 	includeNonVideos: false,
 
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -81,7 +81,7 @@ module.exports = {
 	includeEpisodes: false,
 
 	/**
-	 * Whether to include single episode torrents in the search (not from season packs). 
+	 * Whether to include single episode torrents in the search (not from season packs).
 	 * Like `includeEpisodes` but slightly more restrictive.
 	 */
 	includeSingleEpisodes: false,
@@ -89,7 +89,7 @@ module.exports = {
 	/**
 	 * Include torrents which contain non-video files
 	 * This option does not override includeEpisodes or includeSingleEpisodes.
-	 * 
+	 *
 	 * To search for everything except episodes, use (includeEpisodes: false, includeSingleEpisodes: false, includeNonVideos: true)
 	 * To search for everything including episodes, use (includeEpisodes: true, includeNonVideos: true)
 	 * To search for everything except season pack episodes (data-based)
@@ -165,8 +165,6 @@ module.exports = {
 	 * cross-seed will send POST requests to this url
 	 * with a JSON payload of { title, body }.
 	 * Conforms to the caronc/apprise REST API.
-	 * If necessary, supply your username and password inside the url like so:
-	 * "http://username:password@localhost:8000/notify/cross-seed"
 	 */
 	notificationWebhookUrl: undefined,
 
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -77,19 +77,19 @@ module.exports = {
 
 	/**
 	 * 
-	 * To search for everything except individual episodes, use (includeEpisodes: false, includeSeasonPackEpisodes: false, includeNonVideos: true)
-	 * To search for all episodes, use (includeEpisodes: true, includeSeasonPackEpisodes: true, includeNonVideos: true)
-	 * To search for season pack episodes, but exclude single episode torrents, use (includeEpisodes: false, includeSeasonPackEpisodes: true, includeNonVideos: true)
+	 * To search for all episodes, use (includeEpisodes: true, includeNonVideos: true)
+	 * To search for only single episodes, but exclude season pack episodes (data-based)
+	 *    use (includeEpisodes: false, includeSingleEpisodes: true, includeNonVideos: true)
 	 *
-	 * Whether to search for single episode torrents
+	 * Whether to search for all episode torrents (Overrides includeSingleEpisodes)
 	 */
 	includeEpisodes: false,
 
 	/**
-	 * Whether to search for single episodes from a season pack during
-	 * data based searches.
+	 * Whether to search only for single episodes (not from season packs)
+	 * during data based searches. 
 	 */
-	includeSeasonPackEpisodes: false,
+	includeSingleEpisodes: false,
 
 	/**
 	 * Include torrents which contain non-video files
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -86,8 +86,8 @@ module.exports = {
 	includeEpisodes: false,
 
 	/**
-	 * Search for single episodes (not from season packs)
-	 * during data based searches. 
+	 * Whether to include single episode torrents in the search (not from season packs). 
+	 * Like `includeEpisodes` but slightly more restrictive.
 	 */
 	includeSingleEpisodes: false,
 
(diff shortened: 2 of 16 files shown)


src/cmd.ts
```diff
@@ -137,11 +137,17 @@ function createCommandWithSharedOptions(name, description) {
 			"Include single-episode torrents in the search",
 			fallback(fileConfig.includeEpisodes, false)
 		)
+		.option(
+			"--include-season-pack-episodes",
+			"Include episodes from a season pack in the search",
+			fallback(fileConfig.includeSeasonPackEpisodes, false)
```
mmgoodnow, line 143: this will change the behavior for everyone when we release the new version, so we have a couple options:

- bump the major version and tell everyone about breaking changes (more work)
- default to `true` (tech debt)
- wait until another breaking change comes along and bundle it with that
- offer some other sort of escape hatch so that we don't need another argument here

mmgoodnow (commented):
logic looks good, just the question about the breaking change

mmgoodnow:
thinking on a higher level, I think pre filtering would be a good thing to stop doing altogether on first run. A lot of the pre filtering defaults are just based on my setup at the time I added them, so eventually it would probably make the new user experience a lot smoother to just remove all these gotchas. If we do it all at once and batch the breaking changes together it will minimize disruption

mmgoodnow:
For now, probably the thing to do is to default it to true

src/cmd.ts
zakkarry: I did think about the potential to rely on includeEpisodes if undefined, for instance if they are using a deprecated version of the template/config or something.

I'm not sure how it would best be handled, but if we have includeEpisodes = true and an undefined (neither true or false) includeSeasonPackEpisodes,default to previous behavior (includeSeasonPackEpisodes = true)

We obviously don't want to assume anything about preferences, but changing behavior is sort of a concern i thought about, especially with people rarely touching their configurations after initial setup.

zakkarry:
Could we not keep the same current behavior with the following?

`fallback(fileConfig.includeSeasonPackEpisodes, fileConfig.includeEpisodes)`

mmgoodnow:
i think i want to keep options independent of each other, let's stick with `fallback(fileConfig.includeSeasonPackEpisodes, true)`

src/preFilter.ts
```diff
@@ -18,14 +19,18 @@ export function filterByContent(searchee: Searchee): boolean {
 		});
 	}
 
-	const isSingleEpisodeTorrent =
-		searchee.files.length === 1 && EP_REGEX.test(searchee.files[0].name);
-
-	if (!includeEpisodes && isSingleEpisodeTorrent) {
+	const isSingleEpisodeTorrent = searchee.files.length === 1 && EP_REGEX.test(searchee.name); 
+	const isSeasonPack = typeof searchee.path !== 'undefined' ? SEASON_REGEX.test(path.dirname(searchee.path).split(path.sep).reverse()[0]): false;
```
mmgoodnow, line 23: this dirname -> split -> reverse -> [0] thing could be changed to dirname -> basename

```ts
basename(dirname('/a/b/c/d')) === 'c'
```

src/preFilter.ts
```diff
@@ -18,14 +19,18 @@ export function filterByContent(searchee: Searchee): boolean {
 		});
 	}
 
-	const isSingleEpisodeTorrent =
-		searchee.files.length === 1 && EP_REGEX.test(searchee.files[0].name);
-
-	if (!includeEpisodes && isSingleEpisodeTorrent) {
+	const isSingleEpisodeTorrent = searchee.files.length === 1 && EP_REGEX.test(searchee.name);
```
mmgoodnow, line 22: does this change carry a semantic difference? curious what you found in your testing here

src/preFilter.ts
mmgoodnow, line 23: checking for undefined via `typeof` is the most correct but a bit formal. For the purposes of this check we can probably just check that it's truthy

Additionally, we can use `basename` to our advantage here:

```ts
const comesFromSeasonPack = searchee.path &&
	SEASON_REGEX.test(basename(dirname(searchee.path))
```

src/preFilter.ts
```diff
@@ -18,14 +19,18 @@ export function filterByContent(searchee: Searchee): boolean {
...
-	if (!includeEpisodes && isSingleEpisodeTorrent) {
+	const isSingleEpisodeTorrent = searchee.files.length === 1 && EP_REGEX.test(searchee.name); 
+	const isSeasonPack = typeof searchee.path !== 'undefined' ? SEASON_REGEX.test(path.dirname(searchee.path).split(path.sep).reverse()[0]): false;
+	
+	if (!includeEpisodes && isSingleEpisodeTorrent && !isSeasonPack) {
 		logReason("it is a single episode");
 		return false;
 	}
 
+	if (!includeSeasonPackEpisodes && (searchee.files.length === 1 && isSeasonPack)) {
```
mmgoodnow, line 30: out of curiosity, does isSingleEpisodeTorrent not work here?
```suggestion
	if (!includeSeasonPackEpisodes && isSingleEpisodeTorrent && comesFromSeasonPack) {
```

src/preFilter.ts
```diff
@@ -18,14 +19,18 @@ export function filterByContent(searchee: Searchee): boolean {
...
 	}
 
-	const isSingleEpisodeTorrent =
-		searchee.files.length === 1 && EP_REGEX.test(searchee.files[0].name);
-
-	if (!includeEpisodes && isSingleEpisodeTorrent) {
+	const isSingleEpisodeTorrent = searchee.files.length === 1 && EP_REGEX.test(searchee.name); 
+	const isSeasonPack = typeof searchee.path !== 'undefined' ? SEASON_REGEX.test(path.dirname(searchee.path).split(path.sep).reverse()[0]): false;
+	
+	if (!includeEpisodes && isSingleEpisodeTorrent && !isSeasonPack) {
```
mmgoodnow, line 25: oh, I see what you're saying now about `fallback(fileConfig.includeEpisodes)` now.

I think the best way to make it backwards compatible will probably be to invert the option - instead of a switch to include season pack episodes (when really we want it to be exclude), maybe we can have an option to include standalone episodes (not from a season pack). 

that way, we can preserve the original behavior of `includeEpisodes` but still satisfy the use case. 

I'm also not totally opposed to just making the change, but curious to hear your thoughts

src/preFilter.ts
zakkarry: I think my logic at the time was based on certain season pack types with very poor or absolute naming schemes, specifically anime (absolute numbering which would not match EP_REGEX) and daily or soap opera shows.

It occurred to me that this was incredibly edge-case outside of anime but with a comparison as simple a number...I just opted for this instead.

src/preFilter.ts
zakkarry: After discussion, @akesher123 mentioned utilizing name would yield more consistent/usable naming to test.

> also not sure on this but I think it might be better to use EP_REGEX.test(searchee.name) instead, only because I've seen some pretty terribly named files but torrent name is usually good

Simply took the advice...

I'll check this out further and post back.

mmgoodnow: that sounds fine to me!

src/preFilter.ts
mmgoodnow: good point about the anime, I like that. Let's move the files.length === 1 check into the isSeasonPack variable above though

mmgoodnow: and then probably the best name for the variable after that change is "isSeasonPackEpisode"

src/preFilter.ts
zakkarry: Testing that I was able to do resulted in the following:

- All single file torrents yield identical results between files[0].name and .name. I don't have, or know how to consistently find, torrents that would yield differences in these values. I'd imagine they could exist, though. Probably on public trackers.

- Any torrents with a single season folder hierarchy resulted in searchee.name being the folder name, as opposed to file[0].name being the first file, such as S01E01.

- The only outlier I thought to test was something with multiple folders in the root torrent (multi-season packs) - but this also yielded the main torrent name (ex: Show S01-S05) with searchee.name (as opposed to any individual folders within) while files[0].name obviously returning the first file therewithin - such as S01E01.

Ultimately, my conclusion is that the results could affect whether EP_REGEX matches in extreme edge cases only (where the torrent name is actually MORE messed up than the file's naming). For actual single episodes, I was unable to find applicable outliers due to most of the sources I have being extremely consistent in the type of scenario. 

The one possible reason to utilize searchee.name would be for @akesher123's quoted reasoning, despite me being unable to actually find applicable torrents to run against. His logic does seem to track for those edge cases, wherever they are :P

src/preFilter.ts
zakkarry: Correct me if I'm wrong, but after some thought would it not make sense to save some CPU cycles and eliminate searchee.path's truthy check altogether?

Assuming the SEASON_REGEX.test returns false when it's provided an undefined value (I'd expect this behavior) - this actually seems redundant being in the same initialization.

`const isSeasonPackEpisode = SEASON_REGEX.test(path.basename(path.dirname(searchee.path)));`

akesher123: > `SEASON_REGEX.test(basename(dirname(searchee.path))`

This is not correct. For a season pack episode, `searchee.path` will be /torrents/tv/Show.S01/, dirname will give `torrents/tv`, basename gives `Show.S01`, basename(dirname) gives `tv`. So just basename should be the way to go.

> would it not make sense to save some CPU cycles and eliminate searchee.path's truthy check altogether?

If it's undefined, it should do the same amount of processing on both due to short circuiting. If the left side of the && is false, it not try to evaluate the right side. `.test()` does appear to handle undefined though, so maybe the extra null check that is done if it's true is not worth it, but I prefer the hard null check before running `test`

src/preFilter.ts
mmgoodnow: > All single file torrents yield identical results between files[0].name and .name. I don't have, or know how to consistently find, torrents that would yield differences in these values. I'd imagine they could exist, though. Probably on public trackers.

these exist. Most of the time, the difference between `.name` and `.files[0].name` is that the latter ends in `.mkv` and the former doesn't. Under the hood, these are distinct BitTorrent formats (single file torrents, which download as a file, and multi-file torrents, which download as a folder)

```js
{ 
  name: "Hello", 
  files: [
    { 
      name: "Hello.mkv", 
      path: "Hello/Hello.mkv" 
    }
  ]
}
```

```js
{ 
  name: "Hello.mkv", 
  files: [
    { 
      name: "Hello.mkv", 
      path: "Hello.mkv"
    }
  ]
}
```

However, I think for our purposes here, ^ is not an issue.


The rest all seems pretty sound. I like the change. 👍

src/preFilter.ts
mmgoodnow: I want to be explicit that searchee.path has to exist. Otherwise we're relying on behavior of dirname and basename when it encounters undefined, which feels bad.

zakkarry: > If it's undefined, it should do the same amount of processing on both due to short circuiting. If the left side of the && is false, it not try to evaluate the right side. .test() does appear to handle undefined though, so maybe the extra null check that is done if it's true is not worth it, but I prefer the hard null check before running test

Noted, will keep the ternary as is with the SEASON_REGEX i commented previous your comment.

`const isSeasonPackEpisode = searchee.path && SEASON_REGEX.test(path.basename(path.dirname(searchee.path)));`

Functions as intended.

mmgoodnow: nit, but I'd prefer `&&` to using a ternary here

zakkarry: > nit, but I'd prefer `&&` to using a ternary here

Sure, makes it more readable anyway. Will use the &&.

src/preFilter.ts
zakkarry: Sounds perfect.

src/preFilter.ts
zakkarry: Understood. Appreciate the insight :)

src/preFilter.ts
zakkarry: I think the "include" theme for the content filtering is something to potentially maintain, as we don't have any other excluding (content-based) config options. I'm not really opposed to changing to exclusion, but for mere consistency, and maybe aesthetic of the content config portion, it would seem logical.

Current behavior with `includeEpisodes = true` would essentially be the same with falling-back to fileConfig.includeEpisodes. 

I see this as the optimal way to ensure consistency of behavior for those who don't actively watch changelog/commits in version updates, and would otherwise have the introduction of includeSeasonPackEpisodes go completely unnoticed. 

I also assume most people would not be fussing with their configuration file regularly, even fewer are probably looking out for potentially adding new options. 

Set it and forget it situation. Update the container when notified, but never read changelog.

`fallback(fileConfig.includeSeasonPackEpisodes, fileConfig.includeEpisodes)`

src/cmd.ts
```diff
@@ -134,14 +134,20 @@ function createCommandWithSharedOptions(name, description) {
 		)
 		.option(
 			"-e, --include-episodes",
-			"Include single-episode torrents in the search",
+			"Include all episode torrents in the search",
 			fallback(fileConfig.includeEpisodes, false)
 		)
+		.option(
+			"--include-single-episodes",
+			"Include only single episodes torrents in the search",
```
mmgoodnow, line 142:
```suggestion
			"Include single episode torrents in the search",
```

src/cmd.ts
```diff
@@ -134,14 +134,20 @@ function createCommandWithSharedOptions(name, description) {
 		)
 		.option(
 			"-e, --include-episodes",
-			"Include single-episode torrents in the search",
+			"Include all episode torrents in the search",
```
mmgoodnow, line 137:
```suggestion
			"Include all episode torrents in the search (including from season packs)",
```

src/cmd.ts
```diff
@@ -134,14 +134,20 @@ function createCommandWithSharedOptions(name, description) {
...
+			"--include-single-episodes",
+			"Include only single episodes torrents in the search",
+			fallback(fileConfig.includeSingleEpisodes, false)
+		)
 		.option(
 			"--no-include-non-videos",
 			"Don't include torrents which contain non-videos"
 		)
 		.option("--no-include-episodes", "Don't include episode torrents")
+		.option("--no-include-single-episodes", "Don't include episodes from a season pack")
```
mmgoodnow, line 150:
```suggestion
		.option("--no-include-single-episodes", "Don't include single episodes")
```

src/config.template.cjs
```diff
@@ -76,15 +76,25 @@ module.exports = {
...
+	 */
+	includeSingleEpisodes: false,
+
 	/**
 	 * Include torrents which contain non-video files
-	 * This option does not override includeEpisodes.
-	 * To search for everything except episodes, use (includeEpisodes: false, includeNonVideos: true)
-	 * To search for everything including episodes, use (includeEpisodes: true, includeNonVideos: true)
+	 * This option does not override includeEpisodes or
+	 * includeSeasonPackEpisodes.
```
mmgoodnow, line 97:
```suggestion
	 * includeSingleEpisodes.
```

src/config.template.cjs
```diff
@@ -76,15 +76,25 @@ module.exports = {
...
+	 * Whether to search only for single episodes (not from season packs)
+	 * during data based searches. 
+	 */
+	includeSingleEpisodes: false,
+
 	/**
 	 * Include torrents which contain non-video files
-	 * This option does not override includeEpisodes.
-	 * To search for everything except episodes, use (includeEpisodes: false, includeNonVideos: true)
-	 * To search for everything including episodes, use (includeEpisodes: true, includeNonVideos: true)
```
mmgoodnow, line 87: I think all of this stuff still holds

src/config.template.cjs
```diff
@@ -76,15 +76,25 @@ module.exports = {
...
+	 * Whether to search for all episode torrents (Overrides includeSingleEpisodes)
 	 */
 	includeEpisodes: false,
 
+	/**
+	 * Whether to search only for single episodes (not from season packs)
+	 * during data based searches. 
+	 */
+	includeSingleEpisodes: false,
+
```
mmgoodnow, line 93:
```suggestion
	/**
	 * Whether to include single episode torrents in the search (not from season packs). Like `includeEpisodes` but slightly more restrictive.
	 */
	includeSingleEpisodes: false,
```

src/config.template.cjs
```diff
@@ -76,15 +76,25 @@ module.exports = {
 	outputDir: ".",
 
 	/**
-	 * Whether to search for single episode torrents
+	 * 
+	 * To search for all episodes, use (includeEpisodes: true, includeNonVideos: true)
+	 * To search for only single episodes, but exclude season pack episodes (data-based)
+	 *    use (includeEpisodes: false, includeSingleEpisodes: true, includeNonVideos: true)
+	 *
+	 * Whether to search for all episode torrents (Overrides includeSingleEpisodes)
```
mmgoodnow, line 84:
```suggestion
	 * Whether to search for all episode torrents, including those from season packs. This option overrides includeSingleEpisodes.
```

src/config.template.cjs
```diff
@@ -76,15 +76,25 @@ module.exports = {
...
 	 */
 	includeEpisodes: false,
 
+	/**
+	 * Whether to search only for single episodes (not from season packs)
```
mmgoodnow, line 89:
```suggestion
	 * Search for single episodes (not from season packs)
```

src/preFilter.ts
```diff
@@ -18,14 +19,18 @@ export function filterByContent(searchee: Searchee): boolean {
...
+	const isSingleEpisodeTorrent = searchee.files.length === 1 && EP_REGEX.test(searchee.name); 
+	const isSeasonPackEpisode = searchee.path && searchee.files.length === 1 && SEASON_REGEX.test(path.basename(path.dirname(searchee.path)));
 
-	if (!includeEpisodes && isSingleEpisodeTorrent) {
+	if (!includeEpisodes && !includeSingleEpisodes && isSingleEpisodeTorrent && !isSeasonPackEpisode) {
 		logReason("it is a single episode");
 		return false;
 	}
 
+	if (includeSingleEpisodes && isSeasonPackEpisode) {
```
mmgoodnow, line 30:
```suggestion
	if (!includeEpisodes && isSeasonPackEpisode) {
```

mmgoodnow (commented):
ok now we're cooking. logic looks solid except for that one suggestion I made, then just some copy cleanup

src/cmd.ts
zakkarry: Will commit the removal of this redundant option.

src/cmd.ts
zakkarry: Remove this option due to redundancy with --no-include-episodes.

src/preFilter.ts
zakkarry: I follow. This makes sense.

src/preFilter.ts
zakkarry: i'm marking this resolved due to the change to includeSingleEpisodes

src/config.template.cjs
zakkarry: Marked as outdate, but committed

src/config.template.docker.cjs
```diff
@@ -80,15 +80,25 @@ module.exports = {
 	outputDir: "/cross-seeds",
 
 	/**
-	 * Whether to search for single episode torrents
+	 * 
+	 * To search for all episodes, use (includeEpisodes: true, includeNonVideos: true)
```
mmgoodnow, line 84: I think this blurb still makes more sense in its old form above includeNonVideos.

mmgoodnow (approved):
LGTM pending last couple nits

src/config.template.docker.cjs
zakkarry: moved

PR #499, opened 2023-09-28
feat(client/deluge): injection support for deluge torrent client

### LINKED ISSUE #58: Add Deluge Torrent Client Support
Requesting deluge be added as a supported client 😁

zakkarry:
# Deluge Client Integration

This PR adds support for injection to Deluge v2 torrent clients.

- Closes #58 

### notes:
- requires RpcUrl uses extractCredential function
- uses JSON-RPC endpoint in deluge's webUI
- utilizes node-fetch library for sending calls
- bare function documentation
- logging and validation of config
- returns InjectionResults properly with appropriate handling

### todo:
- [x] add authentication vs connection errors (timeout/refuse/etc)
- [x] improve the label integration (it's crude atm)
- - [x] conform to current label usage behavior
- [x] remove axios (use node-fetch)
- [x] remove extra delugeWebPassword 
- - [x] get deluge's webUrl to be redacted in logging
- [x] more complete logging
- - [x] standardize logging messaging
- [x] transient auth errors handled with limited retry
- More.....

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -1,4 +1,3 @@
-import axios, { AxiosResponse } from "axios";
 import { InjectionResult } from "../constants.js";
 import { CrossSeedError } from "../errors.js";
 import { Label, logger } from "../logger.js";
@@ -6,7 +5,7 @@ import { Metafile } from "../parseTorrent.js";
 import { getRuntimeConfig } from "../runtimeConfig.js";
 import { Searchee } from "../searchee.js";
 import { TorrentClient } from "./TorrentClient.js";
-
+import fetch, { Headers, Response } from "node-fetch";
 export default class Deluge implements TorrentClient {
 	private msgId = 0;
 	private loggedIn = false;
@@ -50,12 +49,10 @@ export default class Deluge implements TorrentClient {
 			);
 		}
 
-		if (
-			response.data.result &&
-			response.headers &&
-			response.headers["set-cookie"]
-		) {
-			this.delugeCookie = response.headers["set-cookie"][0].split(";")[0];
+		if (response.data.result && response.headers.has("Set-Cookie")) {
+			this.delugeCookie = response.headers
+				.get("Set-Cookie")
+				.split(";")[0];
 			this.loggedIn = true;
 			return;
 		}
@@ -86,26 +83,28 @@ export default class Deluge implements TorrentClient {
 			: response.data.error;
 	}
 
-	private async call(body: any): Promise<AxiosResponse | null> {
-		return new Promise((resolve) => {
-			body.id = ++this.msgId;
-			if (this.msgId > 1024) this.msgId = 0;
-			axios
-				.post(
-					this.delugeWebUrl.origin + this.delugeWebUrl.pathname,
-					body,
-					{
-						headers: {
-							"Content-Type": "application/json",
-							Cookie: this.delugeCookie,
-						},
-					}
-				)
-				.then((data) => {
-					resolve(data);
-				})
-				.catch(() => resolve(null));
-		});
+	private async call(body: { method: string; params: any[] }) {
+		this.msgId = (this.msgId + 1) % 1024;
+
+		const headers = new Headers({ "Content-Type": "application/json" });
+		if (this.delugeCookie) headers.set("Cookie", this.delugeCookie);
+
+		const url = this.delugeWebUrl.origin + this.delugeWebUrl.pathname;
+		let response: Response;
+		try {
+			response = await fetch(url, {
+				body: JSON.stringify({ ...body, msgId: this.msgId }),
+				method: "POST",
+				headers: headers,
+			});
+		} catch (networkError) {
+			return null;
+		}
+
+		if (!response.ok) {
+			return null;
+		}
+		return { ...response, data: (await response.json()) as any };
 	}
 
 	/**
@@ -187,7 +186,7 @@ export default class Deluge implements TorrentClient {
 	 */
 	async checkCompleted(searchee: Searchee): Promise<boolean> {
 		try {
-			let params = [
+			const params = [
 				["name", "state", "save_path"],
 				{ hash: searchee.infoHash },
 			];
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -12,18 +12,19 @@ interface DelugeResponse {
 		message?: string;
 		code?: number;
 	};
-	result?: any;
+	result?: string | boolean;
 }
 
 export default class Deluge implements TorrentClient {
 	private msgId = 0;
 	private delugeCookie = "";
-	private delugeWebUrl: URL;
+	private delugeRpcUrl: URL;
 	private delugeLabel: string;
-	private isLabelEnabled: Promise<boolean>;
+	private isLabelEnabled: boolean;
 
 	constructor() {
-		this.delugeWebUrl = new URL(`${getRuntimeConfig().delugeWebUrl}`);
+		const { delugeRpcUrl } = getRuntimeConfig();
+		this.delugeRpcUrl = new URL(delugeRpcUrl);
 		this.delugeLabel = "cross-seed";
 	}
 
@@ -31,28 +32,30 @@ export default class Deluge implements TorrentClient {
 	 * validates the login and host for deluge webui
 	 */
 	async validateConfig(): Promise<void> {
-		const url = new URL(this.delugeWebUrl);
-		if (url.username && !url.password) {
+		if (!this.delugeRpcUrl.password) {
 			throw new CrossSeedError(
-				"you need to define a password in the delugeWebUrl. (eg: http://:<PASSWORD>@localhost:8112)"
+				"you need to define a password in the delugeRpcUrl. (eg: http://:<PASSWORD>@localhost:8112)"
 			);
 		}
 		await this.authenticate();
-		this.isLabelEnabled = this.labelEnabled();
+		this.isLabelEnabled = await this.labelEnabled();
 	}
 
 	/**
 	 * connects and authenticates to the webui
 	 */
 	private async authenticate(): Promise<void> {
-		const response = await this.call({
-			method: "auth.login",
-			params: [this.delugeWebUrl.password],
-		});
-
+		const response = await this.call("auth.login", [
+			this.delugeRpcUrl.password,
+		]);
 		if (response === null) {
 			throw new CrossSeedError(
-				`failed to establish a connection to deluge: ${this.delugeWebUrl.origin}`
+				`failed to establish a connection to deluge: ${this.delugeRpcUrl.origin}`
+			);
+		}
+		if (!response.result) {
+			throw new CrossSeedError(
+				`failed to authenticate with deluge: ${this.delugeRpcUrl.origin}`
 			);
 		}
 	}
@@ -60,17 +63,20 @@ export default class Deluge implements TorrentClient {
 	/**
 	 * ensures authentication and sends JSON-RPC calls to deluge
 	 */
-	private async call(body: { method: string; params: any[] }) {
+	private async call(method: string, params: object, retries = 1) {
 		this.msgId = (this.msgId + 1) % 1024;
+		const url = this.delugeRpcUrl.origin + this.delugeRpcUrl.pathname;
 
 		const headers = new Headers({ "Content-Type": "application/json" });
 		if (this.delugeCookie) headers.set("Cookie", this.delugeCookie);
 
-		const url = this.delugeWebUrl.origin + this.delugeWebUrl.pathname;
 		let response: Response, json: DelugeResponse;
 		try {
 			response = await fetch(url, {
-				body: JSON.stringify({ ...body, id: this.msgId }),
+				body: JSON.stringify({
+					...{ method: method, params: params },
+					id: this.msgId,
+				}),
 				method: "POST",
 				headers: headers,
 			});
@@ -78,34 +84,27 @@ export default class Deluge implements TorrentClient {
 		} catch (networkError) {
 			return null;
 		}
-
-		if (json && json.error && json.error.code && json.error.code === 1) {
+		if (json?.error?.code === 1 && retries > 0) {
 			await this.authenticate();
-			try {
-				response = await fetch(url, {
-					body: JSON.stringify({ ...body, id: this.msgId }),
-					method: "POST",
-					headers: headers,
-				});
-			} catch (networkError) {
-				return null;
-			}
-			json = (await response.json()) as DelugeResponse;
+			return this.call(method, params, 0);
 		}
 		if (!response.ok) {
 			return null;
 		}
+		this.handleResponseHeaders(response.headers);
 
-		if (response.headers && response.headers.has("Set-Cookie")) {
-			this.delugeCookie = response.headers
-				.get("Set-Cookie")
-				.split(";")[0];
-		} else if (body.method == "auth.login") {
-			throw new CrossSeedError(
-				`failed to authenticate the session in deluge: ${this.delugeWebUrl.origin}`
-			);
-		}
 		return json;
+	}
+
+	/**
+	 * parses the set-cookie header and updates stored value
+	 */
+	private handleResponseHeaders(headers: Headers) {
+		if (headers?.has("Set-Cookie")) {
+			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
+			return true;
+		}
+		return false;
 	}
 
 	/**
@@ -113,11 +112,8 @@ export default class Deluge implements TorrentClient {
 	 * returns true if successful.
 	 */
 	private async labelEnabled() {
-		const enabledLabels = await this.call({
-			method: "core.get_enabled_plugins",
-			params: [],
-		});
-		return enabledLabels.result && enabledLabels.result.includes("Label");
+		const enabledLabels = await this.call("core.get_enabled_plugins", []);
+		return enabledLabels?.result?.includes("Label");
 	}
 
 	/**
@@ -126,23 +122,16 @@ export default class Deluge implements TorrentClient {
 	 */
 	private async setLabel(infoHash: string): Promise<void> {
 		if (this.isLabelEnabled) {
-			const setResult = await this.call({
-				method: "label.set_torrent",
-				params: [infoHash, this.delugeLabel],
-			});
-			if (
-				setResult.error &&
-				setResult.error.code &&
-				setResult.error.code == 4
-			) {
-				await this.call({
-					method: "label.add",
-					params: [this.delugeLabel],
-				});
-				await this.call({
-					method: "label.set_torrent",
-					params: [infoHash, this.delugeLabel],
-				});
+			const setResult = await this.call("label.set_torrent", [
+				infoHash,
+				this.delugeLabel,
+			]);
+			if (setResult?.error?.code == 4) {
+				await this.call("label.add", [this.delugeLabel]);
+				await this.call("label.set_torrent", [
+					infoHash,
+					this.delugeLabel,
+				]);
 			}
 		}
 	}
@@ -161,22 +150,15 @@ export default class Deluge implements TorrentClient {
 		const params = this.formatData(
 			`${newTorrent.name}.cross-seed.torrent`,
 			newTorrent.encode().toString("base64"),
-			path
+			path ? path : await this.getSavePath(searchee)
 		);
-		const addResult = await this.call({
-			method: "core.add_torrent_file",
-			params: params,
-		});
+		const addResult = await this.call("core.add_torrent_file", params);
 		if (addResult.result) {
 			this.setLabel(newTorrent.infoHash);
 			return InjectionResult.SUCCESS;
-		} else if (
-			addResult.error &&
-			addResult.error.message &&
-			addResult.error.message.includes("already")
-		) {
+		} else if (addResult?.error?.message?.includes("already")) {
 			return InjectionResult.ALREADY_EXISTS;
-		} else if (addResult.error && addResult.error.message) {
+		} else if (addResult?.error?.message) {
 			logger.debug({
 				label: Label.DELUGE,
 				message: `injection failed: ${addResult.error.message}`,
@@ -201,25 +183,36 @@ export default class Deluge implements TorrentClient {
 	}
 
 	/**
+	 * returns true if the torrent hash is completed in deluge
+	 */
+	private async getSavePath(searchee: Searchee): Promise<string> {
+		try {
+			const params = [["save_path"], { hash: searchee.infoHash }];
+
+			const response = await this.call("web.update_ui", params);
+			return response?.result?.torrents?.[searchee.infoHash]?.save_path;
+		} catch (e) {
+			logger.debug({
+				label: Label.DELUGE,
+				message: `failed to fetch torrent save_path: ${searchee.name} - ${searchee.infoHash}`,
+			});
+		}
+		return null;
+	}
+
+	/**
 	 * returns true if the torrent hash is Seeding (completed)
 	 */
-	async checkCompleted(searchee: Searchee): Promise<boolean> {
+	private async checkCompleted(searchee: Searchee): Promise<boolean> {
 		try {
 			const params = [["state", "progress"], { hash: searchee.infoHash }];
 
-			const response = await this.call({
-				method: "web.update_ui",
-				params: params,
-			});
+			const response = await this.call("web.update_ui", params);
 			return (
-				response.result &&
-				response.result.torrents &&
-				(response.result.torrents[searchee.infoHash].state ===
+				response?.result?.torrents?.[searchee.infoHash]?.state ===
 					"Seeding" ||
-					(response.result.torrents[searchee.infoHash].state ===
-						"Paused" &&
-						response.result.torrents[searchee.infoHash].progress ===
-							100))
+				response?.result?.torrents?.[searchee.infoHash]?.progress ===
+					100
 			);
 		} catch (e) {
 			logger.debug({
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -6,14 +6,21 @@ import { getRuntimeConfig } from "../runtimeConfig.js";
 import { Searchee } from "../searchee.js";
 import { TorrentClient } from "./TorrentClient.js";
 import fetch, { Headers, Response } from "node-fetch";
+
+interface DelugeResponse {
+	error?: {
+		message?: string;
+		code?: number;
+	};
+	result?: any;
+}
+
 export default class Deluge implements TorrentClient {
 	private msgId = 0;
-	private loggedIn = false;
 	private delugeCookie = "";
 	private delugeWebUrl: URL;
 	private delugeLabel: string;
 	private isLabelEnabled: Promise<boolean>;
-	private delugeRetries: number;
 
 	constructor() {
 		this.delugeWebUrl = new URL(`${getRuntimeConfig().delugeWebUrl}`);
@@ -38,9 +45,9 @@ export default class Deluge implements TorrentClient {
 	 * connects and authenticates to the webui
 	 */
 	private async authenticate(): Promise<void> {
-		let response = await this.call({
-			params: [this.delugeWebUrl.password],
+		const response = await this.call({
 			method: "auth.login",
+			params: [this.delugeWebUrl.password],
 		});
 
 		if (response === null) {
@@ -48,41 +55,11 @@ export default class Deluge implements TorrentClient {
 				`failed to establish a connection to deluge: ${this.delugeWebUrl.origin}`
 			);
 		}
-
-		if (response.data.result && response.headers.has("Set-Cookie")) {
-			this.delugeCookie = response.headers
-				.get("Set-Cookie")
-				.split(";")[0];
-			this.loggedIn = true;
-			return;
-		}
-		throw new CrossSeedError(
-			`failed to authenticate the session in deluge: ${this.delugeWebUrl.origin}`
-		);
 	}
 
 	/**
 	 * ensures authentication and sends JSON-RPC calls to deluge
 	 */
-	private async sendCall(method: string, params: any[]) {
-		let response;
-		response = await this.call({
-			params: params,
-			method,
-		});
-
-		if (response.code == 1) {
-			await this.authenticate();
-			response = await this.call({
-				params: params,
-				method,
-			});
-		}
-		return response && response.data && response.data.error === null
-			? response.data
-			: response.data.error;
-	}
-
 	private async call(body: { method: string; params: any[] }) {
 		this.msgId = (this.msgId + 1) % 1024;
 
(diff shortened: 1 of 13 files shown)


src/clients/Deluge.ts
```diff
@@ -0,0 +1,225 @@
...
+	/**
+	 * lists plugins and adds/sets labels
+	 * returns true if successful.
+	 */
+	private async labelEnabled() {
+		const enabledLabels = await this.sendCall(
+			"core.get_enabled_plugins",
+			[]
+		);
+		if (enabledLabels["result"].includes("Label")) {
```
akesher123, line 120: Can just be `return enabledLabels["result"].includes("Label"))`

src/clients/Deluge.ts
```diff
@@ -0,0 +1,225 @@
...
+			return true;
+		}
+		return false;
+	}
+	private async setLabel(torrenthash: string) {
+		if (this.enLabeled) {
```
akesher123, line 126: I don't see anything that will set this to true

src/clients/TorrentClient.ts
```diff
@@ -18,14 +19,16 @@ export interface TorrentClient {
 }
 
 function instantiateDownloadClient() {
-	const { rtorrentRpcUrl, qbittorrentUrl, transmissionRpcUrl } =
+	const { rtorrentRpcUrl, qbittorrentUrl, transmissionRpcUrl, delugeWebUrl, delugeWebPassword } =
```
akesher123, line 22: is it a limitation of deluge that requires the weburl and password to be treated separately unlike the other clients?

src/startup.ts
```diff
@@ -12,15 +12,16 @@ function validateOptions() {
...
 		matchMode,
 		skipRecheck,
 	} = getRuntimeConfig();
 	if (
 		action === "inject" &&
-		!(rtorrentRpcUrl || qbittorrentUrl || transmissionRpcUrl)
-	) {
+		!(rtorrentRpcUrl || qbittorrentUrl || transmissionRpcUrl || (delugeWebUrl && delugeWebPassword))) {
 		throw new CrossSeedError(
 			"You need to specify --rtorrent-rpc-url, --transmission-rpc-url, or --qbittorrent-url when using '-A inject'."
```
akesher123, line 26: remember to add --deluge-web-url here later

zakkarry: oops, oversight.

src/clients/TorrentClient.ts
zakkarry: For deluge's WebUI, the page itself does not require any standard authentication header. Instead, we are confronted with a form for the password which is pushed for authorization with the `auth.login` RPC method.

This auth is done for the /json endpoint as well, but "manually" since its a JSON-RPC endpoint.

I'm determining success by whether `result` in the returned JSON for the `data` block is true.

src/clients/Deluge.ts
zakkarry: Another oversight...will fix this :P

src/clients/TorrentClient.ts
mmgoodnow: FWIW, neither do any of the other torrent clients (maybe transmission does? idr). I set it up that way so as to have fewer options. In hindsight it hasn't been great UX (having to url encode your password is obviously stupid) but here we are

zakkarry: Addressed mostly in (https://github.com/cross-seed/cross-seed/pull/499/commits/39145621296eb067e346d1f4b9c97d82d3200193) I believe.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+		} else {
+			this.enLabeled = this.labelEnabled();
+		}
+	}
+
+	/**
+	 * connects and authenticates to the webui
+	 */
+	private async authenticate(): Promise<boolean> {
+		if (!this.loggedIn) {
```
mmgoodnow, line 51: I have a feeling that `this.loggedIn` is going to fall out of date, either if the deluge server restarts while cross-seed is running, or if we get a login timeout, etc. 

Can we instead assume that we're always logged in, but have login-and-retry logic if we get an auth error during any call? feel free to take a look at the qbit logic:

 https://github.com/cross-seed/cross-seed/blob/c012a088596fac654cc38ad8ce9bd89154f85bf7/src/clients/QBittorrent.ts#L160-L167

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+import { InjectionResult } from "../constants.js";
+import { CrossSeedError } from "../errors.js";
+import { Label, logger } from "../logger.js";
+import { Metafile } from "../parseTorrent.js";
+import { getRuntimeConfig } from "../runtimeConfig.js";
+import { Searchee } from "../searchee.js";
+import { TorrentClient } from "./TorrentClient.js";
+
+export default class Deluge implements TorrentClient {
+	private msgId = 0;
```
mmgoodnow, line 11: what's the deal with msgId? might be worth a comment

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+				method,
+			});
+
+		return response && response.data && response.data.error === null
+			? response.data
+			: response.data.error;
+	}
+
+	private async call(body: any): Promise<AxiosResponse | null> {
+		return new Promise((resolve) => {
```
mmgoodnow, line 104: no need to create a `new Promise`, instead we can await `axios.post()` (or fetch) and return the response

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+			});
+			this.loggedIn = !!response;
+		}
+		return this.loggedIn;
+	}
+
+	/**
+	 * authenticates and sends JSON-RPC calls to deluge
+	 */
+	private async sendCall(method: string, params: any[]) {
```
mmgoodnow, line 84: I think we can combine `call` and `sendCall` into one function

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+		} else
+			response = await this.call({
+				params: params,
+				method,
+			});
+
+		return response && response.data && response.data.error === null
+			? response.data
+			: response.data.error;
+	}
```
mmgoodnow, line 101: we have a little Result utility type in Result.ts if you're interested, it'd allow us to check for `isOk` and `isErr` rather than "result" and "message"

Otherwise I think I'd lean on `throw` and `catch` for errors here

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+
+	private async call(body: any): Promise<AxiosResponse | null> {
+		return new Promise((resolve) => {
+			body.id = ++this.msgId;
+			if (this.msgId > 1024) this.msgId = 0;
```
mmgoodnow, line 106: what happens if a call fails transiently (due to, say, DNS) and msgId gets out of sync? should we maybe wait until we see the response to increment msgId? 
```suggestion
			body.id = ++this.msgId;
			msgId %= 1024;
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+						headers: {
+							"Content-Type": "application/json",
+							Cookie: this.delugeCookie,
+						},
+					}
+				)
+				.then((data) => {
+					resolve(data);
+				})
+				.catch(() => resolve(null));
```
mmgoodnow, line 121: I'd rather rethrow than resolve with `null` and lose the error object here. 

lmk if you want some more detail on how Promise error handling works, also [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) are some high quality docs on promises 🙂

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+	/**
+	 * lists plugins and adds/sets labels
+	 * returns true if successful.
+	 */
+	private async labelEnabled() {
+		const enabledLabels = await this.sendCall(
+			"core.get_enabled_plugins",
+			[]
+		);
+		return enabledLabels["result"].includes("Label");
```
mmgoodnow, line 134: nit: anywhere we do `a["b"]`, we can also do `a.b` which is preferable. We can make similar changes in other places in this PR as well

```suggestion
		return enabledLabels.result.includes("Label");
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+	}
+	private async setLabel(torrenthash: string) {
```
mmgoodnow, line 136: nit: use `infoHash` for any info hash variable names

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+		if (this.enLabeled) {
+			let setResult = await this.sendCall("label.set_torrent", [
+				torrenthash,
+				this.delugeLabel,
+			]);
+			if (
+				setResult["message"] &&
+				setResult["message"].includes("Unknown Label")
```
mmgoodnow, line 144: is there like an `isError` boolean we can check?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+				await this.sendCall("label.add", [this.delugeLabel]);
+				await this.sendCall("label.set_torrent", [
+					torrenthash,
+					this.delugeLabel,
+				]);
+			}
+			return true;
+		} else {
+			return false;
+		}
```
mmgoodnow, line 155: do we need to return this boolean? might be able to get away with returning void

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+			{
+				file_priorities: [],
+				add_paused: false,
+				seed_mode: getRuntimeConfig().skipRecheck,
+				compact_allocation: true,
+				download_location: path,
+				max_connections: -1,
+				max_download_speed: -1,
+				max_upload_slots: -1,
+				max_upload_speed: -1,
```
mmgoodnow, line 209: are all of these options required? what happens if we just don't specify them at all?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	/**
+	 * returns true if the torrent hash is Seeding (completed)
+	 */
+	async checkCompleted(infohash: string): Promise<boolean> {
+		let params = [["name", "state", "save_path"], []];
+		let response;
+		try {
+			let response = await this.sendCall("web.update_ui", params);
```
mmgoodnow, line 222: is there a method we can use to get a single torrent's status by infoHash rather than receiving the entire catalog?

mmgoodnow (commented):
this is pretty exciting! Deluge is far and away the most requested feature so the work you're doing is going to be very impactful.

I think the architecture is really solid—mostly just some syntax, promise and edge case stuff can be improved.

mmgoodnow:
we'll also want to set up the `decodeURIComponent` stuff eventually

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+	private msgId = 0;
+	private loggedIn = false;
+	private delugeCookie = "";
+	private delugeWebUrl: URL;
+	private delugeLabel: string;
+	private enLabeled: Promise<boolean>;
+
+	constructor() {
+		this.delugeWebUrl = new URL(`${getRuntimeConfig().delugeWebUrl}`);
+		this.delugeLabel = getRuntimeConfig().dataCategory;
```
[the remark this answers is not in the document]
zakkarry: Agreed

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	/**
+	 * validates the login and host for deluge webui
+	 */
+	async validateConfig(): Promise<void> {
+		const url = new URL(this.delugeWebUrl);
+		if (url.username && !url.password) {
```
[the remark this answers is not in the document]
zakkarry: I believe this logic is best due to the way the delugeWebUrl needs to be structured to parse properly with URL() and redacting.

Usage is http://:password@localhost:8112 and conforms to URL's parsing password as the password, which leaves username undefined.

They should not be defining a username at all, and in most cases, I foresaw this as being a case of them forgetting the preceding ":" to distinguish that it is the password and not a username (as it would then be assumed as).

If they provide an user:pass - it will just use the password. So this seems to satisfy things the best way IMO.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+			throw new CrossSeedError(
+				"you need to define a password in the delugeWebUrl. (eg: http://:<PASSWORD>@localhost:8112)"
+			);
+		}
+		const result = await this.authenticate();
+		if (!result) {
+			logger.verbose({
+				label: Label.DELUGE,
+				message: "failed to authenticate the deluge daemon.",
+			});
```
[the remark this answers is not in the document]
zakkarry: Noted.

src/clients/Deluge.ts
zakkarry: Will apply as suggested.

src/clients/Deluge.ts
zakkarry: Not specifically, it will default to the client settings, or alternatively when the label is set, that label's settings.

I think seed_mode (skipRecheck) download_location (path) are the only config necessary ones. The rest should be client/label defaults.

src/clients/Deluge.ts
zakkarry: setResult is actually like a full stacktrace (in message) and a code that does I can't seem to locate any documentation on.

Normally, this would be done via a UI (either native or webui) where setting erroneous labels is impossible, so this was likely not taken into account. It does not crash anything though.

setResult response below.

```
{
  message: "Failure: [Failure instance: Traceback (failure with no frames): <class 'deluge.error.WrappedException'>: Unknown Label\n" +
    'Traceback (most recent call last):\n' +
    '  File "/usr/lib/python3.11/site-packages/deluge/core/rpcserver.py", line 334, in dispatch\n' +
    '    ret = self.factory.methods[method](*args, **kwargs)\n' +
    '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n' +
    '  File "/usr/lib/python3.11/site-packages/deluge/plugins/Label-0.3-py3.11.egg/deluge_label/core.py", line 317, in set_torrent\n' +
    "    check_input((not label_id) or (label_id in self.labels), _('Unknown Label'))\n" +
    '  File "/usr/lib/python3.11/site-packages/deluge/plugins/Label-0.3-py3.11.egg/deluge_label/core.py", line 65, in check_input\n' +
    '    raise Exception(message)\n' +
    'Exception: Unknown Label\n' +
    '\n' +
    ']',
  code: 4
}
```

src/clients/Deluge.ts
zakkarry: Yes, a filter_dict is a param I can send, oversight. Implementing.

src/clients/Deluge.ts
zakkarry: Latent code from some troubleshooting. It is obviously not being used and after call was functional I don't know why it was left.

zakkarry: Necessary for the axios implementation.

mmgoodnow: necessary for axios or necessary for deluge? I've used axios before and don't remember having to pass in an incrementing number

zakkarry: It's deluge-specific in this case.

src/clients/Deluge.ts
zakkarry: Implemented. Will be in next commit. 

Just double check my logic conforms to intended behaviors.

src/clients/Deluge.ts
zakkarry: I agree. But doing so with axios seems silly, so will wait to rewrite the functions with fetch.

src/clients/Deluge.ts
zakkarry: I guess this is redundant. I had the idea to forcibly enable labels plugin, and I think that's where this typing originated from.

src/clients/Deluge.ts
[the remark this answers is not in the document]
zakkarry: But if they forget the preceding colon in the link to specify it is a password to the URL constructor, password will be undefined and username will take the expected password value.

I was trying to prevent this misconfiguration/definition. I suppose it misses the case of neither being defined at all, though.

Technically, they should only be using http://:<password>@localhost:8112 as a url, with a preceding colon to conform to the constructor. We can either error out on anything else or add some handling for this.

Not sure about your preference.

src/clients/Deluge.ts
mmgoodnow: Let's remove the unnecessary ones

zakkarry: Already done in latest commit.

src/clients/Deluge.ts
zakkarry: completed in latest commit

src/clients/Deluge.ts
zakkarry: This does not affect command success.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,236 @@
...
+export default class Deluge implements TorrentClient {
+	private msgId = 0;
+	private loggedIn = false;
+	private delugeCookie = "";
+	private delugeWebUrl: URL;
+	private delugeLabel: string;
+	private enLabeled: Promise<boolean>;
+
+	constructor() {
+		this.delugeWebUrl = new URL(`${getRuntimeConfig().delugeWebUrl}`);
```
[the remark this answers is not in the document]
mmgoodnow:
```suggestion
		const { delugeWebUrl } = getRuntimeConfig(); 
		this.delugeWebUrl = new URL(delugeWebUrl);
```

src/clients/Deluge.ts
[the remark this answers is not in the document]
mmgoodnow: I think let's error just on no password. 
```suggestion
		if (!url.password) {
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+		this.delugeLabel = "cross-seed";
+	}
+
+	/**
+	 * validates the login and host for deluge webui
+	 */
+	async validateConfig(): Promise<void> {
+		const url = new URL(this.delugeWebUrl);
```
mmgoodnow, line 34: this.delugeWebUrl is already a URL instance here

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+export default class Deluge implements TorrentClient {
+	private msgId = 0;
+	private delugeCookie = "";
+	private delugeWebUrl: URL;
+	private delugeLabel: string;
+	private isLabelEnabled: Promise<boolean>;
+
+	constructor() {
+		this.delugeWebUrl = new URL(`${getRuntimeConfig().delugeWebUrl}`);
+		this.delugeLabel = "cross-seed";
```
mmgoodnow, line 27: can we inline `this.delugeLabel` and just put the string "cross-seed" wherever it's used?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+			);
+		}
+	}
+
+	/**
+	 * ensures authentication and sends JSON-RPC calls to deluge
+	 */
+	private async call(body: { method: string; params: any[] }) {
+		this.msgId = (this.msgId + 1) % 1024;
+
```
mmgoodnow, line 65: I still like sendCall's old signature `sendCall(method: string, params: any[])` and I think we can still preserve that now that we only have `call`. 

```suggestion
	private async call(body: { method: string; params: any[] }) {
		this.msgId = (this.msgId + 1) % 1024;
		const requestBody: JSON.stringify({ method, params, id: this.msgId }),
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+				body: JSON.stringify({ ...body, id: this.msgId }),
+				method: "POST",
+				headers: headers,
+			});
+			json = (await response.json()) as DelugeResponse;
+		} catch (networkError) {
+			return null;
+		}
+
+		if (json && json.error && json.error.code && json.error.code === 1) {
```
mmgoodnow, line 82: We can use [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) syntax here (also elsewhere):
```suggestion
		if (json?.error?.code === 1) {
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+				response = await fetch(url, {
+					body: JSON.stringify({ ...body, id: this.msgId }),
+					method: "POST",
+					headers: headers,
+				});
+			} catch (networkError) {
+				return null;
+			}
+			json = (await response.json()) as DelugeResponse;
+		}
```
mmgoodnow, line 94: instead of duplicating this fetch call, can we set up a retries system here? we can add a parameter to the end of `call` like so: `call(method: string, params: any[], retries = 1)`. 

Then, we can do a recursive call here: 
```js
if (json?.error?.code === 1 && retries > 0) {
	await this.authenticate();
	return this.call(method, params, 0);
}
```

BTW, this syntax with `retries = 0` is called [Default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+		if (!response.ok) {
+			return null;
+		}
+
+		if (response.headers && response.headers.has("Set-Cookie")) {
+			this.delugeCookie = response.headers
+				.get("Set-Cookie")
+				.split(";")[0];
+		} else if (body.method == "auth.login") {
```
mmgoodnow, line 103: I don't think we need this else if block

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+			json = (await response.json()) as DelugeResponse;
+		}
+		if (!response.ok) {
+			return null;
+		}
+
+		if (response.headers && response.headers.has("Set-Cookie")) {
+			this.delugeCookie = response.headers
+				.get("Set-Cookie")
+				.split(";")[0];
```
mmgoodnow, line 102: maybe we could extract this into a `handleResponseHeaders` function and call it like 
```js
this.handleResponseHeaders(response.headers);
```

in order to keep that complexity out of `call`.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+			});
+			return (
+				response.result &&
+				response.result.torrents &&
+				(response.result.torrents[searchee.infoHash].state ===
+					"Seeding" ||
+					(response.result.torrents[searchee.infoHash].state ===
+						"Paused" &&
+						response.result.torrents[searchee.infoHash].progress ===
+							100))
```
mmgoodnow, line 222: do we really need to check `state` if `progress` is 100? maybe we only need to check progress?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,232 @@
...
+		searchee: Searchee,
+		path?: string
+	): Promise<InjectionResult> {
+		if (searchee.infoHash && !(await this.checkCompleted(searchee))) {
+			return InjectionResult.TORRENT_NOT_COMPLETE;
+		}
+		const params = this.formatData(
+			`${newTorrent.name}.cross-seed.torrent`,
+			newTorrent.encode().toString("base64"),
+			path
```
mmgoodnow, line 164: unfortunately, `path` isn't guaranteed to exist (for torrent based matching, we query the client for `searchee.infoHash` to get its download location and mirror that on the new torrent

mmgoodnow (commented):
Things are coming along! I think the biggest remaining thing is support for torrent-based matching by querying the client for the source torrent's download dir

src/clients/Deluge.ts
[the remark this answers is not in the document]
mmgoodnow: ^ not really a big deal, it's just me being pedantic. However I started a branch with a universal url-with-urlencoded-credentials parsing function that I'm going to use to fix that bug with Transmission, at which point we probably won't need to store `delugeWebUrl` on `this` any more

src/clients/Deluge.ts
zakkarry: progress only returns if the state is not "Seeding" - so calling both will only return when progress is incomplete (while downloading or paused) - If paused, we cannot assume 100%, but it is an entirely possible scenario if seeding goals are set and met for that specific torrent on its tracker, therefore we would need to check progress agnostically, in that case, to cross-seed only completed.

src/clients/Deluge.ts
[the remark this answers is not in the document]
zakkarry: Makes sense...I understand consistency :)

src/clients/Deluge.ts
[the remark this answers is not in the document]
zakkarry: Got it.

mmgoodnow (commented):
The main functionality is looking pretty polished now! I'd say the theme of this review is error handling. It's looking really good in `inject`, and it could probably use a bit more detail in `call`. You have a couple ways of going about it - 
1. Use try/catch with a combination of `fetch`-thrown errors and your own error messages. In javascript, most of the time we just do `new Error("error message")`
2. Create an error enum and return a `Result<SuccessType,ErrorEnum>`. That can be good if you don't necessarily need to relay unknown errors from an upstream call, or you could try to predict all the failure modes ahead of time and write an `if/else` chain. Maybe not the best option for this specific case with `fetch`.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,226 @@
...
+			delugeRpcUrl
+		).unwrapOrThrow(
+			new CrossSeedError("delugeRpcUrl must be percent-encoded")
+		);
+		if (!password) {
+			throw new CrossSeedError(
+				"you need to define a password in the delugeRpcUrl. (eg: http://:<PASSWORD>@localhost:8112)"
+			);
+		}
+		const response = await this.call("auth.login", [password]);
```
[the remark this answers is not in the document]
zakkarry: Oh, I think that was the logic with the previous else if statement I used, which I removed prior to this migration to retry, and then never reimplemented.

But good catch.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,226 @@
...
+		return false;
+	}
+
+	/**
+	 * checks enabled plugins for "Label"
+	 * returns true if successful.
+	 */
+	private async labelEnabled() {
+		const enabledLabels = await this.call("core.get_enabled_plugins", []);
+		return enabledLabels?.result?.includes("Label");
```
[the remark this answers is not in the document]
zakkarry: I've yet to find any scenario in my testing in which this returns a non-array, but it seems like good logic to implement either way.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,226 @@
...
+		searchee: Searchee,
+		path?: string
+	): Promise<InjectionResult> {
+		if (searchee.infoHash && !(await this.checkCompleted(searchee))) {
+			return InjectionResult.TORRENT_NOT_COMPLETE;
+		}
+		const params = this.formatData(
+			`${newTorrent.name}.cross-seed.torrent`,
+			newTorrent.encode().toString("base64"),
+			path ? path : await this.getSavePath(searchee)
```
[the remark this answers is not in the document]
zakkarry: The thought had occurred to me, however, I looked at the qbit as an example and saw it chaining commands. 

Looking a little further now, I see that the endpoints are different, which probably explains this approach. I should have looked a little more. Will address this.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,226 @@
...
+		} catch (networkError) {
+			return null;
+		}
+		if (json?.error?.code === 1 && retries > 0) {
+			await this.authenticate();
+			return this.call(method, params, 0);
+		}
+		if (!response.ok) {
+			return null;
+		}
```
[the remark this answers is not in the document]
zakkarry: I tested a few common scenarios, the JSON-RPC endpoint seems to always return 200 with errors being reported in the actual JSON response.

When the URL is the main webui (not /json), it seems to fail with the "establish connection" error, which would be `response === null`. This could possibly be due to the content-type we use not being acceptable on the main webui, but I'm not sure about this and it's just my assumption.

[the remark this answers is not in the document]
mmgoodnow: hmm, thinking about it more, if you have Deluge behind nginx or something, maybe nginx could return non-200s. So probably good to have this branch here anyway

src/clients/Deluge.ts
```diff
@@ -0,0 +1,228 @@
...
+		);
+		const headers = new Headers({ "Content-Type": "application/json" });
+		if (this.delugeCookie) headers.set("Cookie", this.delugeCookie);
+
+		let response: Response, json: DelugeResponse;
+		try {
+			response = await fetch(href, {
+				body: JSON.stringify({
+					...{ method: method, params: params },
+					id: Math.floor(Math.random() * 1000000),
```
mmgoodnow, line 82: any reason not to use max int32?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,228 @@
...
+				response?.result?.torrents?.[searchee.infoHash]?.state ===
+					"Seeding" ||
+				response?.result?.torrents?.[searchee.infoHash]?.progress ===
+					100;
+			return {
+				complete: completedTorrent,
+				save_path: completedTorrent
+					? response?.result?.torrents?.[searchee.infoHash]?.save_path
+					: undefined,
+			} as TorrentInfo;
```
mmgoodnow, line 219: Typescript should be able to infer that this is a TorrentInfo since we're creating the object ourselves

```suggestion
			};
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,228 @@
...
+		} catch (e) {
+			logger.debug({
+				label: Label.DELUGE,
+				message: `failed to fetch torrent data from deluge: ${searchee.name} - ${searchee.infoHash}`,
+			});
+		}
+		return { complete: false } as TorrentInfo;
```
mmgoodnow, line 226: same here
```suggestion
		return { complete: false };
```

src/clients/Deluge.ts
```diff
@@ -0,0 +1,228 @@
...
+				save_path: completedTorrent
+					? response?.result?.torrents?.[searchee.infoHash]?.save_path
+					: undefined,
+			} as TorrentInfo;
+		} catch (e) {
+			logger.debug({
+				label: Label.DELUGE,
+				message: `failed to fetch torrent data from deluge: ${searchee.name} - ${searchee.infoHash}`,
+			});
+		}
```
mmgoodnow, line 225: maybe log an error here?

I'm not super satisfied with my logging patterns but generally the convention I've been using is that I log a string error message with `logger.error` and then do `logger.debug(e)` underneath it.

Also, it looks like this will fall through to the default return of `{complete: false}` which will trigger a TORRENT_NOT_COMPLETE error, which feels somewhat misleading

src/clients/Deluge.ts
```diff
@@ -0,0 +1,228 @@
...
+	/**
+	 * parses the set-cookie header and updates stored value
+	 */
+	private handleResponseHeaders(headers: Headers) {
+		if (headers?.has("Set-Cookie")) {
+			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
+			return true;
+		}
+		return false;
+	}
```
mmgoodnow, line 111: are we using these return values? if not, we can probably just return void here

mmgoodnow (commented):
Looking good! I've got some small style nits and a little bit of polishing on the new error handling code

src/clients/Deluge.ts
zakkarry: I guess I failed to consider that the torrent file may exist, while not being present in the client. 

For deluge, if torrentDir was set to the `state` directory, I think this is next to impossible. However, upon further thought, I do understand that this is not necessarily the configuration for everyone.

qBit's integration throw's a new error when this is the case, I assume that convention should carry over to Deluge?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,225 @@
...
+		} catch (NetworkError) {
+			throw new CrossSeedError(
+				`failed to establish a connection to deluge: ${href}`
+			);
+		}
+		if (json?.error?.code === DelugeErrorCode.NO_AUTH && retries > 0) {
+			this.delugeCookie = null;
+			await this.authenticate();
+			return this.call(method, params, 0);
+		}
```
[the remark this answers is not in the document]
zakkarry: I'm not sure if I handled this properly, since it would reset everytime a new command is issued.

Are we supposed to CrossSeedError after retries fail?

src/clients/Deluge.ts
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	/**
+	 * connects and authenticates to the webui
+	 */
+	private async authenticate(): Promise<void> {
+		const { delugeRpcUrl } = getRuntimeConfig();
+		const { href, password } = extractCredentialsFromUrl(
+			delugeRpcUrl
+		).unwrapOrThrow(new Error("delugeRpcUrl must be percent-encoded"));
```
mmgoodnow, line 54: sorry, I think I gave some bad advice in the last review. I think you were right originally with `CrossSeedError` on line 54, and then validateConfig doesn't need to wrap it in try/catch at all.

mmgoodnow (commented):
good job on the error handling—I think it's looking pretty robust. I just have a couple nits left, nothing very substantive.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,266 @@
...
+					: getRuntimeConfig().skipRecheck,
+				download_location: path,
+			},
+		];
+	}
+
+	/**
+	 * returns information needed to complete/validate injection
+	 */
+	private async getTorrentInfo(searchee: Searchee): Promise<TorrentInfo> {
```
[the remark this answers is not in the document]
zakkarry: I passed the entire Searchee structure for the error'ing since I provided the .name for clarity in catch.

src/clients/Deluge.ts
```diff
@@ -0,0 +1,266 @@
...
+			throw new Error(
+				"you need to define a password in the delugeRpcUrl. (e.g. http://:<PASSWORD>@localhost:8112)"
+			);
+		}
+		const response = await this.call("auth.login", [password], 0);
+		if (!response.result) {
+			throw new Error(
+				`reached deluge, but failed to authenticate: ${href}`
+			);
+		}
```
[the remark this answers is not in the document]
zakkarry: response.error is null.  Below is the output on a failed auth.login.

`{ result: false, error: null, id: 1373678230 }`

src/clients/Deluge.ts
```diff
@@ -0,0 +1,260 @@
...
+				await this.call("label.add", [label]);
+				await this.call("label.set_torrent", [infoHash, label]);
+			}
+		}
+	}
+
+	/**
+	 * injects a torrent into deluge client
+	 */
+	async inject(
```
mmgoodnow, line 152: can we wrap this whole function in a try/catch logging and returning FAILURE to make sure we don't crash the process?

mmgoodnow (approved):
LGTM pending the try/catch safety of `inject`. I think it might also be a good idea to test:
- injecting against preexisting torrents using labels with custom download dirs (this is a huge headache in qbit but hopefully the deluge impl is more sane)
- injecting against a single file torrent downloaded in "subfolder" or base path mode (I'm not sure of the terminology in deluge)
- injecting against a multi file torrent downloaded in "subfolder" mode
- injecting against a multi file torrent with one file downloaded in "subfolder" mode
- the above 3 but also _not_ in "subfolder mode"

zakkarry:
> LGTM pending the try/catch safety of `inject`. I think it might also be a good idea to test:
> 
>     * injecting against preexisting torrents using labels with custom download dirs (this is a huge headache in qbit but hopefully the deluge impl is more sane)

Do you mean when the label has a "Move on completion to this folder" setting? The status of the torrent should be "Moving" while this process is going on (Immediately after Downloading/Checking before "Seeding", which would - I guess - lead to a delay in the completed state being true or "Execute on completion" plugin running, rather than causing the Completion script to run before it is moved.

Since torrents that are not actually downloaded (cross-seeded) are not going to trigger this execution at all, this also would be a non-issue, but the one scenario it seems like a possible concern is if radarr (for instance) uses a post-import category that moves the torrent data. It then may create issues?

Will investigate this after work or this weekend.

>     * injecting against a single file torrent downloaded in "subfolder" or base path mode (I'm not sure of the terminology in deluge)
> 
>     * injecting against a multi file torrent downloaded in "subfolder" mode
> 
>     * injecting against a multi file torrent with one file downloaded in "subfolder" mode
> 
>     * the above 3 but also _not_ in "subfolder mode"

I'm assuming you mean the "Content Layout" option in qbittorrent being set to "Create Subfolder" - there is no equivalent option for this type of thing in Deluge - so that shouldn't be relevant. 

There is an old plugin for this behavior, however, it is relatively unknown (I didn't even know about it prior to a search on the forums, the post was in 2015) and I don't know if we should operate on the theory that we need to support every possible plugin? (https://forum.deluge-torrent.org/viewtopic.php?p=213539#p213539)

Will still look into this too, though.

mmgoodnow (approved):

mmgoodnow (dismissed):
LGTM

PR #503, opened 2023-09-30
Fix rTorrent injection incorrect paths

### LINKED ISSUE #502: rTorrent injection uses incorrect paths if torrent's name was not added to the original path
In rTorrent, there's an option to not use torrent's name in its path: `Don't add torrent's name to path`.
This results in `cross-seed` injecting torrents with incorrect paths, as it doesn't take that into account.

See the commands reference here: https://rtorrent-docs.readthedocs.io/en/latest/cmd-ref.html#term-d-directory
If the torrent has a single file, its name is not appended to `d.directory`, but if it has multiple files, the name is appended.
This is an inconsistency in the way the 'RTorrent' handler is using its API, even with the check for multiple files (`isMultiFileStr` https://github.com/cross-seed/cross-seed/blob/v5.5.1/src/clients/RTorrent.ts#L174C25-L174C39)

uraid:
Closes #502 .
Tested on:
1. Single-file unrenamed torrent
2. Multi-file unrenamed folder
3. Single-file renamed (without its name in path) torrent
4. Multi-file renamed (without its name in path) torrent

Also checks the data for torrent resume correctly.

### DIFF
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -1,5 +1,5 @@
 import { promises as fs, Stats } from "fs";
-import { basename, resolve, sep } from "path";
+import { dirname, join, resolve, sep } from "path";
 import { inspect } from "util";
 import xmlrpc, { Client } from "xmlrpc";
 import { InjectionResult } from "../constants.js";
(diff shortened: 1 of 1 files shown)


mmgoodnow:
I'm not sure this will work with data-based matching. Paths coming into `saveWithLibtorrentResume` don't come exclusively from the output of `checkOriginalTorrent`.

https://github.com/cross-seed/cross-seed/pull/503/files#diff-18ee5598520ab6145d25916e650421706b7947d1149326d2a3c517641676584bR233

mmgoodnow:
If you can modify the solution to support that, I will happily merge this PR. Thanks for the contribution!

uraid:
You're right - forgot that.
See modification :)

src/clients/RTorrent.ts
```diff
@@ -29,14 +29,19 @@ async function createLibTorrentResumeTree(
 	async function getFileResumeData(
 		file: File
 	): Promise<LibTorrentResumeFileEntry> {
-		const filePath = resolve(dataDir, file.path);
+		let filePath = file.path;
+		if (filePath.indexOf(sep) != -1) {
+			filePath = filePath.split(sep).slice(1).join(sep);
+		}
+
+		const normalizedFilePath = resolve(dataDir, filePath);
```
zakkarry, line 37: Forgive my ignorance, I don't have this setup to test this myself, but does this not make an assumption that the path is at a depth that can't be guaranteed (1)

Does this function as intended if I used (for example) "/media/downloads/torrents" as my default directory for downloads?

uraid: This function tries to split a filename inside a torrent (for torrents with folders). For example, if you have `BLURAY_DISC1/BDMV/index.json` file inside the torrent, it would slice the top folder and filePath would be `BDMV/index.json`.

It then adds the dataDir, in your example `/media/download/torrents/SOME_OTHER_PATH_WE_CHOSE_TO_MOVE_IT`.
The final checked path would be: `/media/download/torrents/SOME_OTHER_PATH_WE_CHOSE_TO_MOVE_IT/BDMV/index.json`

My example is for a renamed torrent, but in the case the torrent has its normal name and was not renamed, it would check: `/media/download/torrents/BLURAY_DISC1/BDMV/index.json` path.

mmgoodnow:
I think data-based matching support will need another change to how we get the libtorrent resume information. 


In this call to `saveWithLibTorrentResume`, you changed the value of `downloadDir` to mean `basePath` but we're still passing a non-base path in with `path`. 
https://github.com/cross-seed/cross-seed/blob/ef83f9065dc470196e90c60fdb48059f206a0de2/src/clients/RTorrent.ts#L227-L227

uraid:
@mmgoodnow Fixed it. Added a boolean to these functions.
I also changed `path ? path : downloadDir` to just be `downloadDir`, as `downloadDir` will [always have the correct value](https://github.com/cross-seed/cross-seed/pull/503/commits/2a7599c2632f0da9e53bdc8772a24de3e366465e#diff-18ee5598520ab6145d25916e650421706b7947d1149326d2a3c517641676584bR216)

mmgoodnow:
Thanks for the contribution!

PR #514, opened 2023-10-03
Handle URL paths better

mmgoodnow:
added a `basePath` argument to `extractCredentialsFromURL` that should hopefully prevent us from making the same mistake I made again.

```js
> extractCredentialsFromUrl('http://admin:adminadmin@localhost:8080/qbit')
OkResult {
  contents: {
    username: 'admin',
    password: 'adminadmin',
    href: 'http://localhost:8080/qbit'
  }
}
> extractCredentialsFromUrl('http://admin:adminadmin@localhost:8080')
OkResult {
  contents: {
    username: 'admin',
    password: 'adminadmin',
    href: 'http://localhost:8080'
  }
}
> extractCredentialsFromUrl('http://admin:adminadmin@localhost:8080/qbit', '/api/v2')
OkResult {
  contents: {
    username: 'admin',
    password: 'adminadmin',
    href: 'http://localhost:8080/qbit/api/v2'
  }
}
> extractCredentialsFromUrl('http://admin:adminadmin@localhost:8080', '/api/v2')
OkResult {
  contents: {
    username: 'admin',
    password: 'adminadmin',
    href: 'http://localhost:8080/api/v2'
  }
}
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -105,7 +105,7 @@ export function extractCredentialsFromUrl(
 			username: decodeURIComponent(username),
 			password: decodeURIComponent(password),
 			href: basePath
-				? origin + join(pathname, basePath)
+				? origin + path.posix.join(pathname, basePath)
 				: pathname === "/"
 				? origin
 				: origin + pathname,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-import { basename } from "path";
+import path, { basename, join } from "path";
 import {
 	EP_REGEX,
 	MOVIE_REGEX,
@@ -96,14 +96,19 @@ export function fallback<T>(...args: T[]): T {
 }
 
 export function extractCredentialsFromUrl(
-	url: string
+	url: string,
+	basePath?: string
 ): Result<{ username: string; password: string; href: string }, "invalid URL"> {
 	try {
 		const { origin, pathname, username, password } = new URL(url);
 		return resultOf({
 			username: decodeURIComponent(username),
 			password: decodeURIComponent(password),
-			href: origin + pathname,
+			href: basePath
+				? origin + join(pathname, basePath)
+				: pathname === "/"
+				? origin
+				: origin + pathname,
 		});
 	} catch (e) {
 		return resultOfErr("invalid URL");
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { basename } from "path";
+import path, { basename, join } from "path";
```
zakkarry, line 1:
```suggestion
import path, { basename, join } from "path";
import { getRuntimeConfig } from "../runtimeConfig.js";
```

Imports getRuntimeConfig() - necessary to find if the client requires URIDecodeComponent (for future proofing since deluge does not require this)

src/utils.ts
```diff
@@ -96,14 +96,19 @@ export function fallback<T>(...args: T[]): T {
...
 export function extractCredentialsFromUrl(
-	url: string
+	url: string,
+	basePath?: string
 ): Result<{ username: string; password: string; href: string }, "invalid URL"> {
 	try {
 		const { origin, pathname, username, password } = new URL(url);
 		return resultOf({
 			username: decodeURIComponent(username),
 			password: decodeURIComponent(password),
```
zakkarry, line 106:
```suggestion
	const { rtorrentRpcUrl, qbittorrentUrl, transmissionRpcUrl } =
		getRuntimeConfig();
	try {
		const { origin, pathname, username, password } = new URL(url);
		return resultOf({
			username: (rtorrentRpcUrl || qbittorrentUrl || transmissionRpcUrl) ? decodeURIComponent(username) : username,
			password: (rtorrentRpcUrl || qbittorrentUrl || transmissionRpcUrl) ? decodeURIComponent(password) : password,
```

Checks returned getRuntimeConfig() values for which client (once again, deluge not requiring decodeURIComponent() means this should be future-proofed)

Will require adding any other future clients requiring decodeURIComponent to this section, though.

Not sure if there is a more preferred way, this was just something that occurred to me and the first method to accomplish this.

zakkarry (changes requested):

src/utils.ts
mmgoodnow: we will need it for Deluge

mmgoodnow: The reason we need it for all the other clients isn't due to the way that auth works in those clients, it's due to the way the URL constructor works in Node. 
image

So, for the same reasons we need it for other clients, we'll need it for Deluge.

zakkarry: I took a further look into what this is actually doing and accomplishing, and was confused - and unaware of ERR_INVALID_URL. 

You're correct. Resolved.

PR #517, opened 2023-10-07
fix(logging/config): validates config file action methods

### LINKED ISSUE #478: Silent Failing when inject is misspelled.
When the action is typoed (in my case I did injest not inject) it silently failed and defaulted to save

```javascript
  /**
     * With "inject" you need to set up one of the below clients.
     * Options: "save", "inject"
     */
    action: "injest",
    /**
``` 
vs 

```javascript
  /**
     * With "inject" you need to set up one of the below clients.
     * Options: "save", "inject"
     */
    action: "inject",
    /**
```

zakkarry:
will not silently proceed to save if erroneous values are set for action in the config file.

- Fixes #478

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -23,6 +23,16 @@ import { getRuntimeConfig } from "./runtimeConfig.js";
 import { Searchee } from "./searchee.js";
 import { saveTorrentFile } from "./torrent.js";
 import { getTag } from "./utils.js";
+import { CrossSeedError } from "./errors.js";
+
+export async function validateAction(): Promise<void> {
+	const { action } = getRuntimeConfig();
+	if (action !== Action.INJECT && action !== Action.SAVE) {
+		throw new CrossSeedError(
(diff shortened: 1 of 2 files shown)


src/torrent.ts
```diff
@@ -168,6 +168,18 @@ export async function getInfoHashesToExclude(): Promise<string[]> {
 	);
 }
 
+export async function validateAction(): Promise<void> {
```
mmgoodnow, line 171: There might be a better home for this file—maybe `action.ts` or somewhere else we have a lot of validation code?

mmgoodnow (commented):
this feels like a somewhat narrow fix. Yes it fixes when someone misspells `action` but they could also misspell `matchMode`, or makes `dataDirs` a string instead of a string[], etc. 

I think something like using [`zod`](https://zod.dev) in a function like `validateConfig` could be a more holistic approach.

src/torrent.ts
zakkarry: Moved to action.ts

PR #518, opened 2023-10-08
feat(daemon): add API auth

### LINKED ISSUE #306: api auth

zze0s:
Hey!

This is simple implementation of adding auth to the API endpoints.

The apiKey can be set as:
 - query param `/api/webhook?apikey=my-secret-key`
 - as a header `X-API-KEY`.

If the `apikey` is set in the config file and it does not match it responds with `401 Unauthorized`.

Could be improved with a micro framework like Express or Fastify or similar that implements middlewares, but this is simple enough for the current 2 endpoints of the project.

The global var `secretApiKey` in `server.ts` might not be the best solution, but it works and makes it available to the needed functions.

:white_check_mark: Tested and it works as intended

Fixes #306

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -181,6 +181,12 @@ module.exports = {
 	host: undefined,
 
 	/**
+	 * Force authentication for API
+	 * Example: "my-long-good-token"
+	 */
+	apikey: undefined,
+
+	/**
 	 * Run rss scans on a schedule. Format: https://github.com/vercel/ms
 	 * Set to undefined or null to disable. Minimum of 10 minutes.
 	 * Examples:
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -181,10 +181,11 @@ module.exports = {
 	host: undefined,
 
 	/**
-	 * Force authentication for API
-	 * Example: "my-long-good-token"
+	 * Whether to require authentication for API.
+	 * Run the command `cross-seed api-key` to find your api key.
+	 * Keys can be provided in an X-Api-Key HTTP header or a query param.
 	 */
-	apikey: undefined,
+	apiAuth: true,
 
 	/**
 	 * Run rss scans on a schedule. Format: https://github.com/vercel/ms
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -3,7 +3,7 @@ import { db } from "./db.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 
 function generateApiKey(): string {
-	return randomBytes(16).toString("hex");
+	return randomBytes(24).toString("hex");
 }
 
 export async function resetApiKey(): Promise<string> {
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -3,7 +3,7 @@ import { db } from "./db.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 
 function generateApiKey(): string {
-	return randomBytes(32).toString("hex");
+	return randomBytes(16).toString("hex");
 }
 
 export async function resetApiKey(): Promise<string> {
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -19,8 +19,8 @@ export async function getApiKeyFromDatabase(): Promise<string> {
 }
 
 export async function checkApiKey(keyToCheck: string): Promise<boolean> {
-	const { auth } = getRuntimeConfig();
-	if (!auth) return true;
+	const { apiAuth } = getRuntimeConfig();
+	if (!apiAuth) return true;
 	const apikey = await getApiKeyFromDatabase();
 	return apikey === keyToCheck;
 }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,0 +1,1 @@
+import { db } from "./db.js";
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,1 +1,26 @@
+import { randomBytes } from "node:crypto";
 import { db } from "./db.js";
+import { getRuntimeConfig } from "./runtimeConfig.js";
+
+function generateApiKey(): string {
+	return randomBytes(32).toString("hex");
+}
+
+export async function resetApiKey(): Promise<string> {
+	const apikey = generateApiKey();
+	await db("settings").update({ apikey });
+	return apikey;
+}
+
+export async function getApiKeyFromDatabase(): Promise<string> {
+	const { apikey } = await db("settings").select("apikey").first();
+	if (!apikey) return resetApiKey();
+	return apikey;
+}
+
+export async function checkApiKey(keyToCheck: string): Promise<boolean> {
+	const { auth } = getRuntimeConfig();
+	if (!auth) return true;
+	const apikey = await getApiKeyFromDatabase();
+	return apikey === keyToCheck;
+}
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -309,7 +309,11 @@ createCommandWithSharedOptions("daemon", "Start the cross-seed daemon")
 		fallback(fileConfig.port, 2468)
 	)
 	.option("--host <host>", "Bind to a specific IP address", fileConfig.host)
-	.option("--api-auth", "Require API auth via API key", fileConfig.apiAuth)
+	.option(
+		"--api-auth",
+		"Require API auth via API key",
+		fallback(fileConfig.apiAuth, false)
+	)
 	.option("--no-api-auth", "Don't require API auth")
 	.option("--no-port", "Do not listen on any port")
 	.option(
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -291,6 +291,7 @@ program
 	.action(async () => {
 		await db.migrate.latest();
 		console.log(await getApiKeyFromDatabase());
+		await db.destroy();
 	});
 
 program
@@ -299,6 +300,7 @@ program
 	.action(async () => {
 		await db.migrate.latest();
 		console.log(await resetApiKey());
+		await db.destroy();
 	});
 
 createCommandWithSharedOptions("daemon", "Start the cross-seed daemon")
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -3,6 +3,7 @@ import chalk from "chalk";
 import { Option, program } from "commander";
 import ms from "ms";
 import { inspect } from "util";
+import { getApiKeyFromDatabase, resetApiKey } from "./auth.js";
 import { generateConfig, getFileConfig } from "./configuration.js";
 import {
 	Action,
@@ -282,6 +283,22 @@ program
 		console.log(
 			createSearcheeFromMetafile(await parseTorrentFromFilename(fn))
 		);
+	});
+
+program
+	.command("api-key")
+	.description("Show the api key")
+	.action(async () => {
+		await db.migrate.latest();
+		console.log(await getApiKeyFromDatabase());
+	});
+
+program
+	.command("reset-api-key")
+	.description("Reset the api key")
+	.action(async () => {
+		await db.migrate.latest();
+		console.log(await resetApiKey());
 	});
 
 createCommandWithSharedOptions("daemon", "Start the cross-seed daemon")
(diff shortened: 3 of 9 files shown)


src/config.template.cjs
```diff
@@ -180,6 +180,12 @@ module.exports = {
 	 */
 	host: undefined,
 
+	/**
+	 * Force authentication for API
```
bakerboy448, line 184: Alphanumeric? Dashes? How long?

zakkarry:
```diff
diff --git a/src/cmd.ts b/src/cmd.ts
index 90a52b8..9f0d3ef 100755
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -317,7 +317,7 @@ createCommandWithSharedOptions("daemon", "Start the cross-seed daemon")
 			});
 			await db.migrate.latest();
 			await doStartupValidation();
-			serve(options.port, options.host, options.apikey);
+			serve(options.port, options.host);
 			jobsLoop();
 		} catch (e) {
 			exitOnCrossSeedErrors(e);
diff --git a/src/constants.ts b/src/constants.ts
index 67c062d..755930d 100644
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -6,6 +6,7 @@ export const PROGRAM_NAME = packageDotJson.name;
 export const PROGRAM_VERSION = packageDotJson.version;
 export const USER_AGENT = `CrossSeed/${PROGRAM_VERSION}`;
 
+export const APIKEY_REGEX = /^[a-z0-9]{20,32}$/i;
 export const EP_REGEX =
 	/^(?<title>.+?)[\s._](?<season>S\d+)?[_.\s]?(?<episode>E\d+(?:[-\s]?E?\d+)?)/i;
 export const SEASON_REGEX =
diff --git a/src/runtimeConfig.ts b/src/runtimeConfig.ts
index 4f6576d..84e3aa7 100644
--- a/src/runtimeConfig.ts
+++ b/src/runtimeConfig.ts
@@ -28,6 +28,7 @@ export interface RuntimeConfig {
 	notificationWebhookUrl: string;
 	torrents: string[];
 	port: number | false | null;
+	apikey: string;
 	searchCadence: number;
 	rssCadence: number;
 	snatchTimeout: number;
diff --git a/src/server.ts b/src/server.ts
index 24a30f5..348da53 100644
--- a/src/server.ts
+++ b/src/server.ts
@@ -10,8 +10,22 @@ import {
 	searchForLocalTorrentByCriteria,
 } from "./pipeline.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
+import { APIKEY_REGEX } from "./constants.js";
+import { getRuntimeConfig } from "./runtimeConfig.js";
+import { CrossSeedError } from "./errors.js";
 
-let secretApiKey = "";
+/**
+ * validates that the apikey is 20-32 characters alphanumeric
+ * throws and exits if not
+ */
+export function validateApiKey() {
+	const { apikey } = getRuntimeConfig();
+	if (!APIKEY_REGEX.test(apikey)) {
+		throw new CrossSeedError(
+			'You need to set a valid "apikey:" value in your config or specify --apikey (20-32 character alphanumeric)'
+		);
+	}
+}
 
 function getData(req): Promise<string> {
 	return new Promise((resolve) => {
@@ -50,11 +64,7 @@ function parseData(data) {
 
 // isValidAPIKey check if api key is valid or not
 function isValidAPIKey(req: IncomingMessage): boolean {
-	// if no api key is set then always return true
-	if (secretApiKey === "") {
-		return true;
-	}
-
+	const { apikey } = getRuntimeConfig();
 	const parsedUrl = url.parse(req.url, true);
 
 	// get apikey from query /params?apikey=supersecretapikey
@@ -63,7 +73,7 @@ function isValidAPIKey(req: IncomingMessage): boolean {
 	// get apikey from header
 	const apiKeyHeader = req.headers["x-api-key"];
 
-	if (apiKeyQueryParam === secretApiKey || apiKeyHeader === secretApiKey) {
+	if (apiKeyQueryParam === apikey || apiKeyHeader === apikey) {
 		return true; // API key is valid.
 	}
 
@@ -221,8 +231,8 @@ async function handleRequest(
 		res.end("Methods allowed: POST");
 		return;
 	}
-
-	switch (req.url) {
+	//take the url before the parameters (?)
+	switch (req.url.split("?")[0]) {
 		case "/api/webhook": {
 			logger.verbose({
 				label: Label.SERVER,
@@ -246,15 +256,7 @@ async function handleRequest(
 	}
 }
 
-export function serve(
-	port: number,
-	host: string | undefined,
-	apiKey: string | undefined
-): void {
-	if (apiKey) {
-		secretApiKey = apiKey;
-	}
-
+export function serve(port: number, host: string | undefined): void {
 	if (port) {
 		const server = http.createServer(handleRequest);
 		server.listen(port, host);
diff --git a/src/startup.ts b/src/startup.ts
index 76cfb40..27665db 100644
--- a/src/startup.ts
+++ b/src/startup.ts
@@ -5,6 +5,7 @@ import { logger } from "./logger.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 import { validateTorrentDir } from "./torrent.js";
 import { validateTorznabUrls } from "./torznab.js";
+import { validateApiKey } from "./server.js";
 
 function validateOptions() {
 	const {
@@ -43,6 +44,7 @@ export async function doStartupValidation(): Promise<void> {
 	const downloadClient = getClient();
 	await Promise.all<void>(
 		[
+			validateApiKey(),
 			validateTorznabUrls(),
 			downloadClient?.validateConfig(),
 			validateTorrentDir(),
```

This fixes a few bugs, mainly when using webhook?apikey= it wouldn't match the switch because the req.url included that parameter. It also removes the secretApiKey variable and uses getRunTimeConfig() to pull the value, validates the apikey being alphanumeric and between 20 and 32 characters at startup (as suggested by @mynameisbogdan and @bakerboy448), and brings some of the changes into line with the structure of the codebase.

One caveat that needs to be addressed is how we handle the introduction of auth because, with my diff, this will essentially break everyone's instance until they add the value to the config. Even if we decide to generate an apikey if one is absent, without the knowledge of the user their announce/webhook calls will be rejected. 

This (erroring without an apikey defined) may be an inconvenience to many, but I think it is the best approach. The error from validateApiKey() states what the issue is and what needs to be defined.

mmgoodnow:
thanks for making this PR - I remember our conversation a while back about it. I haven't read it yet but am hoping to get a chance in the next few days

mmgoodnow:
I think it would make sense to store the API key in `/config/apikey.txt` or in the database (I'm leaning towards the database). That way we don't have to make the user generate it themselves, and we also won't need to go through the hassle of validating it. 

I'm just a little bit unsure of how to expose it to the user - maybe a pair of commands 
`cross-seed api-key` and `cross-seed reset-api-key`? Or we could just tell people to look at `/config/apikey.txt` I guess. I'm open to opinions there. 

@zze0s, let me know if you would like to try to make those changes or if you would like me to take things over from here. Either way, this PR has served as a good kick in the pants to prioritize auth.

zakkarry:
> I think it would make sense to store the API key in `/config/apikey.txt` or in the database (I'm leaning towards the database). That way we don't have to make the user generate it themselves, and we also won't need to go through the hassle of validating it.
> 
> I'm just a little bit unsure of how to expose it to the user - maybe a pair of commands `cross-seed api-key` and `cross-seed reset-api-key`? Or we could just tell people to look at `/config/apikey.txt` I guess. I'm open to opinions there.

I think it would be best served from a security standpoint to store it in the db, but the downside is the frequency in which people, at least myself, may delete their cross-seed.db to reset things. It seems to be a common practice amongst many, and losing the apikey and having to reconfigure scripts/commands in this event may be somewhat cumbersome.

I don't really agree that storing it in plaintext is ideal....although the risk seems minimal.

mmgoodnow:
the DB would have it in plaintext too. I think most other apps in the space also store API keys in plain text because they have to let you copy them from the UI.

bakerboy448:
Accurately perhaps scope creep, but add a db reset function that purges history but keeps the api key and other config?

mmgoodnow:
> Accurately perhaps scope creep, but add a db reset function that purges history but keeps the api key and other config?
We sort of have that in `clear-cache` but it probably doesn't delete enough stuff. we could stand to make `clear-cache` more aggressive for sure.

zakkarry:
> the DB would have it in plaintext too. I think most other apps in the space also store API keys in plain text because they have to let you copy them from the UI.

Of course, but it's not just blatantly in a text file, so a little bit more security by obscurity in my opinion.

zze0s:
> thanks for making this PR - I remember our conversation a while back about it. I haven't read it yet but am hoping to get a chance in the next few days

Happy to do it! Yeah that was a while ago :smile: 

> I think it would make sense to store the API key in `/config/apikey.txt` or in the database (I'm leaning towards the database). That way we don't have to make the user generate it themselves, and we also won't need to go through the hassle of validating it.
> 
> I'm just a little bit unsure of how to expose it to the user - maybe a pair of commands `cross-seed api-key` and `cross-seed reset-api-key`? Or we could just tell people to look at `/config/apikey.txt` I guess. I'm open to opinions there.
> 
> @zze0s, let me know if you would like to try to make those changes or if you would like me to take things over from here. Either way, this PR has served as a good kick in the pants to prioritize auth.

Good to get your feedback on this! Me and zakkaryry discussed the PR and he came up with the proposed improvements. 

The initial idea was to make it as simple as possible, but having it in the db would indeed be the better solution. If you have the time to take over then please do! Maintainers can edit the PR :+1:

src/config.template.cjs
mmgoodnow: we're using `crypto.randomBytes(24).toString('hex')` now. So the key is 48 characters long, but there are 2^192 possible values.

mmgoodnow:
ok, as of https://github.com/cross-seed/cross-seed/pull/518/commits/532a167c183ec5d166c1b40e6a737ef98127c7e3 I think everything's working properly. The config is `--api-auth` as a CLI argument, or `apiAuth: boolean` in the config file. We have a `cross-seed api-key` command and a `cross-seed reset-api-key` command as the configuration interface for the key itself. 

I think for now I don't want to bump the major version so I'll leave the config template value as `true` which will turn it on for new users, but I'll `fallback` to `false` if undefined meaning it won't be a breaking change for existing users. Next time we bump the major version we can change the fallback value to require it for everyone.

zakkarry:
> ok, as of [532a167](https://github.com/cross-seed/cross-seed/commit/532a167c183ec5d166c1b40e6a737ef98127c7e3) I think everything's working properly. The config is `--api-auth` as a CLI argument, or `apiAuth: boolean` in the config file. We have a `cross-seed api-key` command and a `cross-seed reset-api-key` command as the configuration interface for the key itself.
> 
> I think for now I don't want to bump the major version so I'll leave the config template value as `true` which will turn it on for new users, but I'll `fallback` to `false` if undefined meaning it won't be a breaking change for existing users. Next time we bump the major version we can change the fallback value to require it for everyone.

Will try and do basic testing on migration and fresh setup today

zze0s:
> ok, as of [532a167](https://github.com/cross-seed/cross-seed/commit/532a167c183ec5d166c1b40e6a737ef98127c7e3) I think everything's working properly. The config is `--api-auth` as a CLI argument, or `apiAuth: boolean` in the config file. We have a `cross-seed api-key` command and a `cross-seed reset-api-key` command as the configuration interface for the key itself.
> 
> I think for now I don't want to bump the major version so I'll leave the config template value as `true` which will turn it on for new users, but I'll `fallback` to `false` if undefined meaning it won't be a breaking change for existing users. Next time we bump the major version we can change the fallback value to require it for everyone.

Awesome!

onedr0p:
It would be nice if the user could override the api key themselves in the config file. Maybe I drink too much of the 12 factor app koolaid but having this set via the config file is great for automation as the key will always be predictable for the user deploying the application.

mmgoodnow:
I think having the key itself be configurable is not going to be a goal of this feature. API keys are pretty much always generated by the application, so we're at least in good company. 

Status update: I will merge this when the documentation is ready.

mmgoodnow:
v5.8.0

PR #523, opened 2023-10-14
findOnOtherSites only completes if the torrent is complete

ColinHebert:
This changes the logic of `findOnOtherSites` to ensure that the search is considered as "complete" only when the torrent itself is complete.

This avoids issues where the torrent is not complete yet the search is marked as successful therefore there won't be a search done for the torrent once it has completed.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -113,6 +113,17 @@ async function findOnOtherSites(
 		}
 	);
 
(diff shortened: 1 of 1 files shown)


mmgoodnow (approved):

mmgoodnow:
Thanks for the contribution!

PR #526, opened 2023-10-18
feat: announce - return a 202 on torrent not completed

### LINKED ISSUE #525: Expose different status codes for different responses
For example, say a torrent is a potential xseed but it's not done in the client yet. We should respond with a different status code vs the default 204.

buroa:
Closes #525

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -8,6 +8,7 @@ import {
 	checkNewCandidateMatch,
 	searchForLocalTorrentByCriteria,
 } from "./pipeline.js";
+import { InjectionResult, SaveResult } from "./constants.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
 
 function getData(req): Promise<string> {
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -152,18 +156,21 @@ async function announce(
...
 		await indexNewTorrents();
 		const result = await checkNewCandidateMatch(candidate);
-		if (!result) {
-			res.writeHead(204);
-			res.end();
-			return;
+		const isOk = result === InjectionResult.SUCCESS || result === SaveResult.SAVED;
+		if (!isOk) {
+			if (result === InjectionResult.TORRENT_NOT_COMPLETE) {
+				res.writeHead(202);
```
mmgoodnow, line 162: why 202?

buroa: It looked like the best one for this case.

"Accepted response status code indicates that the request has been accepted for processing, but the processing has not been completed; in fact, processing may not have started yet."

If you wanted to use a different one, I'm all up for it.

onedr0p (approved):

PR #545, opened 2023-11-09
refactor(client/deluge): refactor and typing for json-rpc integration

ColinHebert:
This is a simple code-fix pull request. It doesn't provide functional changes.

The focus is code format and syntactic validity

edit: Not so much simple code-fix anymore :P // @zakkarry 

## Summary by CodeRabbit

- **Refactor**
  - Improved the search and labeling features with enhanced error handling and clearer type definitions.
  
- **New Features**
  - Introduced new types for better data handling and user experience.

- **Bug Fixes**
  - Enhanced error handling in various methods to provide more reliable performance.

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -140,9 +140,7 @@ export default class Deluge implements TorrentClient {
 	 * parses the set-cookie header and updates stored value
 	 */
 	private handleResponseHeaders(headers: Headers) {
-		if (headers.has("Set-Cookie")) {
-			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
-		}
+		this.delugeCookie = headers.get("Set-Cookie")?.split(";")[0] ?? null;
 	}
 
 	/**
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -207,17 +207,6 @@ export default class Deluge implements TorrentClient {
 			if (searchee.infoHash) {
 				torrentInfo = await this.getTorrentInfo(searchee);
 				if (!torrentInfo.complete) {
-					if (
-						torrentInfo.save_path == "missing" &&
-						!path &&
-						!searchee.path
-					) {
-						logger.debug({
-							label: Label.DELUGE,
-							message: `Injection failure: ${searchee.name} (${searchee.infoHash}) was not found in the client.`,
-						});
-						return InjectionResult.FAILURE;
-					}
 					return InjectionResult.TORRENT_NOT_COMPLETE;
 				}
 			}
@@ -360,7 +349,8 @@ export default class Deluge implements TorrentClient {
 				message: `Failed to fetch torrent data: ${searchee.name} - (${searchee.infoHash})`,
 			});
 			logger.debug(e);
-			return { complete: false, save_path: "missing" };
+			throw new Error("web.update_ui: failed to fetch data from client");
+			//return { complete: false, save_path: "missing" };
 		}
 	}
 }
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -131,7 +131,8 @@ export default class Deluge implements TorrentClient {
 		try {
 			response = await fetch(href, {
 				body: JSON.stringify({
-					...{ method, params },
+					method,
+					params,
 					id,
 				}),
 				method: "POST",
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -22,22 +22,23 @@ enum DelugeErrorCode {
 	RPC_FAIL = 4,
 	BAD_JSON = 5,
 }
-type injectData = [
-	string,
-	string,
-	{
+type InjectData = [
+	filename: string,
+	filedump: string,
+	options: {
 		add_paused: boolean;
 		seed_mode: boolean;
 		download_location: string;
 	}
 ];
+type WebHostList = [string, string, number, string][];
+type ErrorType = { message?: string; code?: DelugeErrorCode };
+type TorrentStatus = { torrents?: Record<string, TorrentInfo> };
 
 type DelugeJSON<ResultType> = {
 	result?: ResultType;
 	error?: ErrorType;
 };
-type ErrorType = { message?: string; code?: DelugeErrorCode };
-type TorrentStatus = { torrents?: Record<string, TorrentInfo> };
 
 export default class Deluge implements TorrentClient {
 	private delugeCookie: string | null = null;
@@ -91,7 +92,7 @@ export default class Deluge implements TorrentClient {
 			logger.warn(
 				"Deluge WebUI disconnected from daemon...attempting to reconnect."
 			);
-			const webuiHostList = await this.call<object>(
+			const webuiHostList = await this.call<WebHostList>(
 				"web.get_hosts",
 				[],
 				0
@@ -116,7 +117,7 @@ export default class Deluge implements TorrentClient {
 	 */
 	private async call<ResultType>(
 		method: string,
-		params: object,
+		params: unknown[],
 		retries = 1
 	): Promise<DelugeJSON<ResultType>> {
 		const { delugeRpcUrl } = getRuntimeConfig();
@@ -168,9 +169,9 @@ export default class Deluge implements TorrentClient {
 	 * parses the set-cookie header and updates stored value
 	 */
 	private handleResponseHeaders(headers: Headers) {
-		this.delugeCookie = headers.has("Set-Cookie")
-			? headers.get("Set-Cookie").split(";")[0]
-			: null;
+		if (headers.has("Set-Cookie")) {
+			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
+		}
 	}
 
 	/**
@@ -182,7 +183,9 @@ export default class Deluge implements TorrentClient {
 			"core.get_enabled_plugins",
 			[]
 		);
-		return enabledPlugins ? enabledPlugins.result.includes("Label") : false;
+		return enabledPlugins.error
+			? false
+			: enabledPlugins.result.includes("Label");
 	}
 
 	/**
@@ -193,11 +196,7 @@ export default class Deluge implements TorrentClient {
 		const setResult = this.isLabelEnabled
 			? await this.call<void>("label.set_torrent", [infoHash, label])
 			: undefined;
-		if (!setResult) {
-			throw new Error(
-				"label.set_torrent: Client returned empty response"
-			);
-		} else if (setResult.error?.code == DelugeErrorCode.RPC_FAIL) {
+		if (setResult?.error?.code === DelugeErrorCode.RPC_FAIL) {
 			await this.call<void>("label.add", [label]);
 			await this.call<void>("label.set_torrent", [infoHash, label]);
 		}
@@ -301,7 +300,7 @@ export default class Deluge implements TorrentClient {
 		filedump: string,
 		path: string,
 		isTorrent: boolean
-	): injectData {
+	): InjectData {
 		return [
 			filename,
 			filedump,
@@ -333,14 +332,13 @@ export default class Deluge implements TorrentClient {
 				params
 			);
 
-			if (response && typeof response.result === "object") {
+			if (response.result.torrents) {
 				torrent = response.result.torrents?.[searchee.infoHash];
 			} else {
 				throw new Error(
-					"Client returned unexpected response (non-object)"
+					"Client returned unexpected response (object missing)"
 				);
 			}
-
 			if (torrent === undefined) {
 				throw new Error(
 					`Torrent not found in client (${searchee.infoHash})`
@@ -365,7 +363,10 @@ export default class Deluge implements TorrentClient {
 				message: `Failed to fetch torrent data: ${searchee.name} - (${searchee.infoHash})`,
 			});
 			logger.debug(e);
-			throw new Error("web.update_ui: failed to fetch data from client");
+			// @ts-expect-error needs es2022 target (tsconfig)
+			throw new Error("web.update_ui: failed to fetch data from client", {
+				cause: e,
+			});
 			//return { complete: false, save_path: "missing" };
 		}
 	}
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -226,6 +226,8 @@ export default class Deluge implements TorrentClient {
 					message: `Injection failed: ${addResult.error.message}`,
 				});
 				return InjectionResult.FAILURE;
+			} else {
+				return InjectionResult.FAILURE;
 			}
 		} catch (injectResult) {
 			logger.error({
@@ -235,7 +237,6 @@ export default class Deluge implements TorrentClient {
 			logger.debug(injectResult);
 			return InjectionResult.FAILURE;
 		}
-		return InjectionResult.FAILURE;
 	}
 
 	/**
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -25,9 +25,26 @@ enum DelugeErrorCode {
 interface TorrentInfo {
 	complete: boolean;
 	save_path: string;
+	label?: string | undefined;
+}
+interface InjectResponse {
+	result?: boolean;
+	error?: {
+		message?: string;
+	};
+}
+interface TorrentObject {
+	progress: number;
+	save_path: string;
+	state: string;
 	label?: string;
 }
-
+interface InfoResponse {
+	result?: { torrents?: Record<string, TorrentObject> };
+	error?: {
+		message?: string;
+	};
+}
 export default class Deluge implements TorrentClient {
 	private delugeCookie: string | null = null;
 	private delugeLabel = "cross-seed";
@@ -99,14 +116,15 @@ export default class Deluge implements TorrentClient {
 		if (this.delugeCookie) headers.set("Cookie", this.delugeCookie);
 
 		let response: Response, json: DelugeResponse;
+		const id = Math.floor(Math.random() * 0x7fffffff);
 		try {
 			response = await fetch(href, {
 				body: JSON.stringify({
-					...{ method: method, params: params },
-					id: Math.floor(Math.random() * 0x7fffffff),
+					...{ method, params },
+					id,
 				}),
 				method: "POST",
-				headers: headers,
+				headers,
 			});
 		} catch (networkError) {
 			// @ts-expect-error needs es2022 target (tsconfig)
@@ -197,10 +215,14 @@ export default class Deluge implements TorrentClient {
 			const params = this.formatData(
 				`${newTorrent.getFileSystemSafeName()}.cross-seed.torrent`,
 				newTorrent.encode().toString("base64"),
+				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 				path ? path : torrentInfo!.save_path,
 				!!searchee.infoHash
 			);
-			const addResult = await this.call("core.add_torrent_file", params);
+			const addResult = (await this.call(
+				"core.add_torrent_file",
+				params
+			)) as InjectResponse;
 
 			if (addResult?.result) {
 				const { dataCategory } = getRuntimeConfig();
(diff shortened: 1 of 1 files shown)


src/clients/Deluge.ts
```diff
@@ -91,7 +109,7 @@ export default class Deluge implements TorrentClient {
 			});
 		}
 		try {
-			json = await response.json();
+			json = await (response.json() as Promise<DelugeResponse>);
```
zakkarry, line 112:
```suggestion
			json = await response.json();
```

I'm not entirely sure why this is needed, we instantiate `json` as DelugeResponse, so adding an additional cast in setting it, especially as a Promise, doesn't seem necessary?

I might be wrong....@mmgoodnow said something similar to me in my PR for deluge.

https://github.com/cross-seed/cross-seed/pull/499#discussion_r1357766342

src/clients/Deluge.ts
```diff
@@ -116,9 +134,7 @@ export default class Deluge implements TorrentClient {
 	 * parses the set-cookie header and updates stored value
 	 */
 	private handleResponseHeaders(headers: Headers) {
-		if (headers.has("Set-Cookie")) {
-			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
-		}
+		this.delugeCookie = headers.get("Set-Cookie")?.split(";")[0] ?? null;
```
mmgoodnow, line 137: not sure I agree with this change. just because we don't receive a set cookie doesn't mean we should throw our cookie away.

src/clients/Deluge.ts
```diff
@@ -171,25 +188,31 @@ export default class Deluge implements TorrentClient {
 					return InjectionResult.TORRENT_NOT_COMPLETE;
 				}
 			}
-
 			const params = this.formatData(
 				`${newTorrent.name}.cross-seed.torrent`,
 				newTorrent.encode().toString("base64"),
-				path ? path : torrentInfo.save_path,
+				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+				path ? path : torrentInfo!.save_path,
```
mmgoodnow, line 195: what is the point of this change?

src/clients/Deluge.ts
```diff
@@ -25,9 +25,26 @@ enum DelugeErrorCode {
...
+	state: string;
 	label?: string;
 }
-
+interface InfoResponse {
+	result?: { torrents?: Record<string, TorrentObject> };
+	error?: {
+		message?: string;
+	};
+}
```
mmgoodnow, line 47: you can combine these with DelugeResponse if you make DelugeResponse generic
```ts
type InfoResponse: DelugeResponse<{ torrents : Record<string, TorrentObject }>, ErrorType>
```

`ErrorType` might always be the same - an object with `message` and `code`.

if you want to get really fancy, you can make a union:
```ts
type ResultType = { result: string, foo: string }
type ErrorType = { error: string, bar: string}
type DelugeResponse = ResultType | ErrorType;

const x: DelugeResponse = undefined as unknown as DelugeResponse;

if ("error" in x) {
    x.bar
} 
if ("result" in x) {
    x.foo
}

```

src/clients/Deluge.ts
```diff
@@ -201,6 +224,8 @@ export default class Deluge implements TorrentClient {
 					message: `Injection failed: ${addResult.error.message}`,
 				});
 				return InjectionResult.FAILURE;
+			} else {
+				return InjectionResult.FAILURE;
```
mmgoodnow, line 228: want to add some logging here?

src/clients/Deluge.ts
```diff
@@ -237,35 +262,35 @@ export default class Deluge implements TorrentClient {
...
 		try {
-			let torrentLabel: string = undefined;
 			const params = [
 				["state", "progress", "save_path", "label"],
 				{ hash: searchee.infoHash },
 			];
-			const response = await this.call("web.update_ui", params);
-			if (response?.result?.torrents?.[searchee.infoHash] === undefined) {
+			if (!searchee.infoHash)
+				throw new Error("Can't search a torrent without a infoHash");
```
mmgoodnow, line 270: nit: add curly braces

src/clients/Deluge.ts
```diff
@@ -237,35 +262,35 @@ export default class Deluge implements TorrentClient {
...
+
+			const response = (await this.call(
+				"web.update_ui",
+				params
+			)) as InfoResponse;
+			const torrent = response?.result?.torrents?.[searchee.infoHash];
```
mmgoodnow, line 276: good variable extraction here

mmgoodnow (commented):
Thanks for the type updates!

src/clients/Deluge.ts
zakkarry:
```suggestion
				return InjectionResult.FAILURE;
				logger.debug({
					label: Label.DELUGE,
					message: `Unknown injection failure: ${newTorrent.name}`,
				});
```

Something like this might work...

src/clients/Deluge.ts
ColinHebert: Good point, will update

src/clients/Deluge.ts
ColinHebert: `torrentInfo` is technically undefined at that point (only defined once earlier in an if condition); this forces the acknowledgement that torrentInfo will not be undefined here.

The real fix would be a bit longer than that as it requires to rejiggle the logic

src/clients/Deluge.ts
ColinHebert: yeah still not happy with the amount of nullskips, but at least it's contained.

src/clients/Deluge.ts
ColinHebert: Sure, need to add it to the checkstyle

src/clients/Deluge.ts
zakkarry: > yeah still not happy with the amount of nullskips, but at least it's contained.

I've almost eliminated them, I believe, in the commit I just made. Take a look.

src/clients/Deluge.ts
zakkarry: Something like this should be sufficient I think?

src/clients/Deluge.ts
zakkarry: Rejiggled a bit.

src/clients/Deluge.ts
```diff
@@ -22,11 +22,23 @@ enum DelugeErrorCode {
...
+	}
+];
+
+type DelugeJSON<DelugeResult, ErrorType> = {
+	result?: DelugeResult;
+	error?: ErrorType;
+};
+type ErrorType = { message?: string; code?: DelugeErrorCode };
+type DelugeResult = boolean | string | TorrentStatus;
+type TorrentStatus = { torrents?: Record<string, TorrentInfo> };
```
mmgoodnow, line 41: here's what I would do
```ts
// this one's only relevant for getTorrentInfo I think
type TorrentStatus = { torrents?: Record<string, TorrentInfo> };

// we don't know which one of these it's gonna be, it depends on which method we called
type DelugeResult = boolean | string | TorrentStatus;

// errors always have the same shape regardless of which method we called
type ErrorType = { message?: string; code?: DelugeErrorCode };

// DelugeJSON can represent any number of responses based on which method we call
// so it has to take a type argument ResultType which the caller will provide.
type DelugeJSON<ResultType> = {
	result?: ResultType;

// since errors always have the same shape, 
	error?: ErrorType;
};

...

// implementing call (ResultType is a type variable that we then feed into the same-named ResultType of DelugeJSON)
call<ResultType>(method: string, params: Whatever): Promise<DelugeJSON<ResultType>>

...

// calling call inside getTorrentInfo
this.call<TorrentStatus>("core.get_torrent_status", ...);

// calling call inside inject
this.call<boolean>("core.add_torrent", ...);
```

src/clients/Deluge.ts
```diff
@@ -22,11 +22,22 @@ enum DelugeErrorCode {
...
+		download_location: string;
+	}
+];
+
+type DelugeJSON<ResultType> = {
+	result?: ResultType;
+	error?: ErrorType;
+};
+type ErrorType = { message?: string; code?: DelugeErrorCode };
+type TorrentStatus = { torrents?: Record<string, TorrentInfo> };
```
mmgoodnow, line 40: it's technically not a problem but I prefer to define things before using them. In this case, ErrorType is used on line 37 before it's defined on line 39

src/clients/Deluge.ts
```diff
@@ -91,22 +114,27 @@ export default class Deluge implements TorrentClient {
 	/**
 	 * ensures authentication and sends JSON-RPC calls to deluge
 	 */
-	private async call(method: string, params: object, retries = 1) {
+	private async call<ResultType>(
+		method: string,
+		params: object,
```
mmgoodnow, line 119: we at least know params has to be an array, right? how about any[]

src/clients/Deluge.ts
```diff
@@ -22,11 +22,22 @@ enum DelugeErrorCode {
 	RPC_FAIL = 4,
 	BAD_JSON = 5,
 }
-interface TorrentInfo {
-	complete: boolean;
-	save_path: string;
-	label?: string;
-}
+type injectData = [
```
mmgoodnow, line 25: Types use `PascalCase`

```suggestion
type InjectData = [
```

src/clients/Deluge.ts
zakkarry: I think we might be able to get away with string[] with its current usage, but I'd have to double check every usage. It would potentially be tech debt if we ever expanded the capabilities though and

src/clients/Deluge.ts
```diff
@@ -22,11 +22,22 @@ enum DelugeErrorCode {
...
+	string,
+	string,
+	{
```
mmgoodnow, line 28:
```suggestion
	filename: string,
	filedump: string,
	options: {
```

src/clients/Deluge.ts
```diff
@@ -66,19 +81,27 @@ export default class Deluge implements TorrentClient {
...
 				"Deluge WebUI disconnected from daemon...attempting to reconnect."
 			);
-			const webuiHostList = await this.call("web.get_hosts", [], 0);
-			const connectResponse = await this.call(
+			const webuiHostList = await this.call<object>(
+				"web.get_hosts",
+				[],
+				0
+			);
+			const connectResponse = await this.call<undefined>(
```
mmgoodnow, line 99: can we create a higher-fidelity type than `object` for webuiHostList? Based on line 101 where we use it like `webuiHostList.result[0][0]` it seems like it's an array of arrays of something?

src/clients/Deluge.ts
zakkarry: since explicit any[] error's i'm going with your suggestion of unknown[]

src/clients/Deluge.ts
```diff
@@ -140,34 +168,38 @@ export default class Deluge implements TorrentClient {
...
-				await this.call("label.add", [label]);
-				await this.call("label.set_torrent", [infoHash, label]);
-			}
+		const setResult = this.isLabelEnabled
+			? await this.call<void>("label.set_torrent", [infoHash, label])
+			: undefined;
+		if (!setResult) {
+			throw new Error(
+				"label.set_torrent: Client returned empty response"
+			);
```
mmgoodnow, line 199: why is this needed? seems like it wasn't needed before

src/clients/Deluge.ts
```diff
@@ -140,34 +168,38 @@ export default class Deluge implements TorrentClient {
...
 	 * returns true if successful.
 	 */
 	private async labelEnabled() {
-		const enabledLabels = await this.call("core.get_enabled_plugins", []);
-		return enabledLabels?.result?.includes?.("Label");
+		const enabledPlugins = await this.call<string>(
+			"core.get_enabled_plugins",
+			[]
+		);
+		return enabledPlugins ? enabledPlugins.result.includes("Label") : false;
```
mmgoodnow, line 185: I think what you're really looking for here is this:
```suggestion
		return enabledPlugins.result ? enabledPlugins.result.includes("Label") : false;
```

AFAIK, the return value of `call` will always be an object, it's the result and error properties _on_ that object that may or not be there

src/clients/Deluge.ts
```diff
@@ -140,34 +168,38 @@ export default class Deluge implements TorrentClient {
 	 * parses the set-cookie header and updates stored value
 	 */
 	private handleResponseHeaders(headers: Headers) {
-		if (headers.has("Set-Cookie")) {
-			this.delugeCookie = headers.get("Set-Cookie").split(";")[0];
-		}
+		this.delugeCookie = headers.has("Set-Cookie")
+			? headers.get("Set-Cookie").split(";")[0]
+			: null;
```
mmgoodnow, line 173: I don't think this change is beneficial

src/clients/Deluge.ts
```diff
@@ -140,34 +168,38 @@ export default class Deluge implements TorrentClient {
...
-				await this.call("label.set_torrent", [infoHash, label]);
-			}
+		const setResult = this.isLabelEnabled
+			? await this.call<void>("label.set_torrent", [infoHash, label])
+			: undefined;
+		if (!setResult) {
+			throw new Error(
+				"label.set_torrent: Client returned empty response"
+			);
+		} else if (setResult.error?.code == DelugeErrorCode.RPC_FAIL) {
```
mmgoodnow, line 200: nit

```suggestion
		} else if (setResult.error?.code === DelugeErrorCode.RPC_FAIL) {
```

mmgoodnow, line 200: nit: ```suggestion
		} else if (setResult.error?.code === DelugeErrorCode.RPC_FAIL) {
```

File: src/clients/Deluge.ts (hunk 15 above)
Comment (zakkarry) [reply src/clients/Deluge.ts:185]: It will either throw or return JSON response as `Promise<DelugeJSON<ResultType>>` - there are no other possibilities as far as I remember. I can double-check.

File: src/clients/Deluge.ts [hunk 18]
```diff
@@ -300,7 +365,8 @@ export default class Deluge implements TorrentClient {
 				message: `Failed to fetch torrent data: ${searchee.name} - (${searchee.infoHash})`,
 			});
 			logger.debug(e);
-			return { complete: false, save_path: "missing" };
+			throw new Error("web.update_ui: failed to fetch data from client");
```
mmgoodnow, line 368:
```suggestion
			throw new Error("web.update_ui: failed to fetch data from client", { cause: e });
```

src/clients/Deluge.ts
zakkarry: My thought was if the header doesn't container the cookie, which it will every response as long as the cookie is valid, we should invalidate it.

I guess the one aspect that I didn't consider is transient errors mid-stream?

Either way, if we nullify the cookie header `call` will automatically reauthenticate and cookie should thus be reset and things should proceed normally.

src/clients/Deluge.ts
```diff
@@ -262,36 +317,46 @@ export default class Deluge implements TorrentClient {
...
 			];
-			const response = await this.call("web.update_ui", params);
-			if (response?.result?.torrents?.[searchee.infoHash] === undefined) {
+
+			const response = await this.call<TorrentStatus>(
+				"web.update_ui",
+				params
+			);
+
+			if (response && typeof response.result === "object") {
```
mmgoodnow, line 336:
```suggestion
			if (response.result) {
```

src/clients/Deluge.ts
zakkarry: As with the other usage of cause, this isn't supported until we push to es2022 and will require a directive to suppress. 

I can add it if you still want to.

src/clients/Deluge.ts
zakkarry: I think this makes the else unreachable

src/clients/Deluge.ts
mmgoodnow: yeah just throw in the directive. I'm gonna work on getting the tsconfig updated

src/clients/Deluge.ts
mmgoodnow: yeah, but the way that cookies work more generally is that they are expected to persist until they expire or the server sends another Set-Cookie header. 

If deluge changed their implementation on the server to send the Set-Cookie header for only the login endpoint, the UI would still work just fine because browsers cache cookies. But with this change, cross-seed would require a reauth for every call

zakkarry: Gotcha. Futureproofing.

src/clients/Deluge.ts
zakkarry: I think I was handling if the Label plugin stack-traced and crashed. Honestly don't remember the specific case, but there was some condition that had me do it :P

src/clients/Deluge.ts
mmgoodnow: should we just delete the `else` then?

regarding my suggestion, when would `result` be truthy but not `typeof result === "object"`? I think that would only happen if result were a number.

zakkarry: When the torrent is not present, you get the following 
```{
  result: {
    connected: true,
    torrents: {},
    filters: {
      state: [Array],
      tracker_host: [Array],
      owner: [Array],
      label: [Array]
    },```

Comment (zakkarry) [reply src/clients/Deluge.ts:336]: result would always seem to exist unless something catastrophic happened, perhaps we should throw and attempt to reconnect Deluge or at the very least reauth and validate?

Comment (mmgoodnow) [reply src/clients/Deluge.ts]: I think just assuming result exists is fine. Other errors probably wouldn't go through this codepath

File: src/clients/Deluge.ts [hunk 20]
```diff
@@ -182,29 +213,38 @@ export default class Deluge implements TorrentClient {
...
+				logger.debug({
+					label: Label.DELUGE,
+					message: `Injection failure: ${newTorrent.name} was missing critical data.`,
+				});
+				return InjectionResult.FAILURE;
+			}
 
 			const params = this.formatData(
-				`${newTorrent.getFileSystemSafeName()}.cross-seed.torrent`,
+				`${newTorrent.name}.cross-seed.torrent`,
```
mmgoodnow, line 232: what about `getFileSystemSafeName`?

src/clients/Deluge.ts
```diff
@@ -182,29 +213,38 @@ export default class Deluge implements TorrentClient {
...
+				"core.add_torrent_file",
+				params
+			);
+			if (!addResult) {
+				logger.debug({
+					label: Label.DELUGE,
+					message: `Injection failure: Client returned empty response.`,
+				});
+				return InjectionResult.FAILURE;
+			} else if (addResult.result) {
```
mmgoodnow, line 247: if we're worried about deluge returning completely empty json sometimes, maybe if we put this logic into `call`, then we would get this checkfor all calls?

Is this something we _should_ be worried about?

src/clients/Deluge.ts
```diff
@@ -300,7 +363,11 @@ export default class Deluge implements TorrentClient {
 				message: `Failed to fetch torrent data: ${searchee.name} - (${searchee.infoHash})`,
 			});
 			logger.debug(e);
-			return { complete: false, save_path: "missing" };
+			// @ts-expect-error needs es2022 target (tsconfig)
+			throw new Error("web.update_ui: failed to fetch data from client", {
+				cause: e,
+			});
+			//return { complete: false, save_path: "missing" };
```
mmgoodnow, line 370:
```suggestion
```

mmgoodnow (approved):
LGTM pending nits

src/clients/Deluge.ts
zakkarry: I think this PR was started before we did that, and rebasing results in it being removed as a result. My bad.

src/clients/Deluge.ts
zakkarry: I think we might already account for that and this might be deprecated code, call throws on a non-JSON response which I think covers this - and I don't see this anywhere in my logs for the last few weeks.

Seems unnecessary now I think.

src/clients/Deluge.ts
```diff
@@ -91,22 +115,27 @@ export default class Deluge implements TorrentClient {
...
 
-		let response: Response, json: DelugeResponse;
+		let response: Response, json: DelugeJSON<ResultType>;
+		const id = Math.floor(Math.random() * 0x7fffffff);
 		try {
 			response = await fetch(href, {
 				body: JSON.stringify({
-					...{ method: method, params: params },
-					id: Math.floor(Math.random() * 0x7fffffff),
+					...{ method, params },
```
mmgoodnow, line 134:
```suggestion
					method,
					params,
```

mmgoodnow (approved):

PR #550, opened 2023-11-15
add(sanitization/actions): deluge-labeling (injection) and torrent filename (injection/save)

zakkarry:
This PR adds some string sanitization to both the injection and saving of torrents where the name is used. This applies to some trackers that don't use normal file/folder structure and naming for their .torrents and then causes cross-seed to ENOENT on injection or incorrectly save.

Also addresses injecting into Deluge when the source is a torrent already in a '.cross-seed' label, will now check for the '.cross-seed' tag and not duplicate it to '.cross-seed.cross-seed'

### DIFF
--- a/src/parseTorrent.ts
+++ b/src/parseTorrent.ts
@@ -110,6 +110,9 @@ export class Metafile {
 	static decode(buf: Buffer) {
 		return new Metafile(bencode.decode(buf));
 	}
+	getFileSystemSafeName(): string {
+		return this.name.replace("/", "");
+	}
 
 	encode(): Buffer {
 		return bencode.encode(this.raw);
--- a/src/torrent.ts
+++ b/src/torrent.ts
@@ -11,11 +11,7 @@ import { Metafile } from "./parseTorrent.js";
 import { Result, resultOf, resultOfErr } from "./Result.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 import { createSearcheeFromTorrentFile, Searchee } from "./searchee.js";
-import {
-	reformatTitleForSearching,
-	sanitizeTorrentName,
-	stripExtension,
-} from "./utils.js";
+import { reformatTitleForSearching, stripExtension } from "./utils.js";
 
 export interface TorrentLocator {
 	infoHash?: string;
@@ -116,7 +112,7 @@ export function saveTorrentFile(
 	const { outputDir } = getRuntimeConfig();
 	const buf = meta.encode();
 	const name = stripExtension(meta.name);
-	const filename = `[${tag}][${tracker}]${sanitizeTorrentName(name)}.torrent`;
+	const filename = `[${tag}][${tracker}]${meta.getFileSystemSafeName()}.torrent`;
 	fs.writeFileSync(path.join(outputDir, filename), buf, { mode: 0o644 });
 }
 
--- a/src/torrent.ts
+++ b/src/torrent.ts
@@ -111,8 +111,9 @@ export function saveTorrentFile(
 ): void {
 	const { outputDir } = getRuntimeConfig();
 	const buf = meta.encode();
-	const name = stripExtension(meta.name);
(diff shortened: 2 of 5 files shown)


src/utils.ts
```diff
@@ -113,3 +113,8 @@ export function extractCredentialsFromUrl(
 		return resultOfErr("invalid URL");
 	}
 }
+
+export function sanitizeTorrentName(name: string): string {
+	const sanitizedName = name.replace("/", "");
+	return sanitizedName;
+}
```
mmgoodnow, line 120: I don't think filenames can actually contain slashes - where are these torrent names coming from and how is it possible they have slashes?

https://stackoverflow.com/questions/9847288/is-it-possible-to-use-in-a-filename

mmgoodnow (commented):
`.cross-seed` de-nesting stuff looks good, but I'm a little confused about the filenames

src/utils.ts
zakkarry: It's not the filename - it's the 'name' in the torrent file, which is what cross-seed is using to create the filename for the .torrent file.

I didn't think it was possible either, but I had the guy send me a screenshot of the torrent file in a torrent file editor, and it did have that in the name, was a season pack, and the files were shown in the root of the torrent, not in a folder.

When he loaded into his client, it went to the folder layout in qBittorrent though, I just applied same logic to Deluge as well.

mmgoodnow: could we put this as a method on Metafile? `getFileSystemSafeName()` or something like that?

mmgoodnow (approved):
LGTM pending the open thread. I can cut a release later tonight once you merge

src/utils.ts
zakkarry: Moved it to Metafile

src/parseTorrent.ts
```diff
@@ -110,6 +110,9 @@ export class Metafile {
 	static decode(buf: Buffer) {
 		return new Metafile(bencode.decode(buf));
 	}
+	static getFileSystemSafeName(name: string) {
```
mmgoodnow, line 113: I don't think this needs to be static, it can be an instance method—then we can return `this.name.replace(...)`

src/torrent.ts
```diff
@@ -112,7 +112,9 @@ export function saveTorrentFile(
 	const { outputDir } = getRuntimeConfig();
 	const buf = meta.encode();
 	const name = stripExtension(meta.name);
```
mmgoodnow, line 114: will this `stripExtension` bit still work?

zakkarry: You mean if the name has a / in it?

I don't think it will strip it....considering it wont endWith just the extension. But considering that if it ends with a / it wont have an extension, I don't think we would technically really need to worry about that? It's not a single file torrent.

Thoughts?

src/parseTorrent.ts
zakkarry: Updated.

src/torrent.ts
mmgoodnow: makes sense

src/torrent.ts
```diff
@@ -112,7 +112,7 @@ export function saveTorrentFile(
...
-	const filename = `[${tag}][${tracker}]${name}.torrent`;
+	const filename = `[${tag}][${tracker}]${meta.getFileSystemSafeName()}.torrent`;
```
mmgoodnow, line 115: looks like we lost the `stripExtension` bit:

```suggestion
	const filename = `[${tag}][${tracker}]${stripExtension(meta.getFileSystemSafeName())}.torrent`;
```

zakkarry: Oops, good eyes

zakkarry: Ah good catch, yea that would have been lost after implementing it into Metafile.

mmgoodnow (approved):

PR #552, opened 2023-11-17
fix(deluge/webui_unconnected): checks for webui being connected to daemon

zakkarry:
this addresses when the web_ui is not connected to a daemon.

it will get the available daemons and connect to it

the current implementation, if not connected, will return errors or empty json results.

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -56,11 +56,31 @@ export default class Deluge implements TorrentClient {
 				"You need to define a password in the delugeRpcUrl. (e.g. http://:<PASSWORD>@localhost:8112)"
 			);
 		}
-		const response = await this.call("auth.login", [password], 0);
-		if (!response.result) {
+		const authResponse = await this.call("auth.login", [password], 0);
+		if (!authResponse.result) {
 			throw new CrossSeedError(
 				`Reached Deluge, but failed to authenticate: ${href}`
 			);
+		}
+		const connectedResponse = await this.call("web.connected", [], 0);
+
+		if (!connectedResponse.result) {
+			logger.warn(
+				"Deluge WebUI disconnected from daemon...attempting to reconnect."
+			);
+			const webuiHostList = await this.call("web.get_hosts", [], 0);
+			const connectResponse = await this.call(
+				"web.connect",
+				[webuiHostList.result[0][0]],
+				0
+			);
+			if (connectResponse) {
+				logger.info("Deluge WebUI connected to the daemon.");
+			} else {
+				throw new CrossSeedError(
+					"Unable to connect WebUI to Deluge daemon. Connect to the WebUI to resolve this."
+				);
+			}
 		}
 	}
 


src/clients/Deluge.ts
```diff
@@ -56,12 +56,32 @@ export default class Deluge implements TorrentClient {
...
+		if (!connectedResponse.result) {
+			logger.warn(
+				"Deluge WebUI disconnected from daemon...attempting to reconnect."
+			);
+			const webuiHostList = await this.call("web.get_hosts", [], 0);
+			const connectResponse = await this.call(
+				"web.connect",
+				[webuiHostList.result[0][0]],
+				0
+			);
```
mmgoodnow, line 76: either of these calls could throw, I think—can we wrap in a try-catch?

mmgoodnow (commented):
I don't really understand what's going on here, have a few questions
- why do we not hit the deluge daemon directly?
- why is there another layer in the middle?
- Is the web API just more ergonomic?
- Why do we need to explicitly connect the web server to the daemon
  - does that not happen automatically?

zakkarry:
I'm going to try and addresses these from a few different angles, as best I can. I preface with the fact that we discussed the daemon RPC vs JSON-RPC through webui before I wrote the implementation, and I think this will just reiterate those in more specific context.

> I don't really understand what's going on here, have a few questions
> 
>     * why do we not hit the deluge daemon directly?

Basically, the daemon is not accessible remotely by default. From docker it would fail from cross-seed's container unless you specifically enabled it in the preferences. It also requires a username AND password to be specifically added to the 'auth' file - via the UI or manually editing the .conf in deluge's config directory - and used for this exclusively.

It is more intuitive for the user to use what they are already familiar with (one password they're already using and the WebUI is almost certainly going to be enabled by default anyway) than to require them to enable the daemon, configure a user, and password (which needs proper permission levels specified in the auth file as well), and then do the exact same thing via RPC calls in a more difficult to set up sense.

That was something I mentioned, and when I tested the performance of the JSON-RPC through WebUI's endpoint it was almost identical to RPC. The source for deluge shows it's essentially forwarding the calls, and formatting in JSON. That's it, a bit of piping and string manipulation.
 
>     * Why do we need to explicitly connect the web server to the daemon
There are two parts to Deluge in most cases in both the WebUI and thin client modes. The WebUI is basically an extension that, if not running Deluge natively with the plugin enabled, is an entirely separate process. When it's running in daemon mode, for instance, docker, it will need a separate process for the web. This is deluge-web binary, you can see this in the containers if you shell in and `ps aux` - because of this they are not the same memory (obviously), and when you are trying to access the daemon the webui will need to be connected to it somehow. It does this over the daemon's RPC server and translates the message as I outlined above.

It is not usually an issue, if the person uses WebUI they will almost certainly have it always connected, but on certain conditions - ungrateful shutdown for example - the webui will disconnect and the state will not persist across the initialization of the web instance.

Because of this, it will authenticate correctly in cross-seed and pass validation but fail to receive anything other than empty JSON or error's while still returning 'result'.

It's more or less an unlikely event, but in my Python integration of Deluge's JSON-RPC for my mover script for unRAID - it came up. I just brought the logic/change here.

>       
>       * does that not happen automatically?

That depends, as outlined above, there are several conditions that are not consistently reproducible. Prior to a bugfix in binhex's container, it would almost always shut down ungracefully. I had that fixed, so it's a non-issue for me. But if the daemon crashes or for whatever reason the state of the connection is not terminated properly, it could not reconnect upon launch.

It's just logical to check for this, it's just one call to web.connected.

mmgoodnow (approved):

PR #555, opened 2023-11-28
fix(client/qbit): improve FormData requests

fabricionaweb:
- sets min node version to v18.14.0, lowest 18 we can go to work with those apis
- improves FormData requests to qBittorrent by using Blob to send the files
- uses native fetch and formData

## Summary by CodeRabbit

- **Refactor**
  - Updated the app to be compatible with Node.js version 20.

- **Documentation**
  - Updated the README to reflect the new minimum required Node.js version.

- **Chores**
  - Removed unnecessary `node-fetch` imports to align with Node.js built-in features.


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       -
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 18.18.2
```
onedr0p, line 26:
```suggestion
          node-version: 18
```

I would suggest keeping this `18` (which is the latest 18 release) so it doesn't need to be managed unless it needs to be updated to a newer major version (e.g. 20)

package.json
```diff
@@ -14,7 +14,7 @@
 		"cross-seed": "dist/cmd.js"
 	},
 	"engines": {
-		"node": ">=16"
+		"node": ">=18.18.2"
```
onedr0p, line 17:
```suggestion
		"node": ">=18"
```

Same reasons as above, I would set this to `>=18`

src/cmd.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --no-warnings
```
mmgoodnow, line 1: smart, I didn't know this worked!

src/clients/QBittorrent.ts
```diff
@@ -117,9 +112,9 @@ export default class QBittorrent implements TorrentClient {
 			);
 		}
 
-		const cookieArray = response.headers.raw()["set-cookie"];
+		const cookieArray = response.headers.get("set-cookie");
```
mmgoodnow, line 115: small style nit - now that we're accessing a `Headers` object with `.get`, I think we can use the proper casing (previously everything was lowercased in order to support case insensitivity)

```suggestion
		const cookieArray = response.headers.get("Set-Cookie");
```

README.md
```diff
@@ -19,7 +19,7 @@ delivery into your client.
 
 ## Requirements
 
--   [Node 16+](https://nodejs.org/en/download)
+-   [Node >= 18.14.0](https://nodejs.org/en/download)
```
mmgoodnow, line 22:
```suggestion
-   [Node >= 20](https://nodejs.org/en/download)
```

.nvmrc
```diff
@@ -1 +1 @@
-v16
+v18.14.0
```
mmgoodnow, line 1:
```suggestion
v20
```

package.json
```diff
@@ -14,7 +14,7 @@
 		"cross-seed": "dist/cmd.js"
 	},
 	"engines": {
-		"node": ">=16"
+		"node": ">=18.14.0"
```
mmgoodnow, line 17:
```suggestion
		"node": ">=20"
```

mmgoodnow (commented):
LGTM pending these nits. I changed the base branch to `v6-unstable` which will act as a staging area for a batch of PRs with breaking changes

src/clients/QBittorrent.ts
mmgoodnow: ah good catch. after reading up it seems like native fetch in node will not expose Set-Cookie headers to us directly in order to preserve compatibility with browser `fetch`.

mmgoodnow (approved):

PR #561, opened 2023-12-10
fix(announce/size): ignores erroneous size being used ("0")

zakkarry:
addresses potential problems when size is not provided in announce, which in autobrr results in "0" being provided as a size.

result is an automatic size mismatch, when it could/should otherwise be discarded as erroneous

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -104,7 +104,7 @@ async function assessCandidateHelper(
 ): Promise<ResultAssessment> {
 	const { matchMode } = getRuntimeConfig();
 
-	if (size != null && !sizeDoesMatch(size, searchee)) {
+	if (size && !sizeDoesMatch(size, searchee)) {
 		return { decision: Decision.SIZE_MISMATCH };
 	}
 
--- a/src/server.ts
+++ b/src/server.ts
@@ -37,7 +37,11 @@ function parseData(data: string) {
 		if ("infoHash" in parsed) {
 			parsed.infoHash = parsed.infoHash.toLowerCase();
 		}
-		if ("size" in parsed && typeof parsed.size === "string") {
+		if (
+			"size" in parsed &&
+			typeof parsed.size === "string" &&
+			parsed.size !== "0"
+		) {
 			parsed.size = Number(parsed.size);
 		}
 	} catch (e) {
--- a/src/server.ts
+++ b/src/server.ts
@@ -37,11 +37,7 @@ function parseData(data: string) {
 		if ("infoHash" in parsed) {
 			parsed.infoHash = parsed.infoHash.toLowerCase();
 		}
-		if (
-			"size" in parsed &&
-			typeof parsed.size === "string" &&
-			parsed.size !== "0"
-		) {
+		if ("size" in parsed && typeof parsed.size === "string") {
 			parsed.size = Number(parsed.size);
 		}
 	} catch (e) {


mmgoodnow:
what's the downstream behavior here?  I don't understand what this change fixes—it seems like the change is just that we use the string `"0"` rather than the number `0`.

Are announces guaranteed to have size available? I think we only really have two options: 
- if size is not provided, we skip the size check and go straight to snatching the torrent and relying on the rest of the checks
- we reject all of them assuming that autobrr was misconfigured

zakkarry:
My assumption was that parsed.size was a string, and when a user has the `{{ .Size }}` variable is used in the payload in autobrr (I don't believe this is default, at least it wasn't when I set it up) - it provides the size in bytes. (see https://autobrr.com/filters/macros#available-variables )

If the size is not provided in the IRC announce, it apparently provides a 0....which based on it being a string, I just assumed should be checked as "0" - also because it uses Number() ...so if it is used in autobrr's payload and not provided, it will always be zero as far as I'm aware.

I've asked @zze0s to comment on this, but he didn't get back to me. I assumed it was pretty cut and dry though.

mmgoodnow:
I'm not sure how these statements line up with each other:
> If the size is not provided in the IRC announce, it apparently provides a 0

> "0" is obviously erroneous. It should never be zero.

if 0s are possible in IRC announces, then we should do our best to handle them.

zakkarry:
The case that provided this was as follows. Announce made to Autobrr over IRC

```[2023-12-09 21:26:41] AnnounceBot: New Torrent Announcement: <TV :: Episodes HD> Name:'Release.Name.Here-GROUP' uploaded by 'Anonymous' -```

This announce does not contain any size. The variable is not on autobrr for some trackers.  (See this YAML https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/torrentleech.yaml#L76-L82 for it's definition and variables)

Another example: PTP - https://github.com/autobrr/autobrr/blob/develop/internal/indexer/definitions/ptp.yaml#L113-L119

No size variable available, and no size is announced. 

When {{ .Size }} is included in the payload, it seems to give a 0. There is currently not an optional inclusion as far as I'm aware.

mmgoodnow:
Regardless, it looks like the diff in this PR just makes it so that we leave the string "0" alone when we previously would have converted it to a number, which I don't think really does anything.

Can you clarify at a high level what you think the behavior should be for both autobrr and cross-seed?

zakkarry:
Ok, so I think this addresses if it zero, ignoring size mismatch. I think this is appropriate since a torrent's size is never going to be zero.

As I explained, we should disregard initial `Decision.SIZE_MISMATCH` when given a size of zero, and proceed to potentially snatch the torrent if it matches further?

I'm not really sure what the best behavior would be, I'm open to other ideas if this isn't right.

mmgoodnow:
That feels like the "correct" behavior but I'm also not really sure we can do this safely. Effectively this means that for any torrent that doesn't report size on their announces, we snatch every torrent they announce

zakkarry:
That was something i considered, however what is the alternative if we're announcing without size in the payload to begin with?

zakkarry:
See here

https://autobrr.com/3rd-party-tools#cross-seed-filter

mmgoodnow:
oh you're totally right, this makes sense then 👍

mmgoodnow (approved):

zze0s:
Hey! Size is not included in the example because it will lead to extra grabs of .torrent files just to parse the size in the cases where it's is not announced. We do implement APIs to fetch size for some indexers however.

The `Size` variable will be either the correct value in bytes or 0 if included in the payload. There's also `SizeString` for human readable format with unit.

IMO the downloading and parsing of .torrent files to get size should be one of the last steps of the matching algo. 

Try match `title`, `year`, (`season/ep`), `resolution`, `codec`, `hdr` and most importantly, `group` before even consider to grab the .torrent to parse size if it's missing. This will minimize the amount of unnecessary grabs of .torrent's. Some indexers are very strict about that.

zakkarry:
I have a branch `parse` that we were working on with a few additional regex's that I'm going to try to get to implementing in the matching. Fuse isn't cutting it in my opinion, a lot of unnecessary grabs.

Seems like this addresses the cases for '0' being provided though then. Will work on getting the matching algo a little bit more refined, hopefully before the coming weekend is over, but who knows.

PR #563, opened 2023-12-11
feat(linking): new linking structure and torrent-backed linking

### LINKED ISSUE #400: Support torrent auto-removal while considering cross-seeds?
There are several applications/scripts that will auto remove torrents based on certain conditions.

For example: https://github.com/jerrymakesjelly/autoremove-torrents

This uses a config format like this to determine deletion candidates

```
qbittorrent:
  client: qbittorrent
  host: <qbit host>
  username: <user>
  password: <pass>
  strategies:
    delete_done_after_30_days:
      categories:
        - done
      remove: seeding_time > 2600000
  delete_data: true
```

Unfortunately none of these applications that I'm aware of take into account cross-seeding. For example I want to be able to do "delete torrents after 28 days of seeding time, as long as all associated cross-seeds have also seeded for 28 days".

I believe that this behaviour would fit well into the cross-seed application by expanding it's ability to deal with cross-seeded data. Since you already have the logic to detect cross seeds, you would just need to run that again on the torrent clients existing torrents (filtered for category/label), apply whatever logic to determine whether the torrents should be deleted, and then call the torrent deletion endpoint.

Just a suggestion, thanks for considering!

I've written a script to do this myself, very hardcoded to qbittorrent though.
```
import requests
import sys

# Script to...
# - find timestamp categories
# - find torrents in timestamp categories
# - set category of these torrents to cross-seed
# - delete timestamp categories
# Shell arguments (mandatory)
# 1. Host of the qbittorrent webui instance, for example http://localhost:8098
# 2. webui username
# 3. webui password
# This is only necessary because SeedCross inexplicably adds cross-seeding torrents
#    using a timestamp category
# Run via cronjob

# Check for correct argument count
if len(sys.argv) < 4:
    print('Provide qbittorent host, username, and password arguments')
    sys.exit()

qbitHost = sys.argv[1]
user = sys.argv[2]
password = sys.argv[3]

# Create request session that shares cookies
session = requests.Session()

# Login
loginUrl = qbitHost + '/api/v2/auth/login'
loginPayload = { 'username': user, 'password': password }
headers = { 'Referer': qbitHost }
login = session.post(loginUrl, data = loginPayload, headers = headers)

if not login.ok:
    print("Failed to login")
    sys.exit()


# Get torrents
torrentListRes = session.get(qbitHost + '/api/v2/torrents/info')
if not torrentListRes.ok:
    print("Failed to get torrent list")
    sys.exit()
torrents = torrentListRes.json()

# Group torrents by content_path
# Matching content_path means they are cross-seeding
torrentsByPath = {}
for torrent in torrents:
    contentPath = torrent['content_path']
    if contentPath not in torrentsByPath:
        torrentsByPath[contentPath] = []
    torrentsByPath[contentPath].append(torrent)

seedLimitSeconds = 24 * 60 * 60 * 28 # 28 days
seedHashesToDelete = set()
deleteCategories = ['done']
for torrentsWithMatchingPath in torrentsByPath.values():
    allInDeleteCategories = all(map(
        lambda torrent: torrent['category'] in deleteCategories,
        torrentsWithMatchingPath
    ))
    allSeededForSeedLimit = all(map(
        lambda torrent: torrent['seeding_time'] > seedLimitSeconds,
        torrentsWithMatchingPath
    ))
    if allInDeleteCategories and allSeededForSeedLimit:
        seedHashesToDelete.update(map(
            lambda torrent: torrent['hash'],
            torrentsWithMatchingPath
        ))

# Send delete request
if len(seedHashesToDelete) > 0:
    deleteTorrentsPayload = { 'hashes': '|'.join(seedHashesToDelete),
                            'deleteFiles': 'true' }
    deleteRes = session.post(qbitHost + '/api/v2/torrents/delete', data=deleteTorrentsPayload)
```

zakkarry:
adds functionality to link data-based searches to tracker-specific subdirectories inside of linkDir

fixes nested data not linking to single file torrents

adds torrent backed linking support

potentially close #400

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -42,30 +42,32 @@ export async function performAction(
 ): Promise<ActionResult> {
 	const { action, linkDir } = getRuntimeConfig();
 	const trackerLinkDir = linkDir ? join(linkDir, tracker) : undefined;
-	mkdirSync(trackerLinkDir, {
-		recursive: true,
-	});
+	if (trackerLinkDir) {
+		mkdirSync(trackerLinkDir, {
+			recursive: true,
+		});
 
-	if (searchee.path) {
-		if (decision == Decision.MATCH) {
-			await linkExact(searchee.path, trackerLinkDir);
-		} else if (decision == Decision.MATCH_SIZE_ONLY) {
-			// Size only matching is only supported for single file or
-			// single, nested file torrents.
-			const candidateParentDir = dirname(newMeta.files[0].path);
-			let correctedlinkDir = trackerLinkDir;
+		if (searchee.path) {
+			if (decision == Decision.MATCH) {
+				await linkExact(searchee.path, trackerLinkDir);
+			} else if (decision == Decision.MATCH_SIZE_ONLY) {
+				// Size only matching is only supported for single file or
+				// single, nested file torrents.
+				const candidateParentDir = dirname(newMeta.files[0].path);
+				let correctedlinkDir = trackerLinkDir;
 
-			// Candidate is single, nested file
-			if (candidateParentDir !== ".") {
-				mkdirSync(join(trackerLinkDir, candidateParentDir), {
-					recursive: true,
-				});
-				correctedlinkDir = join(trackerLinkDir, candidateParentDir);
+				// Candidate is single, nested file
+				if (candidateParentDir !== ".") {
+					mkdirSync(join(trackerLinkDir, candidateParentDir), {
+						recursive: true,
+					});
+					correctedlinkDir = join(trackerLinkDir, candidateParentDir);
+				}
+				linkFile(
+					searchee.path,
+					join(correctedlinkDir, newMeta.files[0].name)
+				);
 			}
-			linkFile(
-				searchee.path,
-				join(correctedlinkDir, newMeta.files[0].name)
-			);
 		}
 	}
 
--- a/src/action.ts
+++ b/src/action.ts
@@ -150,6 +150,10 @@ export async function performAction(
 			logInjectionResult(InjectionResult.FAILURE, tracker, newMeta.name);
 			saveTorrentFile(tracker, getTag(searchee.name), newMeta);
 			return InjectionResult.FAILURE;
+		} else {
+			logger.warn(
+				"You have inaccessible data required for linking torrents. Consider making the torrent structure available to cross-seed."
+			);
 		}
 	}
 
--- a/src/action.ts
+++ b/src/action.ts
@@ -45,11 +45,9 @@ export async function performAction(
 	if (linkDir) {
 		trackerLinkDir = join(linkDir, tracker);
 	}
-	if (!existsSync(trackerLinkDir)) {
 		mkdirSync(trackerLinkDir, {
 			recursive: true,
 		});
-	}
 	if (searchee.path) {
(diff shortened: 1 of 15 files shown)


zakkarry:
> I dont know if I like it, wont it change from the category destination (in case of qbit)?

switch to deluge :P

src/action.ts
```diff
@@ -41,22 +41,30 @@ export async function performAction(
...
-
+	let trackerLinkDir: string;
+	if (linkDir) {
+		trackerLinkDir = join(linkDir, tracker);
+	}
+	if (!existsSync(trackerLinkDir)) {
+		mkdirSync(trackerLinkDir, {
+			recursive: true,
+		});
+	}
```
mmgoodnow, line 52:
```suggestion
		mkdirSync(trackerLinkDir, {
			recursive: true,
		});
```

src/action.ts
```diff
@@ -41,22 +41,30 @@ export async function performAction(
 	tracker: string
 ): Promise<ActionResult> {
 	const { action, linkDir } = getRuntimeConfig();
-
+	let trackerLinkDir: string;
+	if (linkDir) {
+		trackerLinkDir = join(linkDir, tracker);
+	}
```
mmgoodnow, line 47: small nit, using a ternary here allows this to be marked const:

```suggestion
	const trackerLinkDir = linkDir ? join(linkDir, tracker) : undefined;
	}
```

mmgoodnow (commented):
one thing I'm thinking about is - where do these tracker "names" come from? can we be guaranteed they're stable/consistent? probably…

what would you think of using infohash? would that be too ugly? it feels nice to give each torrent its own "sandbox"

zakkarry:
> what would you think of using infohash? would that look too ugly in the clients? it feels nice to give each torrent its own "sandbox"

I believe tracker name comes from prowlarr/jackett, so it would be whatever the user has it named in their aggregator - it's returned in the XML and parsed in parseXMLResults or whatever the function is named. InfoHash is rather ugly and lengthy relative. I mean, most people probably don't go digging through their files like that, but I would like to keep it somewhat aesthetically pleasing and manageable.

mmgoodnow (dismissed):

zakkarry:
Potentially pending major version bump.

mmgoodnow:
@zakkarry would you mind putting the torrent-backed linking stuff in a separate PR?

zakkarry:
I'll do it later today, about to head out.

zakkarry:
Should I PR to this PR? or what, because otherwise it seems like things could get pretty complicated rebasing if one is merged before the other.

Not sure how to do this, and I sent you a DM about the possibility of changing deluge.ts and the deluge refactor being merged prior to the torrent backed linking, and the mess that would be.

Off to nephew's party though, so message me if you get on.

zakkarry:
Got this branch back to data backed only, and made another branch for torrent linking based on this one.

Will sort out the PR for it later.

mmgoodnow (dismissed):
I wouldn't really consider this a breaking change, but should we merge it into v6 instead of master just in case?

also, one thing worth mentioning - how is this gonna work for people that already have existing links?

mmgoodnow (approved):

Issue #565, opened 2023-12-12
Perform searches on indexers independently


ShanaryS:
I've had similar issues in #564 which was because the indexer is disabled. I'm still seeing repeated searches, some that search for 0 indexers.

If a search fails on an indexer, it blocks other indexers from performing a search. This can stuck in a loop where nothing gets search due to the failed searches eating up the searchLimit.

A searchLimitIndexer would be an elegant solution as each indexer is independent, one failing should not prevent the others from searching. Since searches are already decided on a per indexer basis, this would be a natural extension.

```
2023-12-12 10:18:59 [32minfo[39m: [search] Found 4182 torrents, 3966 suitable to search for matches
2023-12-12 10:18:59 [32minfo[39m: [search] Limited to 10 searches
2023-12-12 10:18:59 [32minfo[39m: [1/10] Searching for 100 Great Science Fiction Short Shorts - Asimov.epub
2023-12-12 10:18:59 [32minfo[39m: [torznab] Searching 1 indexers for 100 Great Science Fiction Short Shorts - Asimov.epub (filtered by timestamps)
2023-12-12 10:18:59 [36mverbose[39m: [torznab] http://qserver:9696/32/api?apikey=[REDACTED]&t=search&q=100+Great+Science+Fiction+Short+Shorts+-+Asimov+epub
2023-12-12 10:18:59 [36mverbose[39m: [torznab] Snoozing indexers 29 with RATE_LIMITED until 2023-12-12 10:23:59
2023-12-12 10:18:59 [33mwarn[39m: Failed to reach http://qserver:9696/32/api
2023-12-12 10:18:59 [34mdebug[39m: request failed with code: 429
2023-12-12 10:18:59 [32minfo[39m: [2/10] Searching for 100.Yen.Love.2014.1080p.BluRay.DDP5.1.x264-PTer
2023-12-12 10:18:59 [32minfo[39m: [torznab] Searching 0 indexers for 100.Yen.Love.2014.1080p.BluRay.DDP5.1.x264-PTer (filtered by timestamps)
2023-12-12 10:18:59 [32minfo[39m: [3/10] Searching for 10000.BC.2008.1080p.MAX.WEB-DL.DDP.5.1.H.265-PiRaTeS[TGx]
[...]

PR #566, opened 2023-12-12
feat(config/searchLimit): Perform searches on indexers independently

ShanaryS:
Closes: #565

Very simple change, I simply created a Map<indexerId, count> to keep track. Added to sample configs as well as an info log to at start of search if enabled, and verbose logs for search on each indexer. I've tested it and it works.

I have never written javascript and just wrote from my experience with other languages so please be careful.

```
2023-12-12 17:21:30 [32minfo[39m: [scheduler] starting job: search
2023-12-12 17:23:25 [32minfo[39m: [search] Found 4183 torrents, 3981 suitable to search for matches
2023-12-12 17:23:25 [32minfo[39m: [search] Limited to 250 searches
2023-12-12 17:23:25 [32minfo[39m: [search] Limited to 4 searches per indexer
.
.
.
2023-12-12 17:24:14 [36mverbose[39m: [torznab] [4/4] Indexer http://qserver:9696/34/api has reached search limit
2023-12-12 17:24:14 [36mverbose[39m: [torznab] [1/4] Indexer http://qserver:9696/27/api searches
2023-12-12 17:24:14 [32minfo[39m: [torznab] Searching 1 indexers for 65 (2023) br hdr10 10bit hevc-d3g (filtered by timestamps)
2023-12-12 17:24:14 [36mverbose[39m: [torznab] http://qserver:9696/27/api?apikey=[REDACTED]&t=search&q=65+2023
```

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -225,12 +225,6 @@ function createCommandWithSharedOptions(name, description) {
 			"The number of searches before stops",
 			parseInt,
 			fallback(fileConfig.searchLimit, 0)
-		)
-		.option(
-			"--search-limit-indexer <number>",
-			"The number of searches per indexer before stops",
-			parseInt,
-			fallback(fileConfig.searchLimitIndexer, 0)
 		);
 }
 
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -225,6 +225,12 @@ function createCommandWithSharedOptions(name, description) {
 			"The number of searches before stops",
 			parseInt,
 			fallback(fileConfig.searchLimit, 0)
+		)
+		.option(
+			"--search-limit-indexer <number>",
+			"The number of searches per indexer before stops",
+			parseInt,
+			fallback(fileConfig.searchLimitIndexer, 0)
 		);
 }
 
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -253,8 +253,15 @@ module.exports = {
 
 	/**
 	 * The number of searches to be done before it stops.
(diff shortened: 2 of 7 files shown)


zakkarry:
We're pretty cautious about adding config settings, so it might be better to default to this behavior than make it a setting if we do implement this, I'm not even sure what the benefit of not using this would be in a real use-case.

Do you have an argument for why this should be an option and not the default behavior?

ShanaryS:
I agree, there is no practical reason not to make it default. I updated the changes so now searchLimit now applies per indexer rather than overall.

ShanaryS:
Added a new feature explained here: https://github.com/cross-seed/cross-seed/issues/565#issuecomment-1951485386

zakkarry:
> Added a new feature explained here: [#565 (comment)](https://github.com/cross-seed/cross-seed/issues/565#issuecomment-1951485386)

That seems to be out of the scope of your original PR?

ShanaryS:
I can make a new issue and PR if necessary, but I added it on this branch as it needs the searchLimit changes. Also has the same motivation as the original so I thought it would be easier to just do it at all at once.

mmgoodnow:
Thanks for the contribution—I appreciate your initiative here and I'm glad you were able to find a solution that works for you. 

Ultimately, your use case seems to be pretty uncommon and I don't think it's worth the added complexity and maintenance burden for us to merge this PR as native functionality of cross-seed. If we hear about this issue more in the future, we'll revisit this decision.

Issue #549, opened 2023-11-13
Not following redirects


kingofzeal:
I've recently set up cross-seed, and while looking through the verbose logs I'm seeing a recurring issue with one of my trackers:

```2023-11-13 00:30:52 ^[[36mverbose^[[39m: [server] POST /api/announce
2023-11-13 00:30:52 ^[[36mverbose^[[39m: [server] Received announce: Family.Guy.S22E05.1080p.WEB.h264-BAE
2023-11-13 00:30:52 ^[[31merror^[[39m: error downloading torrent at https://www.milkie.cc/api/v1/torrents/AtDQQYfH5dlx/torrent?key=*******: 301 Moved Permanently
2023-11-13 00:30:52 ^[[34mdebug^[[39m: response: <html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>

2023-11-13 00:30:52 ^[[36mverbose^[[39m: [decide] Family.Guy.S22E05.720p.WEB.h264-BAE - no match for milkie torrent Family.Guy.S22E05.1080p.WEB.h264-BAE - the torrent file failed to download
```

I'm using autobrr to pass IRC announces into cross-seed, and for other trackers there's no issues. The redirect in this case routes `https://www.milkie.cc` to `https://milkie.cc`. I believe I've found where this URL is coming from to address the root issue, but I also think cross-seed should try to follow redirects if possible.

zakkarry:
I brought this up like 2 weeks ago when it was brought to my attention in the discord....the only issue I thought of with enabling this was how it could potentially be used maliciously....albeit a very marginal chance.

I'll see what I can do to get this fixed soon.

kingofzeal:
Yeah, I can understand the security consideration, but I don't see it at a very high risk - if you're in a position to leverage it, there are so many other things you can do that would be far easier. 

I could also see it being a configuration whitelist, to allow opt-in to the behavior. I'm not sure what you're using under the hood to make these requests, so that could be far more complicated.

And like I said, I _think_ I've addressed the root issue (autobrr had the incorrect root URL), but I still need to verify. 

In any case, I appreciate you looking at it!

PR #567, opened 2023-12-13
follow 3xx redirects on snatch

zakkarry:
previously we were explicitly _not_ following redirects so that we could intercept redirects to magnet links, but that was a workaround because the errors thrown by redirecting to magnet links were uncatchable. Since we've switched to fetch, that's no longer a problem and we can now catch the errors caused by redirecting to magnet links, and let fetch follow the redirects otherwise. 

#549

### DIFF
--- a/src/torrent.ts
+++ b/src/torrent.ts
@@ -34,6 +34,16 @@ export async function parseTorrentFromFilename(
 	return Metafile.decode(data);
 }
 
+function isMagnetRedirectError(error: Error): boolean {
+	return (
+		// node-fetch
+		error.message.includes('URL scheme "magnet" is not supported.') ||
+		// undici
+		// @ts-expect-error error causes "not supported yet"
+		error?.cause.message.includes("URL scheme must be a HTTP(S) scheme")
+	);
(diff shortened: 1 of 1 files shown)


mmgoodnow:
what do they do instead?

mmgoodnow:
it seems like 301 and 302s will hit the `try` on line 95. but these responses won't have the data yet, they are an http redirect.

In the call we are making to `fetch` we are setting `redirect: "manual"` so that we can intercept redirects to magnet links (the previous library we used for http requests threw an error on magnet redirects in an uncatchable way, so is just tech debt).

Instead we might be able to let `fetch` handle redirects for us, and it will probably throw an error that we _CAN_ catch for magnet redirects. In which case, other redirects should Just Work

zakkarry:
I think this is pretty edge-case, but when Prowlarr/Jackett used www for base URL and HUNO did redirect for that (before it was changed to non-www) or autobrr forwarded the same www - it caused issues. There was only one other instance with milkie I am aware of, but it could be a problem in the future as well.

PR #574, opened 2023-12-29
Update Typescript config

mmgoodnow:
This PR updates our TypeScript configuration to allow ES2022 features, and turns on `strictNullChecks`. 

A few of our existing patterns don't really play nice with `strictNullChecks` and I used the non-null assertion operator `!` pretty liberally to fix the issues. 

### `ResultAssessment`
The optional property `metafile` is implicitly tied to `decision` having a value of `MATCH` or `MATCH_SIZE_ONLY`. We should make this explicit at some point by making it a union type.
```ts
export interface ResultAssessment {
	decision: Decision;
	metafile?: Metafile;
}
```
### `Searchee`
This has the same problem. The optional property `infoHash` is implicitly tied to `path` not existing. We could make this explicit at some point, though we also might want to support both existing. 
```ts
export interface Searchee {
	infoHash?: string; // if searchee is torrent based
	path?: string; // if searchee is data based
	files: File[];
	name: string;
	length: number;
}
```

### DIFF
--- a/src/torrent.ts
+++ b/src/torrent.ts
@@ -39,7 +39,6 @@ function isMagnetRedirectError(error: Error): boolean {
 		// node-fetch
 		error.message.includes('URL scheme "magnet" is not supported.') ||
 		// undici
-		// @ts-expect-error error causes "not supported yet"
 		error?.cause.message.includes("URL scheme must be a HTTP(S) scheme")
 	);
 }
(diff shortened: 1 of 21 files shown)


src/torrent.ts
```diff
@@ -39,7 +39,6 @@ function isMagnetRedirectError(error: Error): boolean {
 		// node-fetch
 		error.message.includes('URL scheme "magnet" is not supported.') ||
 		// undici
-		// @ts-expect-error error causes "not supported yet"
 		error?.cause.message.includes("URL scheme must be a HTTP(S) scheme")
```
mmgoodnow, line 42:
```suggestion
		error.cause?.message.includes("URL scheme must be a HTTP(S) scheme")
```

Issue #570, opened 2023-12-16
Option to add torrents with different hashes paused


GordonFreemanK:
Hi!
Great tool, thanks! I have a slight issue with how it queues torrents in `inject` mode though, at least on Transmission.

If a torrent has the same file paths than another one but the hashes differ, on adding the new torrent, cross-seed should:
- trigger a hash verification (already does that, which is great).
- but add them paused.

The reason is that if the hashes genuinely differ on two different trackers, then adding the new one unpaused will make the client always overwrite the data for the old one as soon as verification completes (provided it has seeds for it on the new tracker). This in turn will make the data invalid for the pre-existing tracker...

Of course it's not great that different trackers have different hashes for the same torrent, but the choice of which tracker we should keep the data for should fall on the user - so in this case the torrents should be added paused.

My setup:
- Transmission 4.0.4
- All hosted on docker
- config:

```
module.exports = {
    delay: 10,
    torznab: [
        "http://xxx:9696/yyy/api?apikey=zzz",
        // Some other ones here
    ],
    dataDirs: [],
    matchMode: "safe",
    dataCategory: undefined,
    linkDir: "/downloads",
    linkType: "hardlink",
    skipRecheck: false,
    maxDataDepth: 2,
    torrentDir: "/torrents",
    outputDir: "/cross-seeds",
    includeEpisodes: true,
    includeSingleEpisodes: false,
    includeNonVideos: true,
    fuzzySizeThreshold: 0.02,
    excludeOlder: undefined,
    excludeRecentSearch: "1y",
    action: "inject",
    rtorrentRpcUrl: undefined,
    qbittorrentUrl: undefined,
    transmissionRpcUrl: "http://aaa:bbb@ccc:9091/transmission/rpc",
    delugeRpcUrl: undefined,
    duplicateCategories: false,
    notificationWebhookUrl: undefined,
    port: 2468,
    host: undefined,
    apiAuth: true,
    rssCadence: "30min",
    searchCadence: "1 day",
    snatchTimeout: undefined,
    searchTimeout: undefined,
    searchLimit: 1000,
};
```

The workaround is to use the `save` action mode while making sure `start_added_torrents` is false in the Transmission config.

In terms of implementation I assume it would be as simple as using the `paused` property of the [torrent-add](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#34-adding-a-torrent) RPC call to Transmission

zakkarry:
[...]

PR #585, opened 2024-01-23
feat(client/transmission):  only start added torrents if they are complete

GordonFreemanK:
This PR is created on the back of [this conversation](https://github.com/cross-seed/cross-seed/issues/570). It is an implementation of what is called _Solution 3_ there.

### Background

When [adding a torrent](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#34-adding-a-torrent) in Transmission (`torrent-add`), an optional field `paused` can be specified, to the effect that:
- if `false`, the torrent is started when added
- if `true`, the torrent is added paused
- if not specifed, the `start-added-torrents` property of the [Transmission configuration](https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md#files-and-locations) is used to determine if the torrent should be started when added

Also when adding a torrent, Transmission makes a decision regarding verification based on the `torrent-added-verify-mode` property of the [Transmission configuration](https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md#misc):
- if `full`, the torrent data will always be verified
- if `fast`, the torrent data verification will be skipped under a [certain set of conditions](https://github.com/transmission/transmission/pull/2626).

_Note:_ This behaviour cannot be controlled in the `torrent-add` request body.

### Bug

The way Transmission works, when all of the following conditions are met:
- a torrent is added for the purpose of cross-seeding (i.e. the torrent data already exists)
- the torrent is started when added (i.e. `paused` is set to `false` in the `torrent-add` request **or** `paused` is not set in the `torrent-add` request **and**  `start-added-torrents` is set to `true` in the Transmission configuration)
- the torrent verification is triggered (i.e. `torrent-added-verify-mode` is set to `full` in the Transmission configuration **or** Transmission decides to trigger it based on the criteria described in the [above linked PR](https://github.com/transmission/transmission/pull/2626)
- the torrent verification fails (i.e. the torrent data was corrupted, i.e. the torrents are actually different between two different trackers)

Then Transmission will start the torrent and overwrite whichever data pre-existed in the torrent files.
There is no way to tell Transmission, when adding a torrent, that **it should be started if and only if torrent verification is skipped or succeeds**.

### Solution

cross-seed aims to minimize user interaction as much as possible by starting torrents when added, but in this case a user interaction will always be required: when two different trackers have torrents with the same structure but where the data is different, the user must decide which of the torrents they would like to keep.

### Implementation

This PR implements a strategy to start torrents only when safe. Instead of a single request to Transmission to add a torrent with `paused` set to `false`, we instead perform three successive requests:
- add the torrent paused (`paused` set to `true`)
- retrieve the torrent status and completion percentage
- if the torrent is "safe" (i.e., **only if the torrent status is stopped and the torrent is complete**), then start the torrent

### Limitations

- With this behaviour in place, it would be recommended that the `torrent-added-verify-mode` property of the Transmission configuration be set to `fast` (which is the default). If set to `full`, then in practice all torrents will always be added paused (since there's no way the verification will have completed by the time the second call described above retrieves the torrent we just added).

### DIFF
--- a/src/clients/Transmission.ts
+++ b/src/clients/Transmission.ts
@@ -44,8 +44,6 @@ type TorrentDuplicateResponse = { "torrent-duplicate": TorrentMetadata };
 type TorrentAddedResponse = { "torrent-added": TorrentMetadata };
 type TorrentAddResponse = TorrentAddedResponse | TorrentDuplicateResponse;
 
-type TorrentStartResponse = {};
-
 function doesAlreadyExist(
 	args: TorrentAddResponse
 ): args is TorrentDuplicateResponse {
@@ -54,7 +52,6 @@ function doesAlreadyExist(
 
 export default class Transmission implements TorrentClient {
 	xTransmissionSessionId: string;
-
 	private async request<T>(
 		method: Method,
 		args: unknown = {},
@@ -162,7 +159,7 @@ export default class Transmission implements TorrentClient {
 		const [{ downloadDir, percentDone }] = queryResponse.torrents;
 
 		if (downloadDir == undefined || percentDone === undefined) {
-			return resultOfErr(InjectionResult.FAILURE)
+			return resultOfErr(InjectionResult.FAILURE);
 		}
 		if (percentDone < 1) {
 			return resultOfErr(InjectionResult.TORRENT_NOT_COMPLETE);
@@ -232,7 +229,7 @@ export default class Transmission implements TorrentClient {
 
 			const [{ percentDone, status }] = getResponse.torrents;
 			if (status === TorrentStatus.Stopped && percentDone === 1) {
-				await this.request<TorrentStartResponse>("torrent-start", {
+				await this.request<unknown>("torrent-start", {
 					ids: [torrentId],
 				});
 				return InjectionResult.SUCCESS;
--- a/src/clients/Transmission.ts
+++ b/src/clients/Transmission.ts
@@ -9,7 +9,7 @@ import { extractCredentialsFromUrl } from "../utils.js";
 import { TorrentClient } from "./TorrentClient.js";
 
 const XTransmissionSessionId = "X-Transmission-Session-Id";
-type Method = "session-get" | "torrent-add" | "torrent-get";
+type Method = "session-get" | "torrent-add" | "torrent-get" | "torrent-start";
 
 interface Response<T> {
 	result: "success" | string;
@@ -17,7 +17,11 @@ interface Response<T> {
 }
 
 interface TorrentGetResponseArgs {
-	torrents: { downloadDir: string; percentDone: number }[];
+	torrents: {
+		downloadDir?: string;
+		percentDone?: number;
+		status?: TorrentStatus;
+	}[];
 }
 
 interface TorrentMetadata {
@@ -26,9 +30,21 @@ interface TorrentMetadata {
 	name: string;
 }
 
+enum TorrentStatus {
+	Stopped = 0,
+	VerificationQueued,
+	Verifying,
+	DownloadQueued,
+	Downloading,
+	SeedingQueued,
+	Seeding,
+}
+
 type TorrentDuplicateResponse = { "torrent-duplicate": TorrentMetadata };
 type TorrentAddedResponse = { "torrent-added": TorrentMetadata };
 type TorrentAddResponse = TorrentAddedResponse | TorrentDuplicateResponse;
+
+type TorrentStartResponse = {};
 
 function doesAlreadyExist(
 	args: TorrentAddResponse
@@ -184,10 +200,8 @@ export default class Transmission implements TorrentClient {
 			}
 		}
 
-		let addResponse: TorrentAddResponse;
-
 		try {
-			addResponse = await this.request<TorrentAddResponse>(
+			const addResponse = await this.request<TorrentAddResponse>(
 				"torrent-add",
 				{
 					"download-dir": downloadDir,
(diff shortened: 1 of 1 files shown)


GordonFreemanK:
The PR is complete to the best of my knowledge but I left it as a draft because I haven't tested the code. Let me know if there's any tip you can give me to do so.

mmgoodnow:
https://github.com/cross-seed/cross-seed/blob/master/CONTRIRUTING.md

src/clients/Transmission.ts
```diff
@@ -27,10 +31,22 @@ interface TorrentMetadata {
...
 }
 
+enum TorrentStatus {
+	Stopped = 0,
+	VerificationQueued,
+	Verifying,
+	DownloadQueued,
+	Downloading,
+	SeedingQueued,
+	Seeding,
```
mmgoodnow, line 41: can you add a link in a comment so we can verify these enum values are in the correct order?

mmgoodnow (commented):
This seems like a reasonable approach. One thing that _might_ be a cause for concern is that the operation is no longer atomic - if injection fails, we might leave it in a broken halfway state. However maybe in practice this doesn't really happen enough to care (I don't use Transmission so you're probably the best equipped to answer that question).

mmgoodnow:
@GordonFreemanK are you still trying to merge this PR?

zakkarry:
> @GordonFreemanK are you still trying to merge this PR?

If so, needs a rebase and to target `v6-unstable`....plus whatever comments you made being addressed.

src/clients/Transmission.ts
GordonFreemanK: The torrent statuses are described in this section of the RPC spec, a bit further in the paragraph (search for `A number between 0 and 6, where`):
https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#33-torrent-accessor-torrent-get

src/clients/Transmission.ts
```diff
@@ -170,26 +186,44 @@ export default class Transmission implements TorrentClient {
...
+
+			const torrentId = addResponse["torrent-added"].id;
+
+			const getResponse = await this.request<TorrentGetResponseArgs>(
+				"torrent-get",
+				{
+					ids: [torrentId],
+					fields: ["percentDone", "status"],
+				}
+			);
```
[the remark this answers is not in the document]
GordonFreemanK: The whole torrent recheck is definitely not synchronous, as the app just kicks it off if it sees an issue with the file, but doesn't wait for the full re-check to complete (obviously, or the API call would just be hanging):
https://github.com/transmission/transmission/blob/fd4d097d13ab4236550fa01d411fd63fc161f38b/libtransmission/torrent.cc#L1002

If you're asking if:
- checking the logic to see if a re-seed is complete (i.e. essentially if all files are there and the first chunk passes the checksum test)
- if yes, set the torrent as complete, and if not, trigger a reckeck
- change the status of the torrent so that torrent-get will return the correct status

Then I don't know, honestly. I have never visually seen it take any time at all, as far as I can tell it is absolutely instant. But that's all I have, as the Transmission codebase is large.

zakkarry:
@GordonFreemanK i see you force pushed, did you rebase against `v6-unstable` ?

If so you also need to change the branch this PR is merging to, shouldn't be master but also `v6-unstable`

GordonFreemanK:
Hi I'm sorry I got into trouble trying to rebase, I think I messed up
yesterday evening, did I do anything that needs to be reverted?

zakkarry:
It doesn't look like you actually rebased anything by the diff, I'll try and rebase for you gimme a minute.

zakkarry:
There's a few type issues in transmission after rebasing due to changes we made in v6.

Your IDE should tell you, I think there are 3.

I've rebased for you and targeted v6-unstable.

GordonFreemanK:
> There's a few type issues in transmission after rebasing due to changes we made in v6.
> 
> Your IDE should tell you, I think there are 3.
> 
> I've rebased for you and targeted v6-unstable.

Thanks a lot! I got stumped by this, my knowledge of how git forks work needs improvements
I've fixed an issue with undefined-checking in the response of torrent-get, if that's what you meant. I get a successful build&lint locally now.

zakkarry:
```/home/runner/work/cross-seed/cross-seed/src/clients/Transmission.ts
  47:29  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead  @typescript-eslint/ban-types

✖ 1 problem (1 error, 0 warnings)```

GordonFreemanK:
> ```
> /home/runner/work/cross-seed/cross-seed/src/clients/Transmission.ts
>   47:29  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
> - If you want a type meaning "any object", you probably want `object` instead.
> - If you want a type meaning "any value", you probably want `unknown` instead.
> - If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
> - If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead  @typescript-eslint/ban-types
> 
> ✖ 1 problem (1 error, 0 warnings)```
> ```
> 
> lint problems.

Was this on latest?
I fixed it in https://github.com/cross-seed/cross-seed/pull/585/commits/fc422dcd24794f190e0495bb288a0a92c345b271

zakkarry:
Sorry no, restarted the workflow. It doesn't trigger when you push.

zakkarry:
rebased for you.

zakkarry:
you're all rebased and current, i also squashed the two consecutive lint failures.

will try and figure out the deployment issue, and then you can test it i guess

zakkarry:
If this isn't dead @GordonFreemanK - please reopen a new PR and target master (rebasing shouldn't have any conflicts since I just rebased prior.

Reference this PR # in the body of the PR as well please.

PR #587, opened 2024-01-29
feat(blocklist): excludes releases for cross-seeding based on config option

### LINKED ISSUE #379: Irreconcilable false positive - blocklist feature?
Hey,
So after our chat on discord I thought I'd just add a feature request for a simple way to block certain torrents from being added or even considered as possibilities to cross-seed.
This came about after achieving the first "irreconcilable false positive", as you put it.
Here is the log for the diff:
```
$ cross-seed diff 601ccf003689a3ec51844b2b998f02c6ec168bd7.torrent bde99189a563179158baf05e9fcc3ce8de4a1740.torrent
Configuration file already exists.
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

  {
    files: [
      {
+       length: 2613,
-       length: 2554,
        name: 'Flags.of.Our.Fathers.2006.BluRay.1080p.DTS-HD.MA.5.1.AVC.HYBRID.REMUX-FraMeSToR.NFO',
        path: 'Flags.of.Our.Fathers.2006.BluRay.1080p.DTS-HD.MA.5.1.AVC.HYBRID.REMUX-FraMeSToR/Flags.of.Our.Fathers.2006.BluRay.1080p.DTS-HD.MA.5.1.AVC.HYBRID.REMUX-FraMeSToR.NFO'
...
      }
    ],
+   length: 36422340563,
-   length: 36422340504,
    name: 'Flags.of.Our.Fathers.2006.BluRay.1080p.DTS-HD.MA.5.1.AVC.HYBRID.REMUX-FraMeSToR'
  }
    at diff (file:///usr/src/cross-seed/dist/diff.js:6:9)
    at Command.diffCmd (file:///usr/src/cross-seed/dist/diff.js:21:5)
    at async Command.parseAsync (/usr/src/cross-seed/node_modules/commander/lib/command.js:923:5)
    at async file:///usr/src/cross-seed/dist/cmd.js:182:1 {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: [Object],
  expected: [Object],
  operator: 'deepStrictEqual'
}
```
I know you gave me a few solutions to bypass this, but honestly cross-seed has been working so flawlessly that I'm tempted to just leave everything as it is and hope that you can get to this eventually.

Not sure how time consuming or difficult this would be, if I may suggest a simple way of keeping an updated blocklist:
Having a "blocklist" file that we can manually edit and just add hashes into it, but not sure if that would block it at the tracker level from even being searched. Honestly, I'm sure you'll come up with a much more elegant solution.

Thanks for your time and the great program that you've created :-)

zakkarry:
This PR adds the config option `blockList` which is an array of strings to be checked for in `filterByContent` which will exclude the releases from cross-seeding.

- [x] Zod configSchema added
- [x] transform undefined/null to empty array
- [x] validate empty array does not erroneously match or error
- [x] validate keywords are blocked and logged properly
- [x] add to decision to prevent webhook searching
- [x] default datadir filtering disregarding useless data searches

closes #379


src/cmd.ts
```diff
@@ -245,6 +245,12 @@ function createCommandWithSharedOptions(name, description) {
 			"The number of searches before stops",
 			parseInt,
 			fallback(fileConfig.searchLimit, 0)
+		)
+		.option(
+			"--block-list <strings>",
+			"The strings you wish to block",
```
mmgoodnow, line 251: can we be clearer about what gets matched against these strings? i.e. torrent/release names?

mmgoodnow (approved):
looks good. I wonder if this should be extended to info hashes as well? so a blocklist could look like `["Movie.2015.720p.WEB-DL.x264-ABCDE", "1234thd1n2t3hd4nth1d"]` and have it Just Work?

src/cmd.ts
zakkarry: We could use a simple (enough) regex to match for infohash and separate the searching so we don't run through needlessly.

Yes i can improve the messaging.

src/dataFiles.ts
```diff
@@ -22,13 +36,8 @@ export function findPotentialNestedRoots(
...
 		);
 		return [root, ...allDescendants];
-	} else if (
-		DATA_EXTENSIONS.includes(extname(root)) &&
-		!root.includes("sample")
-	) {
-		return [root];
 	} else {
-		return [];
+		return [root];
```
mmgoodnow, line 40: returning `[root]` rather than `[]` in the else case here seems like it might change more behavior than we want it to. Specifically I'm not sure `depth` works any more.

 What we can probably do to fix that is to move the `depth` check into the `checkIgnorePath` if statement - 
```js
if (depth <= 0 || checkIgnorePath()) { 
  return []
}
```

Then, we know below that statement we're just left with the standard dir case and the standard non-dir case.

src/dataFiles.ts
```diff
@@ -1,16 +1,30 @@
...
-import { DATA_EXTENSIONS } from "./constants.js";
+import { basename, extname, join } from "path";
+import {
+	BADFOLDERS_REGEX,
+	IGNORED_FOLDERS,
+	VIDEO_EXTENSIONS,
+} from "./constants.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 
+function checkIgnorePath(root: string, isDir: boolean) {
```
mmgoodnow, line 10: can we find a name for `checkIgnorePath` that makes it clearer what it does without having to look at its implementation? here's my best idea but open to others 

```suggestion
function shouldIgnorePathHeuristically(path: string, isDir: boolean) {
```

src/dataFiles.ts
```diff
@@ -1,16 +1,30 @@
...
+	return (
+		(isDir &&
+			(IGNORED_FOLDERS.includes(basename(root).toLowerCase()) ||
+				BADFOLDERS_REGEX.test(basename(root)))) ||
+		(!isDir && !VIDEO_EXTENSIONS.includes(extname(basename(root))))
```
mmgoodnow, line 15: can we refactor this `isDir && foo || !isDir && bar` into an `if` for readability? 

```js
const name = basename(root);
if (isDir) {
  return BADFOLDERS_REGEX.test(name) || IGNORED_FOLDERS.includes(name.toLowerCase())
} else {
  return !VIDEO_EXTENSIONS.includes(extname(basename(root))
}
```

src/constants.ts
```diff
@@ -60,3 +61,12 @@ export enum LinkType {
 	SYMLINK = "symlink",
 	HARDLINK = "hardlink",
 }
+
+export const IGNORED_FOLDERS = [
```
mmgoodnow, line 65: what's the difference between a "bad folder" and an "ignored folder"?

src/preFilter.ts
```diff
@@ -53,6 +62,20 @@ export function filterByContent(searchee: Searchee): boolean {
 	return true;
 }
 
+export function releaseInBlockList(
+	searchee: Searchee,
+	blockList: string[]
+): boolean {
+	return blockList.some((str) => {
```
mmgoodnow, line 69:
```suggestion
	return blockList.some((blockedStr) => {
```

src/preFilter.ts
```diff
@@ -53,6 +62,20 @@ export function filterByContent(searchee: Searchee): boolean {
...
+		return (
+			searchee.name.includes(str) ||
+			(str.length === 40 &&
+				searchee.infoHash &&
+				str === searchee.infoHash)
```
mmgoodnow, line 74: we can probably achieve the same thing in less comparisons with just `str === searchee.infoHash`
```suggestion
				str === searchee.infoHash
```

src/constants.ts
```diff
@@ -15,9 +15,9 @@ export const MOVIE_REGEX =
 export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?<group>[\w ]+)(?:\))?(?=(?:\.\w{1,5})?$)/i;
 
-export const VIDEO_EXTENSIONS = [".mkv", ".mp4", ".avi"];
+export const VIDEO_EXTENSIONS = [".mkv", ".mp4", ".avi", ".ts"];
```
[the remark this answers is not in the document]
zakkarry: They are transport stream captures, usually from a broadcast format, and associated with raw TV captures. Just tried to be thorough.

src/dataFiles.ts
```diff
@@ -1,16 +1,30 @@
...
+	);
```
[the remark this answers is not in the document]
zakkarry: absolutely.

mmgoodnow (approved):

PR #590, opened 2024-02-01
add(logging/pushnotifier): adds status code and error to logging

zakkarry:
This PR adds logging details to pushNotifier.ts for status code and errors given when notifications are sent.

### DIFF
--- a/src/pushNotifier.ts
+++ b/src/pushNotifier.ts
@@ -24,15 +24,29 @@ export class PushNotifier {
 		this.url = url;
 	}
 
-	notify({ title = "cross-seed", body, ...rest }: PushNotification): void {
+	async notify({
+		title = "cross-seed",
+		body,
+		...rest
+	}: PushNotification): Promise<void> {
 		if (this.url) {
-			fetch(this.url, {
-				method: "POST",
-				headers: { "Content-Type": "application/json" },
-				body: JSON.stringify({ title, body, ...rest }),
-			}).catch(() => {
-				logger.error({ message: "Failed to send push notification" });
-			});
+			try {
+				const response = await fetch(this.url, {
+					method: "POST",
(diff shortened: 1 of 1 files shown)


src/pushNotifier.ts
```diff
@@ -24,15 +24,29 @@ export class PushNotifier {
...
-				method: "POST",
-				headers: { "Content-Type": "application/json" },
-				body: JSON.stringify({ title, body, ...rest }),
-			}).catch(() => {
-				logger.error({ message: "Failed to send push notification" });
-			});
+			try {
+				const response = await fetch(this.url, {
+					method: "POST",
+					headers: { "Content-Type": "application/json" },
```
mmgoodnow, line 36: mind sticking a `User-Agent` header here while we're at it? If you search through the codebase for `User-Agent` you should be able to find the standard user agent

src/pushNotifier.ts
```diff
@@ -24,15 +24,29 @@ export class PushNotifier {
...
+				});
+
+				if (!response.ok) {
+					logger.error({
+						message: `Failed to send push notification (Status Code: ${response.status} ${response.statusText})`,
+					});
+				}
+			} catch (error) {
+				logger.error({
+					message: `Failed to send push notification (Error: ${error})`,
```
mmgoodnow, line 47: a convention I've developed is doing `logger.debug(error)` after `logger.error(whatever)`, but not putting an error object directly into a template string.

I realized far too late that Winston doesn't log stack traces by default, but it's extensible enough that we can implement a custom formatter that special cases on `Error` objects and also logs their stack trace. And once we do that, we can get this error's stack trace to log "for free" if we use the convention.

mmgoodnow (dismissed):
looks good!

PR #593, opened 2024-02-03
add(group-match): uses compareLocale for startsWith

NooNameR:
As discussed using `localeCompare` for case insensitive comparison:
https://stackoverflow.com/questions/2140627/how-to-do-case-insensitive-string-comparison/2140723#2140723

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -18,7 +18,7 @@ import {
 	parseTorrentFromURL,
 	SnatchError,
 } from "./torrent.js";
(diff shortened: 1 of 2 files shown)


NooNameR:
this is the problem this PR solves:
```typescript
console.log("FrameStor".localeCompare("FrámeStor", undefined, { sensitivity: "base" }) === 0); //true
console.log("FrameStor".toLowerCase().startsWith("FrámeSt".toLowerCase())) //false
```

PR #597, opened 2024-02-07
update(decide/logging): distinguish size over filetree mismatch when rejecting

zakkarry:
when rejecting a release in decide.ts - if the rejection is due to file size distinguish between this and general file-tree mismatch

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -129,7 +129,7 @@ async function assessCandidateHelper(
 	searchee: Searchee,
 	hashesToExclude: string[]
 ): Promise<ResultAssessment> {
-	const { matchMode } = getRuntimeConfig();
+	const { matchMode, linkDir } = getRuntimeConfig();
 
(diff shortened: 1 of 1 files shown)


src/decide.ts
```diff
@@ -153,17 +153,24 @@ async function assessCandidateHelper(
...
+	if (!searchee.path && !linkDir) {
 		return { decision: Decision.FILE_TREE_MISMATCH };
 	}
+	// linkdir exists and is torrent, so torrent should be linked?
 	if (
+		searchee.path &&
 		matchMode == MatchMode.RISKY &&
-		!statSync(searchee.path).isDirectory() &&
-		compareFileTreesIgnoringNames(candidateMeta, searchee)
+		!statSync(searchee.path).isDirectory() //&&  not sure if this is necessary anymore with datamatching - we tested this and files can be linked into dirs and shit
```
mmgoodnow, line 172: this !isDirectory check can likely be changed to searchee.files.length === 1

PR #599, opened 2024-02-17
feat(config/logging): error messaging on invalid syntax in config file

mmgoodnow:
If a user's config file has invalid syntax, cross-seed throws the raw `SyntaxError` from node trying to `require` the file. This PR adds some error handling to catch-and-rethrow a friendlier, more actionable error message than what the SyntaxError showed.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -38,6 +38,7 @@ try {
 		console.error(e.message);
 		process.exit(1);
 	}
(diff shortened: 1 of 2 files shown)


mmgoodnow:
seems like it's working now, not sure I changed anything??…

mmgoodnow:
image

PR #609, opened 2024-03-08
chore: move category and tag to constants

fabricionaweb:
Moving the tag and category to `constants.ts`

This would make it easy to patch and also reduce magic const.

There is also one fix on `cmd.ts`, it was breaking on me due missing array errors, so I just made an safe type there.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -5,8 +5,8 @@ const packageDotJson = require("../package.json");
 export const PROGRAM_NAME = packageDotJson.name;
 export const PROGRAM_VERSION = packageDotJson.version;
 export const USER_AGENT = `CrossSeed/${PROGRAM_VERSION}`;
-export const TORRENT_TAG = PROGRAM_NAME;
-export const TORRENT_CATEGORY_SUFFIX = `.${PROGRAM_NAME}`;
+export const TORRENT_TAG = 'cross-seed';
+export const TORRENT_CATEGORY_SUFFIX = `.cross-seed`;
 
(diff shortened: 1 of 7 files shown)


zakkarry:
I'm not sure if this would benefit at all, except to minimize the necessary consts, but couldn't we use just one variable and inline a period for the categories/labels when the suffix is used?

zakkarry:
I'd opt for a constant for the period, which perhaps we could later integrate into the config file if we chose, and then a constant for 'cross-seed' - and then just chain them where necessary.

Just my opinion.

src/constants.ts
```diff
@@ -5,6 +5,8 @@ const packageDotJson = require("../package.json");
 export const PROGRAM_NAME = packageDotJson.name;
 export const PROGRAM_VERSION = packageDotJson.version;
 export const USER_AGENT = `CrossSeed/${PROGRAM_VERSION}`;
+export const TORRENT_TAG = PROGRAM_NAME;
+export const TORRENT_CATEGORY_SUFFIX = `.${PROGRAM_NAME}`;
```
mmgoodnow, line 9: I think this is more indirection than we really need, and even if they're the same now, we could change the "stylized" version of the PROGRAM_NAME and wouldn't necessarily want to change the tag. 

```suggestion
export const TORRENT_TAG = 'cross-seed';
export const TORRENT_CATEGORY_SUFFIX = `.cross-seed`;
```

mmgoodnow (commented):
LGTM pending the one nit.

PR #613, opened 2024-03-14
Retracker

mmgoodnow:


src/decide.ts
```diff
@@ -85,6 +85,20 @@ export function compareFileTrees(
...
+	const cmp = (elOfA: File, elOfB: File) => {
+		const lengthsAreEqual = elOfB.length === elOfA.length;
+		const pathsAreEqual = elOfB.name === elOfA.name;
+		return lengthsAreEqual && pathsAreEqual;
+	};
+
+	return candidate.files.every((elOfA) =>
+		searchee.files.some((elOfB) => cmp(elOfA, elOfB))
+	);
+}
```
mmgoodnow, line 101: I can't tell the difference between this function and `compareFileTrees` above it. is it in progress?

src/action.ts
```diff
@@ -123,40 +124,86 @@ export async function performAction(
...
-		logInjectionResult(result, tracker, newMeta.name);
-		if (result === InjectionResult.FAILURE) {
-			saveTorrentFile(tracker, getTag(searchee.name), newMeta);
-		}
-		return result;
-	} else {
+	// NEW compare() for STRICTLY COMPARING FILES (nested linking)
+	const nestedMatch = compareFileTreesIgnoringFolders(newMeta, searchee);
+	const perfectMatch = compareFileTrees(newMeta, searchee);
+
```
mmgoodnow, line 187: I think it'd be good to keep the decision code separate from the action code. 
Could we achieve this by checking that the decision was `MATCH` vs `MATCH_SIZE_ONLY`? or maybe do we need to add another member to that `Decision` enum to represent this? 

```suggestion
	// NEW compare() for STRICTLY COMPARING FILES (nested linking)
	const isNestedMatch = compareFileTreesIgnoringFolders(newMeta, searchee);
	const isPerfectMatch = compareFileTrees(newMeta, searchee);

```

src/action.ts
```diff
@@ -123,40 +124,86 @@ export async function performAction(
...
+	}
+
+	let destinationDir: string | undefined,
+		ogDownloadDir,
+		linkedFilesRootResult;
+
+	// SET THE CLIENTS SAVE PATH - PREVENTS ERROR IF DATADIR/LINKDIR IS SET
+	// BUT NOT THE TORRENT DATA DIRECTORY AND/OR LINKING THROWS OR IS NOT SET
+	if (searchee.infoHash) {
+		ogDownloadDir = await getClient().getDownloadDir(searchee);
```
mmgoodnow, line 144: do we need ogDownloadDir to be defined up on line 138? I see some commented code below that references it, not sure what the status of that is. If it's unnecessary, I think `ogDownloadDir` could be changed to a const inside this `if` block.
```suggestion
		const ogDownloadDir = await getClient().getDownloadDir(searchee);
```

src/action.ts
```diff
@@ -123,40 +124,86 @@ export async function performAction(
...
+		destinationDir = ogDownloadDir.isOk()
+			? ogDownloadDir.unwrapOrThrow()
+			: undefined;
+	}
+	if (
+		linkDir &&
+		typeof destinationDir === "string" &&
+		existsSync(destinationDir)
```
mmgoodnow, line 152: so this check that the destination dir exists, is that the main addition that's preventing error here? what was the scenario that causes it? I'm wondering if we can handle the error inside `linkAllFilesInMetafile` because that's already calling `getDownloadDir`.

src/action.ts
```diff
@@ -123,40 +124,86 @@ export async function performAction(
...
-	if (linkDir) {
-		const linkedFilesRootResult = await linkAllFilesInMetafile(
+	// SAVES FIRST, PREVENTS LINKING THOUGH IN THIS ORDER
+	if (action === Action.SAVE) {
+		saveTorrentFile(tracker, getTag(searchee.name), newMeta);
+		const styledName = chalk.green.bold(newMeta.name);
+		const styledTracker = chalk.bold(tracker);
+		logger.info(`Found ${styledName} on ${styledTracker} - saved`);
+		return SaveResult.SAVED;
+	}
```
mmgoodnow, line 135: I like how you moved this up to the top, it reduces complexity a lot.

src/action.ts
```diff
@@ -123,40 +124,86 @@ export async function performAction(
...
+
+	// SO SORRY!!!! TERNARY HELL D:
+	const downloadDir =
+		linkedFilesRootResult && linkedFilesRootResult.isOk()
+			? destinationDir
+			: perfectMatch
+			? destinationDir
+			: newMeta.isSingleFileTorrent && nestedMatch
+			? posix.join(destinationDir, searchee.name) // using posix for dev env only. needs to be removed.
+			: undefined;
```
mmgoodnow, line 196: ideally this logic lives inside `linkAllFilesInMetafile` and the result is accessible as `linkedFilesRootResult.unwrapOrThrow()`

mmgoodnow (commented):
I'm not sure I fully understand the problem this solves yet - I'll be able to give it a more informed review the second time.

src/decide.ts
zakkarry: It compares only the filenames, not the paths. This differs when the files are nested for path but not for name.

mmgoodnow: I think it's comparing paths per lines 94 and 95

src/action.ts
zakkarry: That commented code was for reference from the old previous code, it's irrelevant now.

It can be.

src/action.ts
zakkarry: Basically preventing the situation where linkDir is defined from, for instance, a media library from radarr - but your actual torrent data folder is not. Thus, it would ENOENT since the path it gets from the client is non-existent in the container.

This is mostly relevant for docker containers.

src/action.ts
zakkarry: I think something we may run into with that approach without some changes to the `Decision` return is that we could have multi file torrents match on size that are completely different. This basically will determine if they are the same torrents but nested, or perfect matches.

zakkarry:
> I'm not sure I fully understand the problem this solves yet - I'll be able to give it a more informed review the second time.

My basic thought process was to minimize a lot of the nested code, although the ternary makes me cringe still, and simplify the flow. Linking is a bit complex though, and some of the readability may be lost initially when you look at it.

We can discuss some of this when you have time on Discord if you want.

src/decide.ts
zakkarry: Yes, sorry. Pushing that now.

Issue #610, opened 2024-03-09
feat(decide/match): Allow partial size match above ratio to eliminate nearly all false negatives


ShanaryS:
## Summary
This feature solves the problem where the presence of a .nfo or subtitles prevent a cross-seed from happening. Essentially holding a multi gb torrent hostage due to a couple bytes. I ran this over 100 torrents and got an extra 150 matches ON TOP of the standard data-based risky. Essentially, this feature brings to false negatives to pretty much zero while adding ZERO false positives. It is exactly as safe as risky data-based matching.

For example, if you have a torrent with a single file ```ubuntu.iso``` and there exists a candidate with files ```ubuntu.iso, ubuntu.nfo```, currently the match will fail and no cross-seeding happens. This feature allows the candidate to match with whatever files present as long as it's above the ratio the user specified e.g 0.99. This essentially solves #357 and #160 in a more general way as the type of file isn't the problem, just their size.

## Implementation

Here is a commit with the implementation: https://github.com/cross-seed/cross-seed/commit/c9f95af

In the decide stage, the candidate file tree is compared for file size matches and is successful if the total matched size is above the ratio. If there is only a single file in the candidate, then it must be an exact size match. This is done as a last step when it would have otherwise reported file tree mismatch. I also implemented the reordering in #597 as it was necessary.

In the action stage, if it's a data based search we hardlink the files needed by matching file size. If either candidate or searchee has duplicate file sizes we save instead as we can't match the files correctly, though this should be rare. If it was a .torrent match, we just save immediately as we don't have the path to hardlink, maybe #572 allows it? I have an external script that fuzzy name search all the torrents and hardlinks that way but that may be out of scope for this project.

Now of course, however it gets to the client it will be stuck at 99.9% and the user would need to manually start the torrent to finish downloading the last few bytes. That could be implemented here but it feels wrong to have cross-seed initiate a download. This can be easily automated elsewhere.

This requires the risky match mode and minParitalMatch to be below 1. If the minPartialMatch is at 1, then the feature is disabled and works as normal. If it's below 1 then it will perform the new behavior. Since this matches multiple files, you could set it at 0.99999 to only get complete matches of torrents with multiple files in data based searching, which I believe wasn't done before. We may want to add a lower bound like 0.9 on minPartialMatch to reduce the chance of user error.

zakkarry:
All PR's for non-major bugs should target the v6-unstable branch, just FYI.

[...]

PR #614, opened 2024-03-20
feat(decide/match): Added partial matching

ShanaryS:
Solves: #610

After running this across all my torrents, it doubled the number of cross seeds I had. I was previously using dataDir risky with includeEpisodes and includeNonVideos.

This implementation creates a new match mode called "partial". It uses ```1 - fuzzySizeThreshold``` to determine minPartialMatch as creating a new option is redundant.

The only thing missing is to always recheck and pause if we don't want cross-seed to ever download anything. It requires passing Decision to inject() so all the clients can handle MATCH_PARTIAL as a special case. I opted not to include it now so the main part is easier to review.

I have not tested it on this branch but I've been using it on a branch from master for the past few weeks with qBittorrent.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -73,11 +73,48 @@ function fuzzyLinkOneFile(
 	return join(destinationDir, newMeta.name);
 }
 
+function fuzzyLinkPartial(
+	searchee: Searchee,
+	newMeta: Metafile,
+	destinationDir: string,
+	sourceRoot: string
+): string {
+	let filesToLink: [string, string][] = [];
+	let matchedSizes = 0;
+	let availableFiles = searchee.files.slice();
+	newMeta.files.forEach((newFile) => {
+		let matchedFiles = availableFiles.filter(
+			(file) => file.length === newFile.length
+		);
+		if (matchedFiles.length > 1) {
+			matchedFiles = matchedFiles.filter(
+				(file) => file.name === newFile.name
+			);
+		}
+		if (matchedFiles.length) {
+			matchedSizes += newFile.length;
+			availableFiles.splice(availableFiles.indexOf(matchedFiles[0]), 1);
+			filesToLink.push([matchedFiles[0].path, newFile.path]);
+		}
+	});
+	filesToLink.forEach(([oldPath, newPath]) => {
+		const destFilePath = join(destinationDir, newPath);
+		mkdirSync(dirname(destFilePath), { recursive: true });
+		if (statSync(sourceRoot).isFile())
+		{
+			linkFile(sourceRoot, destFilePath);
+		} else {
+			linkFile(join(dirname(sourceRoot), oldPath), destFilePath);
+		}
+	});
+	return join(destinationDir, newMeta.name);
+}
+
 async function linkAllFilesInMetafile(
 	searchee: Searchee,
 	newMeta: Metafile,
 	tracker: string,
-	decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY
+	decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY | Decision.MATCH_PARTIAL
 ): Promise<
 	Result<
 		string,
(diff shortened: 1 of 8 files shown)


zakkarry:
As i previously told you, our branch is a work in progress and constantly evolving in significant ways.

You are outdated, and have conflicts because of this.

ShanaryS:
Yeah I'm aware of how tricky this is atm due to the active branches. I'll update it once everything else is settled but I wanted to have a draft out.

zakkarry:
Sure no problem, I understand, I just want you to be aware of the potential issues you may encounter.

We think that tracker-linkDir is approaching a ready-to-merge state, and then we can finish our stack error PR  and v6 should be at a somewhat internal testing state.

After those are merged, this PR should be able to be easier to manage base-wise.

ShanaryS:
I did have a question about how the fuzzyLinkOneFile:
https://github.com/cross-seed/cross-seed/blob/23db35945fb67549ae9ff7c96ef3a622a40fb306/src/action.ts#L66-L69

Would sourceRoot and path potentially create x/x/x/y/z instead of x/y/z? Where x is the torrent name and y/z is the file path. I did mine different so that would have to change, though it worked in the branch off master.

zakkarry:
Once again, for specific questions like this, I'd recommend you use discord to clarify things in the dev channel.

I could provide you with specific test case outputs and stuff if necessary.

mmgoodnow:
> Would sourceRoot and path potentially create x/x/x/y/z instead of x/y/z? Where x is the torrent name and y/z is the file path. As I believe sourceRoot and path can contain the torrent name.

not sure what you're asking here, but it'll create a file tree that looks like the candidate torrent, backed by files that came from the original torrent/data files.

mmgoodnow:
oh sorry! didn't mean to close this, I think it happened automatically because I just merged the base branch.

mmgoodnow:
@ShanaryS I think you're going to have to change the base and reopen. It doesn't seem like I have the permissions to do that

mmgoodnow (commented):
so is `partial` effectively a superset of `risky` ?

src/action.ts
```diff
@@ -111,22 +148,26 @@ async function linkAllFilesInMetafile(
...
 		return resultOf(linkExactTree(sourceRoot, fullLinkDir));
-	} else {
+	} else if (decision === Decision.MATCH_SIZE_ONLY) {
 		return resultOf(
 			fuzzyLinkOneFile(searchee, newMeta, fullLinkDir, sourceRoot)
 		);
+	} else {
+		return resultOf(
+			fuzzyLinkPartial(searchee, newMeta, fullLinkDir, sourceRoot)
+		);
```
mmgoodnow, line 158: could there be any "collision" between these two match types that would mean the ordering of these branches matters?

src/decide.ts
```diff
@@ -166,18 +194,36 @@ async function assessCandidateHelper(
...
+			minPartialMatch
+		);
+		if (!partialMatch) {
+			return { decision: Decision.SIZE_MISMATCH };
+		}
+	} else {
+		const sizeMatch = compareFileTreesIgnoringNames(candidateMeta, searchee);
+		if (!sizeMatch) {
+			return { decision: Decision.SIZE_MISMATCH };
+		}
```
mmgoodnow, line 211: if you look at the rest of this function, it's entirely early returns with no nesting. Can we find a way to flatten this logic?

src/decide.ts
```diff
@@ -102,6 +104,30 @@ export function compareFileTreesIgnoringNames(
...
+		if (matchedFiles.length > 1) {
+			matchedFiles = matchedFiles.filter((elOfB) => {
+				elOfA.name === elOfB.name;
+			});
+		}
+		if (matchedFiles.length) {
+			matchedSizes += elOfA.length;
+			availableFiles.splice(availableFiles.indexOf(matchedFiles[0]), 1);
+		}
+	});
```
mmgoodnow, line 127: nit: `for (const elOfA of candidate.files) {}` is preferred over `Array.prototype.forEach`

Can you explain how this `compareFileTreesForPartialMatch` works at a high level?

src/action.ts
```diff
@@ -111,22 +148,26 @@ async function linkAllFilesInMetafile(
...
 	newMeta: Metafile,
-	decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY,
+	decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY | Decision.MATCH_PARTIAL,
 	searchee: Searchee,
 	tracker: string
 ): Promise<ActionResult> {
 	const { action, linkDir } = getRuntimeConfig();
 
-	if (action === Action.SAVE) {
+	if (action === Action.SAVE || (decision === Decision.MATCH_PARTIAL && !linkDir)) {
```
mmgoodnow, line 170: this should be a config schema validation, let's not let people get into this state in the first place.

zakkarry:
I'm not sure if its possible to change the target branch of this PR now, so I think basing off `v6-unstable` and proceeding with the draft would be the way to go.

src/action.ts
ShanaryS: I don't think that's possible. We would want this ordering anyway as it preserves the most information to match with the most restrictive to the least.

src/decide.ts
ShanaryS: It takes each file in the candidate and looks for an exact size match in the searchee. For each file that matches, it's removed from the searchee pool and its size is added to a count. If the ratio of machedSizes to total candidate size is above minPartialMatch, it's successful.

If there are duplicate sizes (https://github.com/cross-seed/cross-seed/issues/610#issuecomment-1987101404), we try to get an exact name match to know which file is correct. If there are still multiple files, we just choose any as if both the size and name matches, it's probably safe. If the name doesn't match when there are duplicates, the file is not counted.

> so is partial effectively a superset of risky ?

Yes. They are the same if minPartialMatch is 1 or it's a single file torrent.

ShanaryR2: I think availableFiles is redundant since due to the name check so it's now just going through searchee.files each time.

ShanaryS: availableFiles is redundant due to checking the file name so it's now using searchee.files each time.

PR #615, opened 2024-03-20
feat(decide/match): partial file tree match mode

ShanaryS:
Replacement for #614 as the previous base branch was deleted and it can no longer be updated. Apologies for the complications.

Implements #610 and all changes in comments and review of previous PR. Also ensuring torrents are added in paused state if skipRecheck is false for partial matching and reporting FILE_TREE_MISMATCH there are duplicate sizes without a name match.

### DIFF
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -3,7 +3,7 @@ import { stat, unlink, writeFile } from "fs/promises";
 import { dirname, join, resolve, sep } from "path";
 import { inspect } from "util";
 import xmlrpc, { Client } from "xmlrpc";
-import { InjectionResult, TORRENT_TAG } from "../constants.js";
+import { Decision, InjectionResult, TORRENT_TAG } from "../constants.js";
 import { CrossSeedError } from "../errors.js";
 import { Label, logger } from "../logger.js";
 import { Metafile } from "../parseTorrent.js";
@@ -276,9 +276,10 @@ export default class RTorrent implements TorrentClient {
 	async inject(
 		meta: Metafile,
 		searchee: Searchee,
+		decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY | Decision.MATCH_PARTIAL,
 		path?: string
 	): Promise<InjectionResult> {
-		const { outputDir } = getRuntimeConfig();
+		const { outputDir, skipRecheck } = getRuntimeConfig();
 
 		if (await this.checkForInfoHashInClient(meta.infoHash)) {
 			return InjectionResult.ALREADY_EXISTS;
@@ -304,6 +305,11 @@ export default class RTorrent implements TorrentClient {
 
 		await saveWithLibTorrentResume(meta, torrentFilePath, basePath);
 
+		const paused =
+			decision === Decision.MATCH_PARTIAL
+				? skipRecheck ? 0 : 1
+				: 0;
+
 		for (let i = 0; i < 5; i++) {
 			try {
 				await this.methodCallP<void>("load.start", [
@@ -312,6 +318,7 @@ export default class RTorrent implements TorrentClient {
 					`d.directory_base.set="${directoryBase}"`,
 					`d.custom1.set="${TORRENT_TAG}"`,
 					`d.custom.set=addtime,${Math.round(Date.now() / 1000)}`,
+					`d.custom2.set=paused,${paused}`,
 				]);
 				break;
 			} catch (e) {
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -276,7 +276,10 @@ export default class RTorrent implements TorrentClient {
 	async inject(
 		meta: Metafile,
 		searchee: Searchee,
-		decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY | Decision.MATCH_PARTIAL,
+		decision:
+			| Decision.MATCH
+			| Decision.MATCH_SIZE_ONLY
+			| Decision.MATCH_PARTIAL,
 		path?: string
 	): Promise<InjectionResult> {
 		const { outputDir, skipRecheck } = getRuntimeConfig();
@@ -307,7 +310,9 @@ export default class RTorrent implements TorrentClient {
 
 		const loadType =
 			decision === Decision.MATCH_PARTIAL
-				? skipRecheck ? "load.start" : "load"
+				? skipRecheck
+					? "load.start"
+					: "load"
 				: "load.start";
 
 		for (let i = 0; i < 5; i++) {
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -305,20 +305,19 @@ export default class RTorrent implements TorrentClient {
 
 		await saveWithLibTorrentResume(meta, torrentFilePath, basePath);
 
-		const paused =
+		const loadType =
 			decision === Decision.MATCH_PARTIAL
-				? skipRecheck ? 0 : 1
-				: 0;
+				? skipRecheck ? "load.start" : "load"
+				: "load.start";
 
 		for (let i = 0; i < 5; i++) {
 			try {
-				await this.methodCallP<void>("load.start", [
+				await this.methodCallP<void>(loadType, [
 					"",
 					torrentFilePath,
 					`d.directory_base.set="${directoryBase}"`,
 					`d.custom1.set="${TORRENT_TAG}"`,
 					`d.custom.set=addtime,${Math.round(Date.now() / 1000)}`,
-					`d.custom2.set=paused,${paused}`,
 				]);
 				break;
 			} catch (e) {
--- a/src/clients/Transmission.ts
+++ b/src/clients/Transmission.ts
@@ -168,7 +168,10 @@ export default class Transmission implements TorrentClient {
 	async inject(
 		newTorrent: Metafile,
 		searchee: Searchee,
-		decision: Decision.MATCH | Decision.MATCH_SIZE_ONLY | Decision.MATCH_PARTIAL,
+		decision:
+			| Decision.MATCH
+			| Decision.MATCH_SIZE_ONLY
+			| Decision.MATCH_PARTIAL,
 		path?: string
 	): Promise<InjectionResult> {
 		const { skipRecheck } = getRuntimeConfig();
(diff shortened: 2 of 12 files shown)


src/decide.ts
```diff
@@ -102,6 +104,45 @@ export function compareFileTreesIgnoringNames(
...
+		let matchedSearcheeFiles = searchee.files.filter(
+			(searcheeFile) => searcheeFile.length === candidateFile.length
+		);
+		if (matchedSearcheeFiles.length) {
+			matchedSizes += candidateFile.length;
+		}
+	}
+	return matchedSizes / candidate.length >= 1 - fuzzySizeThreshold;
+}
+
```
mmgoodnow, line 123: I know my earlier comments asked for a certain code architecture for `assessCandidateHelper`, but I think your initial intuition was better. would you mind changing it back? Sorry for the churn!

src/clients/Transmission.ts
```diff
@@ -186,13 +189,18 @@ export default class Transmission implements TorrentClient {
 
 		let addResponse: TorrentAddResponse;
 
+		const skipRecheckTorrent =
+			decision === Decision.MATCH_PARTIAL
+				? skipRecheck
+				: true;
```
mmgoodnow, line 195: what are you going for with these changes? It looks like you're trying to preserve original behavior, but I think the real story of the original behavior is just that we never finished implementing skipRecheck.

src/clients/RTorrent.ts
```diff
@@ -312,6 +318,7 @@ export default class RTorrent implements TorrentClient {
 					`d.directory_base.set="${directoryBase}"`,
 					`d.custom1.set="${TORRENT_TAG}"`,
 					`d.custom.set=addtime,${Math.round(Date.now() / 1000)}`,
+					`d.custom2.set=paused,${paused}`,
```
mmgoodnow, line 321: for rTorrent, we control paused vs not by using `load` vs `load.start` on line 315.

mmgoodnow (commented):
It's surprising how little code it requires to unlock so much value here! Thanks for engaging on this.

src/decide.ts
ShanaryS: This function was added to distinguish between a SIZE_MISMATCH and FILE_TREE_MISMATCH in partial. I think it makes sense for when the file sizes match but there are duplicates sizes with wrong names. The only other way to do this is to return a decision rather than a bool. The code structure can be preserved if I remove reporting a FILE_TREE_MISMATCH for partial (which should be very rare).

Do you want to keep this behavior or remove it?

src/clients/Transmission.ts
ShanaryS: Yes I wanted to keep original behavior but add torrent in paused state when using match partial, unless they enable skipRecheck. The only place this was necessary was for .torrent matches for partial as data matches already handled this.

ShanaryS:
Just found an insane edge case. If the candidate piece size is larger than the candidate size, rechecking fails in qBit. Which makes sense as pieces are atomic so even though you have 99.9% of the data, you basically have 0.

So I'll have to add a check that piece length is smaller than candidate length.

mmgoodnow:
> If the candidate piece size is larger than the candidate size, rechecking fails in qBit. Which makes sense as pieces are atomic so even though you have 99.9% of the data, you basically have 0.

I can't really see any reason this _wouldn't_ work. Partial pieces are pretty common I think, on most torrents the last piece is partial. If you have the data, I'm not sure what would be wrong

zakkarry:
> > If the candidate piece size is larger than the candidate size, rechecking fails in qBit. Which makes sense as pieces are atomic so even though you have 99.9% of the data, you basically have 0.
> 
> I can't really see any reason this _wouldn't_ work. Partial pieces are pretty common I think, on most torrents the last piece is partial. If you have the data, I'm not sure what would be wrong

Peice size would not affect the data check. Remenants of partial data being downloaded to complete the chunk/peices are typically stored by libtorrent in .part files in the torrent root directory.

zakkarry:
Conflicts now...rebasing is going to be fun. If you need help let me know..

zakkarry:
Rebased for you.

mmgoodnow:
I haven't been able to spend time on this, but it is still on my radar

ShanaryS:
I rebased onto v6 and added decision when logging results.

So I've been using this for a month now from master, will be using on v6 from now on. The benefit you will see from this heavily depends on the initial searchee and if your trackers have the extra files.
* If all your trackers have the extra files or none has it, you will not see a difference. 
* If some trackers have it and some don't, you won't see a difference if you got the one with extra files first. As data based risky would created a searchee without it.

So how many MATCH_PARTIAL you see will be different for everyone.

src/decide.ts
```diff
@@ -102,6 +104,48 @@ export function compareFileTreesIgnoringNames(
...
+export function compareFileTreesPartialIgnoringNames(
+	candidate: Metafile,
+	searchee: Searchee
+): boolean {
+	const { fuzzySizeThreshold } = getRuntimeConfig();
+	let matchedSizes = 0;
+	for (const candidateFile of candidate.files) {
+		let matchedSearcheeFiles = searchee.files.filter(
+			(searcheeFile) => searcheeFile.length === candidateFile.length
+		);
```
mmgoodnow, line 116: it looks like we only use `!!matchedsearcheefiles.length` so this can be changed to `some` which will short-circuit after the first one it finds

```suggestion
		let matchedSearcheeFiles = searchee.files.some(
			(searcheeFile) => searcheeFile.length === candidateFile.length
		);
```

src/decide.ts
```diff
@@ -102,6 +104,48 @@ export function compareFileTreesIgnoringNames(
...
+		if (matchedSearcheeFiles.length) {
+			matchedSizes += candidateFile.length;
+		}
+	}
+	return matchedSizes / candidate.length >= 1-fuzzySizeThreshold;
```
mmgoodnow, line 121: this code needs to be run through prettier
`npx prettier --write "src/**/*.ts"`

mmgoodnow (commented):
lgtm pending nits

ShanaryS:
Resolved all comments and reviewed once more. All good for me.

mmgoodnow:
oh shit, I think your npx command invoked a different version of prettier than we're on. I'll resolve that

src/decide.ts
```diff
@@ -102,6 +104,48 @@ export function compareFileTreesIgnoringNames(
...
+	return matchedSizes / candidate.length >= 1 - fuzzySizeThreshold;
+}
+
+export function compareFileTreesPartial(
+	candidate: Metafile,
+	searchee: Searchee
+): boolean {
+	if (candidate.length <= candidate.pieceLength) {
+		return false;
+	}
```
mmgoodnow, line 130: I still don't really understand this bit. what makes it not work when the piece length is too big?

zakkarry: piece length shouldn't have any issue with the data matching unless the candidate is not a complete torrent....i also do not understand what's going on with the piece stuff.

ShanaryS: I've had about 10 ebooks all with 1mb piece size and torrent size less than that. They all are all stuck at 0 after rechecking. Some torrents which are between one and two pieces are above 50% but below the expected amount. Maybe all these happen to be different with the exact same size but there is some weirdness here.

In fact most partial torrents never seem to download exactly what's needed. They all download a bit more, usually one or two pieces or something in between if qbit stats are to be trusted.

The ebooks are all public so I can provide them so you guys can test.

zakkarry: They sound like false positives by what you're saying. The files would otherwise be identical regardless of the piece size, and that's what counts during rechecking.

I made a post, [here](https://forum.deluge-torrent.org/viewtopic.php?p=238007&hilit=pieces#p238007) - the second sentence explains how pieces function relative to the torrent file process. Might help clarify.

I'd ask you to download both of the "matches" separately and run `sha1sum` on both files. Those 10 ebooks with 1MB pieces size that are all less than 1MB would all be 1 piece. If the `sha1sum` of that piece is different, based on rechecking to 0. Thus false-positive. Torrent checking/files use sha1. 

I'd expect the others to be much the same.

zakkarry: So after discussions with @ShanaryS in discord, he showed me what was happening. This is the basic "worst-case" replicable steps.

1. Get a torrent downloaded with a size less than the piece length, quantity of files is irrelevant but 1 makes it easiest.
2. Find a torrent with that same file, and an additional (txt file, for example, i common)
3. add that second torrent (this would be the candidate) to your client.
4. because the entire piece fails in the candidate during sha1 checking, it redownloads the entire piece
5. this effectively makes cross-seeding pointless, as you downloaded the entire torrent again anyway for the missing txt file

This has less effect, but still some, on somewhat larger torrents, but becomes more and more insignificant the larger the torrent is relative to the total size of the torrent you are cross-seeding. It can also be exacerbated by "unlucky" piece distribution or cutoff points.

He has an approach he is going to use (change) that may help to address this incrementally.

ShanaryS: The fix I implemented is to use the ratio of missing pieces rather than missing size instead. This handles torrents with low pieces well. But it will also match less torrents with large piece sizes. For example a 320mb torrent with 16mb piece size requires a 0.05 fuzzy but it may only be missing a file that's a few bytes. Previously that could match with pretty much any non-zero fuzzy but ended up downloading the entire 16mb piece. So the new logic better estimates what needs to be downloaded.

Now there is still a chance that the file being spread across multiple pieces could cause more to be downloaded that expected but we can't know that until it's rechecked. Assuming they have skipRecheck to false the user has to be the one to start it so I think it's fine.

src/decide.ts
```diff
@@ -143,7 +140,9 @@ export function compareFileTreesPartial(
 			matchedSizes += candidateFile.length;
 		}
 	}
-	return matchedSizes / candidate.length >= 1 - fuzzySizeThreshold;
+	const totalPieces = Math.ceil(candidate.length / candidate.pieceLength);
+	const missingPieces = Math.ceil(matchedSizes / candidate.pieceLength);
+	return missingPieces / totalPieces <= fuzzySizeThreshold;
```
mmgoodnow: I like this as it's more accurate to "how much am I willing to download for a partial match"

mmgoodnow (approved):

zakkarry:
rebased for you, i guess its ready for merging since @mmgoodnow approved :O

PR #617, opened 2024-03-22
fix(client/deluge): defend against deluge timeouts

mmgoodnow:
this PR adds a 10 second timeout for all Deluge calls, and also moves error handling for labeling from a global `catch` in `inject` into the `setLabel` method.

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -279,6 +279,7 @@ export default class Deluge implements TorrentClient {
 				return InjectionResult.FAILURE;
 			}
 		} catch (injectResult) {
+			console.error(injectResult);
 			if (injectResult.includes("label.set_torrent")) {
 				logger.warning({
 					label: Label.DELUGE,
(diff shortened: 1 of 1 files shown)


src/clients/Deluge.ts
```diff
@@ -203,6 +218,11 @@ export default class Deluge implements TorrentClient {
 				await this.call<void>("label.add", [label]);
 				await this.call<void>("label.set_torrent", [infoHash, label]);
 			}
+		} catch (e) {
+			logger.warn({
+				label: Label.DELUGE,
+				message: `Failed to label ${infoHash} as ${label}`,
```
zakkarry, line 224: Having moved this from inject() to setLabel() seems to have reduced the error (warning) messaging to not include the `newTorrent.name` - while I don't think its particularly likely to ever be necessary for debugging in a position that we can't still use the infoHash or its position in the logs, I do wonder if there's a reason we shouldn't just pass the newTorrent.name to setLabel as well for more readability in the logs for users.

mmgoodnow: I had the same thought process about having to pass in the name, but it's not hard to do. I'll make the change

PR #618, opened 2024-03-22
Allow user-provided API keys

mmgoodnow:
This solves for some complicated use cases where users want to check their configurations for multiple apps into a repository.


src/configSchema.ts
```diff
@@ -176,6 +176,7 @@ export const VALIDATION_SCHEMA = z
 			.array(z.string())
 			.nullish()
 			.transform((value) => (Array.isArray(value) ? value : [])),
+		apiKey: z.string().min(24).nullish(),
```
zakkarry, line 179: Are we not putting ANY restrictions on the characters we accept?

mmgoodnow: nah I think length is enough

PR #619, opened 2024-03-22
update(config): reasonable defaults on config template, small refactors in fs code

mmgoodnow:
This PR aims to make the onboarding flow easier by setting a bunch of options to opinionated defaults that we have already been widely recommending that people use. With this change, users should be able to jump straight to daemon mode, because the `searchCadence` option is set, and every decision we make on behalf of new users will make the onboarding a lot less complicated.

It also deletes the duplicate `config.template.docker.cjs` file in favor of `config.template.js` - there are only a couple small differences between them, and most volume mounts will need more customization anyway.


src/configuration.ts
```diff
@@ -66,7 +66,7 @@ export function appDir(): string {
 		(process.platform === "win32"
 			? path.resolve(process.env.LOCALAPPDATA!, packageDotJson.name)
 			: path.resolve(process.env.HOME!, `.${packageDotJson.name}`));
-	accessSync(appDir);
+	accessSync(appDir, constants.R_OK | constants.W_OK);
```
mmgoodnow, line 69: I forgot to add the access mode in my "errors on invalid config" PR - this access checks for when the Docker daemon creates the config dir on a user's behalf, owned by root so cross-seed can't do anything if it's running as a non-root user.

src/runtimeConfig.ts
```diff
@@ -29,10 +29,10 @@ export interface RuntimeConfig {
...
 	torrents: string[];
 	port: number | null;
-	searchCadence: number;
-	rssCadence: number;
-	snatchTimeout: number;
-	searchTimeout: number;
+	searchCadence?: number;
+	rssCadence?: number;
+	snatchTimeout?: number;
+	searchTimeout?: number;
```
mmgoodnow, line 35: these are nullish

zakkarry: I'm not sure I understand what you're getting at, in the Zod schema, they are listed as nullish() - and given that they are optional here, it seems I've already got this covered properly?

zakkarry: Did some testing, with both null and undefined it was previously fine, this does not get translated from fileConfig -> runtimeConfig at all if it's undefined, and null gets stored as null.

I'm not sure if you were expecting issues with this, but I don't believe anything was wrong?

src/config.template.cjs
```diff
@@ -94,7 +94,7 @@ module.exports = {
 	 * 		dataDirs: ["/path/here", "/other/path/here"],
 	 * 		dataDirs: ["C:\\My Data\\Downloads"]
 	 */
-	dataDirs: undefined,
+	dataDirs: [],
```
zakkarry, line 97: I like this, I also think that since I've ran across some users thinking that because I included multiple examples, that each path needs it's own line of `dataDirs:` perhaps removing the three examples for just one would be better? Or maybe explicitly stating to only use one and that they are just multiple possible examples.

The only thing is whether we transform an empty array to null or undefined, I believe it is currently transformed to null.

src/config.template.cjs
```diff
@@ -124,7 +124,7 @@ module.exports = {
 	 * cross-seed will use links of this type to inject data-based matches into your client.
 	 * Options: "symlink", "hardlink"
 	 */
```
zakkarry, line 126:
```suggestion
	 * cross-seed will use links of this type to inject data-based matches into your client.
	 * Options: "symlink", "hardlink"
         * Details on linking available below
         * https://www.cross-seed.org/docs/basics/faq-troubleshooting#what-linktype-should-i-use-data-based-searching
         */

File: src/config.template.cjs [hunk 5]
```diff
@@ -142,14 +142,14 @@ module.exports = {
...
-	skipRecheck: false,
+	skipRecheck: true,
 
 	/**
 	 * Determines how deep into the specified dataDirs to go to generate new searchees.
 	 * Setting this to higher values will result in more searchees and more API hits to
 	 * your indexers.
 	 */
-	maxDataDepth: 2,
+	maxDataDepth: 1,
```
zakkarry, line 152: I'm confused on why this is change is necessary, does this prevent diving into nested directories? I think if you have the clients data setup and any sort of sorting, it would prevent effective searching?

I think path depth is something I'm not entirely certain on, though, or the purpose or benefit in reducing this.

src/config.template.cjs
```diff
@@ -222,13 +222,13 @@ module.exports = {
 	 * "2w"
 	 * "3 days"
 	 */
-	excludeRecentSearch: undefined,
+	excludeRecentSearch: "1 year",
```
zakkarry, line 225: I think my approach to searching with cross-seed might be way different than yours. Perhaps this is something better discussed in a chat, because it seems like a potentially involved conversation.

src/config.template.cjs
```diff
@@ -258,7 +258,7 @@ module.exports = {
 	 * "2w"
 	 * "3 days"
 	 */
-	searchCadence: undefined,
+	searchCadence: "1 day",
```
zakkarry, line 261: See above comment about searching/excludeRecentSearch

src/config.template.cjs
```diff
@@ -289,7 +289,7 @@ module.exports = {
 	 * Combine this with "excludeRecentSearch" and "searchCadence" to smooth long-term API usage patterns.
 	 * Default is no limit.
 	 */
-	searchLimit: undefined,
+	searchLimit: 100,
```
zakkarry, line 292: For onboarding, I'm not sure that this would really please everyone. The general response I've gotten is people want the fastest possible search, and doing 100 a day initially (which would be the most common place these settings would be used prior to customizations) would make this a rather long process for many users initially.

zakkarry (commented):
I only did one config file, for obvious reasons (duplicating is dumb) - but here are my first impressions of this.

Also have some comments on your self-review.

src/config.template.cjs
zakkarry: Also as a note, for the v6 docs this URL will need "data-based-searching" part removed.

src/config.template.cjs
mmgoodnow: Seeing as this is just OOTB defaults, I think I'd rather err on the side of less tracker spam here. 
> I think if you have the clients data setup and any sort of sorting, it would prevent effective searching?

Not sure what you mean by sorting, but if you mean putting `/downloads` as your data dir instead of `/downloads/movies` and `/downloads/shows`, that's never been officially supported and I think it makes bogus searches for "shows" and "movies". I think ideally we just ask people do do the latter. It will be fairly apparent what's wrong when people make that mistake as well.

src/config.template.cjs
mmgoodnow: what's your proposal?

src/config.template.cjs
mmgoodnow: I think this is an opportunity for us to be opinionated and friendly to the trackers. It won't satisfy everyone, but if impatient users want it to do everything at once, they can still change the settings. Many users _will_ be okay with it, and it'll lead to better experiences in the long run.

src/config.template.cjs
mmgoodnow: I'm open to an "only search once" type thing, or some other kind of `excludeOlder` type thing.

src/runtimeConfig.ts
mmgoodnow: Not at runtime, but the types were wrong. Without it  marked as optional, we get bogus IDE messages like this, and it also means there's a possibility of introducing a bug later by forgetting to check for null/undefined.

image

mmgoodnow:
I'm sort of wondering…should we still have separate config templates for docker vs not docker? The only differences between them are `torrentDir` and `outputDir`, which maybe doesn't matter as much in the age of links (because people, in most cases, need to make their mounts transparent)

zakkarry:
rebased for you to resolve conflicts.

src/config.template.cjs
stickel: I like this change as well. I wonder if updating the example to this would make it more clear:

```ts
example:
    dataDirs: ["/path/here"], OR
    dataDirs: ["/path/here", "/other/path/here"], OR
    dataDirs: ["C:\\My Data\\Downloads"]
```

My only concern is whether people reading this remove the "OR" or keep it and break the config?

src/config.template.cjs
stickel: Maybe a note in the text above (and/or in the docs) mentioning that if someone is using the default for `maxDataDepth` they should have more explicit `dataDir` paths (`dataDir: ["/path/to/torrents/category"]` vs `dataDir: ["/path/to/torrents"]`).

src/config.template.cjs
```diff
@@ -211,24 +211,25 @@ module.exports = {
...
-	 * more recently than this long ago.
+	 * Exclude torrents which have been searched more recently than this long ago.
+	 * Doesn't exclude previously failed searches.
 	 * Examples:
 	 * "10min"
 	 * "2w"
 	 * "3 days"
 	 */
-	excludeRecentSearch: undefined,
+	excludeRecentSearch: "3 days",
```
stickel, line 226: Does anyone else think this setting should be more strict? By that I mean the options. For example, days can be `3 days` or `3d`. Should it be more opinionated with only one way of choosing periods?

mmgoodnow: Good point. Just because ms supports it doesn't mean we have to tell people all the possible ways they can name their units. I'm in favor of the full length "5 minutes" style.

stickel: Yup, I agree with using the longer form.

src/config.template.cjs
```diff
@@ -122,9 +124,10 @@ module.exports = {
 
 	/**
 	 * cross-seed will use links of this type to inject data-based matches into your client.
+	 * https://www.cross-seed.org/docs/basics/faq-troubleshooting#what-linktype-should-i-use
 	 * Options: "symlink", "hardlink"
 	 */
-	linkType: "symlink",
+	linkType: "hardlink",
```
Audionut, line 130: This would seem to break things when trying to hardlink to differing partition/disk/mount than what is set in `linkDir:`, meaning OOTB, this would seem to break things for those who have media spread.

Would seem to require https://github.com/cross-seed/cross-seed/issues/625 first.

zakkarry: I think at this point, my impression was that we are more opting for these defaults to be recommendations and we are recommending people utilize hardlinks.

These are not necessarily going to work for everyone, all the time, and will always require some level of adjusting or configuration.

I'm not sure having settings that automatically work for everything OOTB because its more or less "inferior" or whatever you want to call it is a route I would opt for...but I'm open for discussions.

src/config.template.cjs
```diff
@@ -167,144 +180,174 @@ module.exports = {
...
 	/**
 	 * The number of searches to be done before it stops.
-	 * Combine this with "excludeRecentSearch" and "searchCadence" to smooth long-term API usage patterns.
-	 * Default is no limit.
+	 * Combine this with "excludeRecentSearch" and "searchCadence" to smooth
+	 * long-term API usage patterns.
+	 * Set to null for no limit.
 	 */
-	searchLimit: undefined,
+	searchLimit: 100,
```
zakkarry, line 328: I still feel that if we are going to suggest this limit, perhaps a disclaimer clarifying this behavior for initial searches might be worth considering.

See previously resolved review comment. https://github.com/cross-seed/cross-seed/pull/619/files#r1536108533

src/config.template.cjs
```diff
@@ -167,144 +180,174 @@ module.exports = {
...
 
 	/**
-	 * The list of infohashes or strings which are contained in torrents name that you want to be excluded from cross-seed
-	 * This is the same format as torznab, surround the entire set of quoted strings in square brackets
-	 * Leave as undefined to disable
+	 * The list of infohashes, or strings which are contained in torrents name
```
zakkarry, line 331:
```suggestion
	 * The list of infohashes, or strings which are contained in torrents name,

**Review by zakkarry — CHANGES_REQUESTED** [review]
Has there been any testing of Zod/Errors since your typing and addZodIssue changes?

I don't see anything that would be an issue, but Zod is a finicky thing.

I also would like to run through the error outputs one more time before merging just to double check if it hasn't been done yet.

File: src/config.template.cjs (hunk 10 above)
Comment (Audionut) [reply src/config.template.cjs:139]: Given the issues I've seen people experience regarding hardlinks and arr's, I'm not convinced this is the appropriate default setting. In my mind, this has more potential for endless complaints than setting for symlinks.

At the end of the day, people *should* research and pick the appropriate setting, but unfortunately, people often don't.

It basically boils down to:
Does the desire to strongly point people to hardlinks, outweigh the amount of bitching likely to be experienced.

Comment (zakkarry) [reply src/config.template.cjs:139]: I understand the potential consequences...perhaps we leave it set to "DIDNT_READ_CONFIG" and link to the docs :P

**Comment by mmgoodnow** [comment]
> Has there been any testing of Zod/Errors since your typing and addZodIssue changes?

I have tested them and everything seems to be working. The changes in `configSchema.ts`, other than `port` changing from `boolean` to `literal(false)`, are just a function rename and adding types where they were implicitly `any`.

File: src/config.template.cjs (hunk 12 above)
Comment (zakkarry) [reply src/config.template.cjs:331]: Not sure if trailing punctuation is discarded or what...actually.

File: src/config.template.cjs (hunk 10 above)
Comment (mmgoodnow) [reply src/config.template.cjs:139]: I think hard links and symlinks both have their own problems, but I'm inclined to listen to the people that have the most experience diagnosing and helping people with issues, which are @zakkarry and @bakerboy448, who have both advocated for hardlinks. If we switch the other way and suddenly see a massive uptick in support requests for hardlink issues, we can always change the recommendation back.

File: src/config.template.cjs [hunk 13]
```diff
@@ -77,87 +77,100 @@ module.exports = {
...
 	 *
-	 * Default: false
+	 * Default: false.
 	 */
 	legacyLinking: false,
 
 	/**
-	 * Whether to skip recheck in Qbittorrent. If using "risky" matchMode it is HIGHLY
-	 * recommended to set this to false.
+	 * Whether to skip recheck in Qbittorrent. If using "risky" matchMode it is
```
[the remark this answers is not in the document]
mmgoodnow: It doesn't apply to rtorrent or transmission - this is part of the corner cutting we did to get data based matching out the door

src/config.template.cjs
```diff
@@ -167,144 +180,174 @@ module.exports = {
...
 	 * Examples:
-	 * "10min"
-	 * "2w"
-	 * "3 days"
+	 * "1 day".
+	 * "3 days".
+	 * "2 weeks".
 	 */
-	searchCadence: undefined,
+	searchCadence: "1 day",
```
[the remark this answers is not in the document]
mmgoodnow: I deleted the "don't run this often" comment as we're recommending more frequent searches with stricter timestamp filtering.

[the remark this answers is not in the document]
zakkarry: Are we thinking that 1d is recommended and not a bit aggressive?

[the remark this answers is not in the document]
mmgoodnow: I want to recommend 1d

Tharic99:
Documentation located here - https://www.cross-seed.org/docs/tutorials/unraid still needs to be updated to reflect the change from config.template.docker.cjs file to the config.template.js file. Current unraid walkthrough fails looking for the old file still.

zakkarry:
> Documentation located here - https://www.cross-seed.org/docs/tutorials/unraid still needs to be updated to reflect the change from config.template.docker.cjs file to the config.template.js file. Current unraid walkthrough fails looking for the old file still.

Thanks for pointing this out, the template link and `cp` command have both been pushed right now. It will take a few minutes to build and deploy to the live docs.

mmgoodnow:
is there a reason we can't do the "run once and it creates the file for you" flow?

zakkarry:
> is there a reason we can't do the "run once and it creates the file for you" flow?

It wasn't what was there when I took them over, so I didn't change it.

I can make those changes though.

zakkarry:
updated for you @mmgoodnow

PR #621, opened 2024-03-23
feat(matching): feature parity for torrent/data matching

mmgoodnow:
This PR makes torrentDir nullable, and adds support for non-linked perfect matches when using data based matching.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -276,7 +276,7 @@ async function findSearchableTorrents() {
 			.map((t) => t.unwrapOrThrow());
 	} else {
 		if (typeof torrentDir === "string") {
-			allSearchees.push(...(await loadTorrentDirLight()));
+			allSearchees.push(...(await loadTorrentDirLight(torrentDir)));
 		}
 		if (Array.isArray(dataDirs)) {
 			const searcheeResults = await Promise.all(
--- a/src/action.ts
+++ b/src/action.ts
@@ -155,6 +155,9 @@ export async function performAction(
 			await saveTorrentFile(tracker, getTag(searchee.name), newMeta);
 			return InjectionResult.FAILURE;
 		}
+	} else if (searchee.path) {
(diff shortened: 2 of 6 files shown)


zakkarry:
rebased for you.

src/pipeline.ts
```diff
@@ -276,7 +276,7 @@ async function findSearchableTorrents() {
 			.map((t) => t.unwrapOrThrow());
 	} else {
 		if (typeof torrentDir === "string") {
```
zakkarry, line 278: I'm not sure it's not a ridiculous thought, but maybe checking for length to make sure someone doesn't think an empty string is the same as disabling?

I know that seems extreme, but I've seen it in support. This can either be done in Zod or here, I'm not sure transforming for them and allowing the presumption that its a valid config is right. I'll yield to you to make that call if you think its a worthwhile addition.

zakkarry (changes requested):
only one note that I see.

src/pipeline.ts
zakkarry: I just realized, this might be addressed already with verifyPath....just need to confirm and other than that it lgtm

mmgoodnow: it ENOENTs

zakkarry: Should we be accounting for this with a transform in Zod or what, if it ENOENT's wouldn't that fail to start?

mmgoodnow: I think it's covered by verifyPath.

mmgoodnow: What would we transform it to?

zakkarry: i guess if its empty null/undefined?

mmgoodnow: I think I'd rather it just be an error. An empty string is most likely a mistake, which throwing an error surfaces.

zakkarry (approved):

PR #624, opened 2024-03-24
add(config/excludeRecentSearch): disclaimer about failed searches

Audionut:
Minor edit to the config to make it clear that previously failed searches will not be excluded, and hence that the usage of excludeRecentSearch can also be used to search for previously failed searched.

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -216,8 +216,8 @@ module.exports = {
 	excludeOlder: "2 weeks",
 
 	/**
-	 * Exclude torrents which have been searched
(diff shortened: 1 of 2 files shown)


zakkarry:
I think given the changes to the config we are making in `reasonable-defauls` already that would be the appropriate branch to target and base off.

Audionut:
If you're happy with this brief description, sure. It seems to still fit ok given the extended description in the proper help.

zakkarry:
Rebased and target the proper branch. Merging.

PR #627, opened 2024-04-04
feat(daemon/announce): daemon functionality without torznab defined

buroa:
I use autobrr to send announcements to cross-seed and do not need any indexers defined in the config as I don't use the search functionality. It would be nice if this warned the user instead of throwing an error.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -100,7 +100,7 @@ export const VALIDATION_SCHEMA = z
 		delay: z.number().nonnegative({
 			message: ZodErrorMessages.delay,
 		}),
-		torznab: z.array(z.string().url()).nullish(),
+		torznab: z.array(z.string().url()),
 		dataDirs: z
 			.array(
 				z
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -100,7 +100,7 @@ export const VALIDATION_SCHEMA = z
 		delay: z.number().nonnegative({
 			message: ZodErrorMessages.delay,
 		}),
-		torznab: z.array(z.string().url()),
+		torznab: z.array(z.string().url()).nullish(),
 		dataDirs: z
 			.array(
 				z
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -437,7 +437,7 @@ export async function validateTorznabUrls() {
 	const indexersWithSearch = await getEnabledIndexers();
 
 	if (indexersWithSearch.length === 0) {
-		throw new CrossSeedError("no working indexers available");
+		logger.warn("no working indexers available");
 	}
 }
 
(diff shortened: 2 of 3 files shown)


onedr0p:
Should this be merged in the v6-unstable branch?

mmgoodnow:
I guess my one worry is if this will negatively impact new users, but a warning seems okay here. @zakkarry what do you think?

zakkarry:
> I guess my one worry is if this will negatively impact new users, but a warning seems okay here. @zakkarry what do you think?

I'm not sure how prevalent this use-case would be, but I don't have any real objections to it.

The only other solution to this use-case I could envision is some sort of CLI arg to circumvent the throw.

zakkarry:
> Should this be merged in the `v6-unstable` branch?

I'll yield to @mmgoodnow on this, I can change the target and rebase if necessary. The configSchema would also need to be adapted to this change as well.

zakkarry:
I've got this rebased and the schema updated locally now.

mmgoodnow:
Yes, let's merge into v6

src/configSchema.ts
```diff
@@ -100,7 +100,7 @@ export const VALIDATION_SCHEMA = z
 		delay: z.number().nonnegative({
 			message: ZodErrorMessages.delay,
 		}),
-		torznab: z.array(z.string().url()),
+		torznab: z.array(z.string().url()).nullish(),
```
mmgoodnow, line 103: how about just an empty array?

zakkarry: I figured that would more likely be an error/misconfiguration compared to deliberately setting undefined or null. Mainly because the default is an empty array.

zakkarry: Maybe a note in the config description about this would be appropriate?

mmgoodnow: we still accept an empty array though, so *also* allowing null doesn't really prevent that misconfiguration from happening. imo the user will still see the warning if it is an accident, and then in the code we don't have to worry about torznab ever being null.

zakkarry:
simple method i tested and seems to work for omitting rss and search jobs when indexers are not defined

mmgoodnow (approved):

src/torznab.ts
```diff
@@ -149,8 +149,10 @@ export async function queryRssFeeds(): Promise<Candidate[]> {
 export async function searchTorznab(
 	name: string
 ): Promise<{ indexerId: number; candidates: Candidate[] }[]> {
-	const { excludeRecentSearch, excludeOlder } = getRuntimeConfig();
-
+	const { excludeRecentSearch, excludeOlder, torznab } = getRuntimeConfig();
+	if (torznab.length === 0) {
+		throw new Error("no working indexers available");
```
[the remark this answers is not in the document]
zakkarry: I force pushed an update to the messaging.

PR #634, opened 2024-04-15
dep(env): change from native to dotenv package

fabricionaweb:
Nodejs have changed the behaviour of --env-file https://github.com/nodejs/node/pull/50588 and this broke our app.

The .env should never be mandatory and environment variables can comes from outside of scope of the app on many ways. There is no reasons to stop the app because of missing this optional file.

Setting up the dotenv package, as it would not break the app when missing .env file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
 				"better-sqlite3": "^9.4.0",
 				"chalk": "^5.0.0",
 				"commander": "^8.3.0",
+				"dotenv": "^16.4.5",
 				"fuse.js": "^6.6.2",
 				"knex": "^2.4.2",
 				"lodash-es": "^4.17.21",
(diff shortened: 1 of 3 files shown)


mmgoodnow:
can we tell users to set the `NODE_OPTIONS` environment variable? `NODE_OPTIONS="--env-file .env cross-seed daemon"`

mmgoodnow:
> Unfortunatelly it doesnt works ☠️

That's unfortunate. I think the best option for now is to tell people to use the long form `node` incantation:

```shell
node --env-file .env "$(npm bin)/cross-seed" daemon
```

When we are targeting v22, we will be able to use `process.loadEnvFile()` wrapped in a `try/catch` so it doesn't throw. Until then, I don't want to introduce `dotenv` which may have slightly different behavior from node's implementation.

Issue #631, opened 2024-04-13
feat(regex): Anime episode matching


ShanaryS:
Some anime release groups follow the standard episode formats and match correctly. However a very large portion does not. They usually follow "[group] title - episode" with variations on that. The goal here to search the non standard anime using "title episode". This gives the best matching and is how sonarr does it (They do a bunch of simultaneous searches using all title variations and search types, often 5+ per indexer). However since we are looking for cross seeds an not just searching, we can assume the naming is the same with one caveat.

Often release will include the English and Japanese title as "title (altTitle)" with no consistency with which goes where. Whatever in the parenthesis is matched as altTitle which could also be used for the search query. Since we don't want to do multiple searches, we could randomly chose between the two if altTitle is present. That should allow both to be searched eventually once excludeRecentSearch rolls over. Some trackers might rename the title to be only in English with their standard so we should cover both.

These searches are also helped by removing all [] brackets as well as Japanese style brackets. The regex accounts for these but it would be helpful even in non anime releases. https://regex101.com/r/m8AO1m/3

I tried to make this match only the anime style but in order to catch as many conventions as possible it can match movies and ep/season regex. So this must be used after those fail.

I tested this across 2000+ releases on nyaa with some old releases and isolated all the unique conventions. It should parse pretty much every thing that EP_REGEX fails to get. In the test string I have some marked as failures as they won't match reformatted titles but they should be rare. I also couldn't find a reasonable way to match altTitle at the start or end which was less common.

https://regex101.com/r/F9uPn0/4

bakerboy448:
Sonarr does not search episode titles except for specials

zakkarry:
Not interested in leveraging outside services as a requirement for standard features, if the torrent names don't come up from a search it kinda is what it is in my view.

This is why I didn't bother with anime and still don't think it's a venture worth pursuing to _MUCH_ extent.

If @ShanaryS can provide something contained that increases cross-seeds without detrimenting anything else, he's free to pursue it. 

Just be advised that we weren't unaware of the lack of functionality as it relates to anime, we didn't want to deal with the headaches.

Issue #629, opened 2024-04-08
bug(regex/group): Release group regex can match resolution


ShanaryS:
This is for both the current regex and v6.

If name ends with WEBRip-720p: https://regex101.com/r/4b9bs3/1

I think XXXp and XXXXp should not match where X is a digit.

---

I think a more general strategy could beneficial. Where we have an additional stricter regex. This would be:
1. Group must be a single word
2. Group cannot be all numbers

All release groups currently found on trash currently follow this with the exception of a few low quality groups. We would use this to match release groups in risky and if it fails we treat it like no group. This allows bad naming conventions to not cause an erroneous release group match while still matching matching with 99% of actual release groups. In safe mode, since the entire name needs to match it can be kept as is.

Downside is potentially more snatches for bad naming conventions (we actually want this behavior since risky snatches if a group is missing). Upside is we allow for more cross seeds for torrents with bad naming.

Here is a comparison:
v6 - https://regex101.com/r/JojVPu/1
Strict - https://regex101.com/r/JojVPu/2

zakkarry:
Not sure if this is perfect, seems to work for me.

https://regex101.com/r/QF836v/1

Play with and let me know if it breaks anywhere.

Also need to know if any release groups potentially would be affected by this.

ShanaryS:
Maybe 1337x would get affected if that's a thing. But I can't think of any release groups like that.

Did a quick test across all my torrents, these are what changed:
```
Starting test
2024-04-08 00:01:34 [32minfo[39m: [search] --- Danny Phantom - S01E01 - Mystery Meat WEBRip-720p.mkv: 720p vs undefined
2024-04-08 00:01:34 [32minfo[39m: [search] --- Danny Phantom - S01E02 - Parental Bonding WEBRip-720p.mkv: 720p vs undefined
2024-04-08 00:01:34 [32minfo[39m: [search] --- Danny Phantom - S01E03 - One of a Kind! WEBRip-720p.mkv: 720p vs undefined
2024-04-08 00:01:34 [32minfo[39m: [search] --- Danny Phantom - S01E04 - Attack of the Killer Garage Sale WEBRip-720p.mkv: 720p vs undefined
2024-04-08 00:01:34 [32minfo[39m: [search] --- Danny Phantom - S01E05 - Splitting Images WEBRip-720p.mkv: 720p vs undefined
[...]

PR #635, opened 2024-04-16
feat(regex): Improving anime matching

ShanaryS:
Implements: #631

To summarize, this feature has zero effect on anything non-anime and correctly parses the vast majority.

---

Did a lot of digging into anime and the best way to match correctly. My goal for this PR is to do it right from the start, it should not be possible to make a meaningfully better regex for anime. I avoided covering edge cases but there is like 5 common formats used and more information to track like [tags] and altTitle, so the regex ended up being long anyways.

I got ~35,000 unique titles from nyaa dating back to 2019 in both English and non-English translated anime. Ran these data progressively through the regex in order (ep, season, movie, anime) noting any improvements that can be made to the existing regex. Each time I would remove what matched to simulate what could reach the next regex.

Due to the nature of anime titles, the regex will get false positives with pretty much anything. However since it will only be used after the other regex fails, it will have zero effect on movies, seasons, and episode matches. It would still match books and music though (which I know is not officially supported). So I opted to add another test in code, to check if the searchee contains at least one video file. Since movies, seasons, and episodes would have already matched, this guarantees whatever remains is an anime. This limits the regex to only being used when searching and tagging saveTorrentFile, but the only other use case is for a db lookup fast path so doesn't really matter.

---

Modifications to existing functionality:
1. Years are only allowed to be between 1800-2099 to reduce false positives (Let's set a reminder to update in 2100)
2. Using trailing [pix] instead of [pi] to detect resolution for cases like 1920x1080 (Didn't update for release group regex)
3. Movie titles can't end with trailing -
4. SEASON_REGEX now supports ~ between season and seasonmax. https://regex101.com/r/eLspOa/2
5. Updated release group regex to one discussed in #629 with added restriction of not all numbers. Lots of anime titles end with ` - ep` or ` - ep.mkv`.
6. Cleansing square brackets (and Japanese brackets) along with separators as they don't contain useful information in the title
7. EP_REGEX supports the common anime format `SXX - E?YY` where the E is optional when SXX is present. The problem is that SEASON_REGEX picks this up where SXX is season and YY is seasonmax when E is missing. Causing a huge number of anime episode matches to be reported as seasons. Out of all the unique anime titles tested, ~10% had this issue. Considering the standard for multi season releases is `SXX-SYY`, sonarr only allows single season releases and most? private trackers ban multi-season releases, I think we should assume anime style over multi season pack in this ambiguous case. https://regex101.com/r/i0igR5/5

Additions:
1. ANIME_REGEX to create queries in format `title episode` parsed from common anime release formatting. The regex is quite long as there a many parts that could be present and need to be handled. I limited the use of * and + to just group, title, and altTitle and used {} to bound the length else where. Most releases contain a [group] tag at the start which often gets renamed to -group at the end on some trackers. Releases will also often contain both the English and Japanese title which either one could be used on other trackers. This is stored into altTitle and is randomly chosen from when deciding to search (either can be English or Japanese). Unless we want to make two simultaneous searches or store the last title used in the db, this is the best compromise.  https://regex101.com/r/F9uPn0/14
2. Only applying anime regex if at least one file is a video. This guarantees the regex won't match music or books. So the anime regex can only be used when searchee is present, which limits it to search queries and tagging .torrents in outputDir (Other use case was querying database which wouldn't change anything).

Shortcomings:
1. Existing regex don't handle alternate titles which helps a lot with anime though I didn't want to change them too much
2. Some anime series release have years which gets match as movies. Nothing can be done but it gets a decent query.
3. Anime can't be matched in safe mode since the release group is at the start and the group regex doesn't look for it (nor should it). Risky would be needed anyways considering how often anime gets renamed between trackers.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -89,8 +89,8 @@ async function findOnOtherSites(
 		.ignore();
 
 	let response: { indexerId: number; candidates: Candidate[] }[];
-	try {
-		response = await searchTorznab(searchee.name);
+	try {		
+		response = await searchTorznab(searchee);
 	} catch (e) {
 		logger.error(`error searching for ${searchee.name}`);
 		logger.debug(e);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -39,7 +39,7 @@ export function wait(n: number): Promise<void> {
 	return new Promise((resolve) => setTimeout(resolve, n));
 }
 
-export function getTag(name: string, isVideo?: boolean): MediaType {
+export function getTag(name: string, isVideo: boolean): MediaType {
 	return EP_REGEX.test(name)
 		? MediaType.EPISODE
 		: SEASON_REGEX.test(name)
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -68,18 +68,22 @@ export function cleanseSeparators(str: string): string {
 		.trim();
 }
 
+function getAnimeQuery(name: string, isVideo: boolean | undefined): string | null {
+	// Must be done after episode, season, and movie regex to be accurate
+	if (!isVideo) return null;
+	const { title, altTitle, release } = name.match(ANIME_REGEX)?.groups ?? {};
+	if (title) {
+		const animeQuery = altTitle && altTitle.length ? Math.random() > 0.5 ? title : altTitle : title;
+		return `${animeQuery} ${release}`;
+	}
+	return null;
+}
+
 export function reformatTitleForSearching(name: string, isVideo?: boolean): string {
 	const seasonMatch = name.match(SEASON_REGEX);
 	const movieMatch = name.match(MOVIE_REGEX);
 	const episodeMatch = name.match(EP_REGEX);
-	let animeQuery: string | undefined = undefined;
-	if (isVideo) {
-		const { title, altTitle, release } = name.match(ANIME_REGEX)?.groups ?? {};
-		if (title) {
-			animeQuery = altTitle && altTitle.length ? Math.random() > 0.5 ? title : altTitle : title;
-			animeQuery += ` ${release}`;
-		}
-	}
+	const animeQuery = getAnimeQuery(name, isVideo);
 	const fullMatch =
 		episodeMatch?.[0] ?? seasonMatch?.[0] ?? movieMatch?.[0] ?? animeQuery ?? name;
 	return cleanseSeparators(fullMatch);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -3,6 +3,7 @@ import {
 	EP_REGEX,
 	MOVIE_REGEX,
 	SEASON_REGEX,
+	ANIME_REGEX,
 	VIDEO_EXTENSIONS,
 } from "./constants.js";
 import { Result, resultOf, resultOfErr } from "./Result.js";
@@ -11,6 +12,7 @@ export enum MediaType {
 	EPISODE = "episode",
 	SEASON = "pack",
 	MOVIE = "movie",
+	ANIME = "anime",
 	OTHER = "unknown",
 }
 
@@ -37,13 +39,15 @@ export function wait(n: number): Promise<void> {
 	return new Promise((resolve) => setTimeout(resolve, n));
 }
 
-export function getTag(name: string): MediaType {
+export function getTag(name: string, isVideo?: boolean): MediaType {
 	return EP_REGEX.test(name)
 		? MediaType.EPISODE
 		: SEASON_REGEX.test(name)
 		? MediaType.SEASON
 		: MOVIE_REGEX.test(name)
 		? MediaType.MOVIE
+		: isVideo && ANIME_REGEX.test(name)
+		? MediaType.ANIME
 		: MediaType.OTHER;
 }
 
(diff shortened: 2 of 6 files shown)


src/utils.ts
```diff
@@ -58,17 +62,26 @@ export async function time<R>(cb: () => R, times: number[]) {
...
 	const movieMatch = name.match(MOVIE_REGEX);
 	const episodeMatch = name.match(EP_REGEX);
+	let animeQuery: string | undefined = undefined;
+	if (isVideo) {
+		const { title, altTitle, release } = name.match(ANIME_REGEX)?.groups ?? {};
+		if (title) {
+			animeQuery = altTitle && altTitle.length ? Math.random() > 0.5 ? title : altTitle : title;
+			animeQuery += ` ${release}`;
+		}
+	}
```
mmgoodnow, line 82: can this be refactored into a helper function

src/utils.ts
```diff
@@ -37,13 +39,15 @@ export function wait(n: number): Promise<void> {
...
 
-export function getTag(name: string): MediaType {
+export function getTag(name: string, isVideo?: boolean): MediaType {
 	return EP_REGEX.test(name)
 		? MediaType.EPISODE
 		: SEASON_REGEX.test(name)
 		? MediaType.SEASON
 		: MOVIE_REGEX.test(name)
 		? MediaType.MOVIE
+		: isVideo && ANIME_REGEX.test(name)
```
mmgoodnow, line 49: it seems like most of the changes here are threading this isVideo boolean around. can it be calculated on the fly? what is it used for?

mmgoodnow:
> Cleansing square brackets (and Japanese brackets) along with separators as they don't contain useful information in the title

is sub group not useful information?

mmgoodnow:
> Due to the nature of anime titles, the regex will get false positives with pretty much anything. However since it will only be used after the other regex fails, it will have zero effect on movies, seasons, and episode matches.

Is there a way we could more reliably test whether something is anime? this feels like it might box us into a corner when it comes to supporting books or music.

mmgoodnow:
> Considering the standard for multi season releases is SXX-SYY, sonarr only allows single season releases and most? private trackers ban multi-season releases, I think we should assume anime style over multi season pack in this ambiguous case.

Solid reasoning.

mmgoodnow:
> Some anime series release have years which gets match as movies. Nothing can be done but it gets a decent query.

Will this be an issue with the tagging? I guess it's a "best effort" situation…

ShanaryS:
> is sub group not useful information?

Yes but not for searching like this as sites may rename it to standard format. I think we should use it but the same way release groups are. It would need to be updated in decide.js when comparing release groups before snatching.

> Is there a way we could more reliably test whether something is anime? this feels like it might box us into a corner when it comes to supporting books or music.

The isVideo check was added to prevent that. We could check the presence of the [tag] at the start but I don't think it's better. As long as we use the anime regex after episodes, seasons, and movies then the isVideo check should guarantee it's anime.

> Will this be an issue with the tagging? I guess it's a "best effort" situation…

Yeah that's why I opted to remove square brackets in the title. If it gets matched as a movie it'll search `title year` which isn't ideal but hard to solve. We could handle this but the movie regex would have to get more complicated with look aheads or adding more logic in code between movies and anime. Both of which I assume would not be desired.

src/utils.ts
ShanaryS: Since this is dependent on searchee I updated it to pass that around instead. `getTag` and `reformatTitleForSearching` still takes a boolean though as it could be used multiple times by the caller.

src/utils.ts
ShanaryS: We could do this as extra searchees, downside is more load on trackers.

To do it at the searchee generation stage it would mean all anime searchees will have their searchee.name updated to have only title or altTitle. Which may actually be good since we can put the [tag] at the end as `-tag` and won't have to update release group parsing.

Otherwise we could handle it right when making queries but some functions would have to change to allow the potential of multiple and we would have to consider how to respect `delay`.

mmgoodnow: I think handling it right when making queries makes the most sense to me. I'd be inclined not to respect `delay` for this case but could be swayed.

ShanaryS: I'm not fully sure this are the only changes necessary to support multiple queries from the same indexer.

Updated so that multiple queries are supported. Right now only anime uses it with altTitle if present but should be easily expandable in the future. If we want to add a delay between the queries we would need to update searchUrls in makeRequests to not be a flat map but a loop.

I also checked the value of adding altTitle to EP_REGEX with the dataset. Only about 1% of the anime matches contains an alternate title. So updating the non anime regex to support altTitle is not worth it.

src/utils.ts
```diff
@@ -68,24 +68,25 @@ export function cleanseSeparators(str: string): string {
 		.trim();
 }
 
-function getAnimeQuery(name: string, isVideo: boolean | undefined): string | null {
-	// Must be done after episode, season, and movie regex to be accurate
-	if (!isVideo) return null;
+export function getAnimeQueries(name: string): string[] {
+	// Only use if getTag returns anime as it's conditional on a few factors
+	let animeQueries: string[] = [];
 	const { title, altTitle, release } = name.match(ANIME_REGEX)?.groups ?? {};
```
mmgoodnow, line 74: what is `release`?

ShanaryS: Usually episode but it can match other numbers so I didn't want to make promises.

zakkarry:
rebased for you

src/pipeline.ts
```diff
@@ -89,8 +89,8 @@ async function findOnOtherSites(
 		.ignore();
 
 	let response: { indexerId: number; candidates: Candidate[] }[];
-	try {
-		response = await searchTorznab(searchee.name);
+	try {
```
mmgoodnow, line 92: looks like this file needs to be prettiered

mmgoodnow (approved):
lgtm ~~pending reformat~~ actually I just remembered I have a PR open to upgrade prettier anyway. LGTM

PR #638, opened 2024-04-18
feat(logging): mismatch decisions note searchee-candidate differences

### LINKED ISSUE #578: Logging Improvement: Verbose log Torznab Size and local size
Title to improve debugging....better yet include this in the decision reject message instead of simply saying the size doesn't match

zakkarry:
closes #578 

this adds verbose logging for the local size and candidate cross-seed size and release group mismatches appended to the `[decide]` line as well as the matching string for blockList when a release is blocked

formatting appended to reason: `(<local> -> <candidate>)`

ex: `2024-04-18 02:34:54 verbose: [decide] Show.S01.1080p.BluRay.x264-GRP- no match for tracker torrent  Show.S01.1080p.BluRay.Remux.DTS-HD.MA5.1.H.264-GRP - its size does not match - (42.6GB -> 103.8GB)`

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -316,9 +316,8 @@ function makeDecisionNote(
 	switch (decision) {
 		case Decision.SIZE_MISMATCH:
 			return ` - (${humanReadableSize(
-				searchee.length,
-				2
-			)} -> ${humanReadableSize(candidate.size, 2)})`;
+				searchee.length
+			)} -> ${humanReadableSize(candidate.size)})`;
 		case Decision.RELEASE_GROUP_MISMATCH:
 			return ` - (${searchee.name
 				.match(RELEASE_GROUP_REGEX)?.[0]
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -28,13 +28,19 @@ export interface ResultAssessment {
 
 const createReasonLogger =
 	(Title: string, tracker: string, name: string) =>
-	(decision: Decision, cached, note: string = ""): void => {
+	(
+		decision: Decision,
+		cached,
+		searchee: Searchee,
+		candidate: Candidate
+	): void => {
 		function logReason(reason): void {
 			logger.verbose({
 				label: Label.DECIDE,
-				message: `${name} - no match for ${tracker} torrent ${Title} - ${reason}${note}`,
+				message: `${name} - no match for ${tracker} torrent ${Title} - ${reason}`,
 			});
 		}
+
 		let reason;
 		switch (decision) {
 			case Decision.MATCH_PARTIAL:
@@ -44,7 +50,9 @@ const createReasonLogger =
 			case Decision.MATCH:
 				return;
 			case Decision.SIZE_MISMATCH:
-				reason = "its size does not match";
+				reason = `its size does not match - (${humanReadableSize(
+					searchee.length
+				)} -> ${humanReadableSize(candidate.size)})`;
 				break;
 			case Decision.NO_DOWNLOAD_LINK:
 				reason = "it doesn't have a download link";
@@ -62,10 +70,17 @@ const createReasonLogger =
 				reason = "it has a different file tree";
 				break;
 			case Decision.RELEASE_GROUP_MISMATCH:
-				reason = "it has a different release group";
+				reason = `it has a different release group - (${searchee.name
+					.match(RELEASE_GROUP_REGEX)?.[0]
+					?.trim()} -> ${candidate.name
+					.match(RELEASE_GROUP_REGEX)?.[0]
+					?.trim()})`;
 				break;
 			case Decision.BLOCKED_RELEASE:
-				reason = "it matches the blocklist";
+				reason = `it matches the blocklist - ("${releaseInBlockList(
+					searchee,
+					getRuntimeConfig().blockList
+				)}")`;
 				break;
 			default:
 				reason = decision;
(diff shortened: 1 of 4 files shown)


src/preFilter.ts
```diff
@@ -65,12 +65,12 @@ export function filterByContent(searchee: Searchee): boolean {
...
 	return blockList.some((blockedStr) => {
-		return (
-			searchee.name.includes(blockedStr) ||
+		return searchee.name.includes(blockedStr) ||
 			blockedStr === searchee.infoHash
-		);
+			? blockedStr
+			: false;
 	});
 }
```
mmgoodnow, line 75: maybe string | undefined or string | null would be a more apt return type for this.

src/decide.ts
```diff
@@ -244,7 +246,41 @@ async function assessAndSaveResults(
...
+				.match(RELEASE_GROUP_REGEX)?.[0]
+				?.trim()})`;
+		case Decision.BLOCKED_RELEASE:
+			return ` - ("${
+				releaseInBlockList(searchee, blockList) as string
+			}")`;
+		default:
+			return "";
+	}
+}
```
mmgoodnow, line 283: any reason not to put this logic inside the switch statement that starts on line 40?

src/decide.ts
```diff
@@ -244,7 +246,41 @@ async function assessAndSaveResults(
...
+function calcSizeLoggingNote(size: number): string {
+	const calc = (size: number, convertBytes: boolean) => {
+		return convertBytes
+			? Math.round((size / (1024 * 1024)) * 100) / 100
+			: Math.round((size / 1024) * 100) / 100;
+	};
+	const sizeMB = calc(size, true);
+	const logNote = sizeMB > 1000 ? `${calc(sizeMB, false)}GB` : `${sizeMB}MB`;
+	return logNote;
+}
```
mmgoodnow, line 258: I found a stack overflow snippet that will handle kB all the way to TB. Let's put it in `utils.ts`

```js

function humanReadableSize() {
    const k = 1000
    const sizes = ['B', 'kB', 'MB', 'GB', 'TB']

    const i = Math.floor(Math.log(bytes) / Math.log(k))

    return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`
}
```

https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript

mmgoodnow (commented):
nice QOL improvement here

src/decide.ts
zakkarry: GiB (1024) vs GB (1000) - while it won't cause a problem if we apply it universally, do you think we should stick to one over the other for any reason?

src/decide.ts
zakkarry: If you look at the amount of stuff, it just seemed to me that it would make the actual decision-based messaging switch a lot more cluttered.

There was not really a reason other than that...

src/decide.ts
mmgoodnow: I think I would rather use the base 10 units than the base 2 units, unless you think the base 2 units are more common in this domain (I'm not sure here)

zakkarry: I'm not sure, I guess the best action would be to stay consistent with what Arrs uses?

src/preFilter.ts
zakkarry: `some` returns a boolean, so I'd have to change it to something like this?

```ts
export function releaseInBlockList(
	searchee: Searchee,
	blockList: string[]
): string | undefined {
	let foundString: string | undefined;

	blockList.some((blockedStr) => {
		if (
			searchee.name.includes(blockedStr) ||
			blockedStr === searchee.infoHash
		) {
			foundString = blockedStr;
			return true;
		}
		return false;
	});

	return foundString;
}
```

mmgoodnow: ah I think this has a bug then. `some` doesn't return the first value you return it that's truthy, it just returns `true`
image

I think Array.prototype.find might be what you need

zakkarry:
```ts
export function releaseInBlockList(
	searchee: Searchee,
	blockList: string[]
): string | undefined {
	return blockList.find((blockedStr) => {
		return (
			searchee.name.includes(blockedStr) ||
			blockedStr === searchee.infoHash
		);
	});
}
```

zakkarry: I guess I just assumed it was truthy because it was a string, not because it was actually `true`...my bad

src/decide.ts
zakkarry: Prowlarr uses base 1024 in it's webui.

I just found a release in the logs, and searched in prowlarr for it. It matches 1024.

mmgoodnow: I still like the stack overflow function. We can change it to 
```js
function humanReadableSize(bytes: number): string {
    const k = 1024
    const sizes = ['B', 'KiB', 'MiB', 'GiB', 'TiB']
    const i = Math.floor(Math.log(bytes) / Math.log(k))
    return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`
}
```

zakkarry: I just made it support both, and default to base 10. Using base 10 since that was your preference, if we want to change later we can just pass an argument.

zakkarry: You're were missing a few variables in there :P I fixed it, just pushed all the changes.

Only outstanding thing is whether you want me to put the notes in the createReasonLogger switch.

src/utils.ts
```diff
@@ -36,7 +36,20 @@ export function nMsAgo(n: number): number {
...
+	ibi: boolean = false
+) {
+	const k = ibi ? 1024 : 1000;
+	const sizes = ibi
+		? ["B", "kiB", "MiB", "GiB", "TiB"]
+		: ["B", "kB", "MB", "GB", "TB"];
+
+	const i = Math.floor(Math.log(bytes) / Math.log(k));
+
+	return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`;
```
mmgoodnow, line 51: I don't think it's really worth it to make this customizable—it just makes it more likely that we'll introduce inconsistencies down the road. this function only really needs a single argument - `dm` can probably be hardcoded to 2

src/decide.ts
mmgoodnow: That seems like a natural place for the logic to go, especially when compared to the fact that we have to re-switch on the same variable in order to do it in this helper function

.github/workflows/ci.yml
```diff
@@ -1,79 +1,72 @@
...
-      - 'master'
+    push:
+        branches:
+            - "**"
+        tags:
+            - "v*"
+    pull_request:
+        branches:
+            - "master"
+    workflow_dispatch:
```
mmgoodnow, line 12: what is workflow dispatch?

zakkarry: When I was trying to figure out why workflows weren't running for those other PRs, I looked up how to be able to trigger them manually.

This allows you to (from actions if its on `master`) or via gh workflow run, to manually trigger workflows to run.

mmgoodnow: ah, I see.

src/preFilter.ts
```diff
@@ -65,8 +65,8 @@ export function filterByContent(searchee: Searchee): boolean {
 export function releaseInBlockList(
```
mmgoodnow, line 65: how about `findBlockedStringInReleaseMaybe`

zakkarry: lol cool

PR #643, opened 2024-04-20
fix(client/qbittorrent): missing message property in error.includes

ShanaryS:
This was called from a feature I'm testing

`2024-04-19 22:01:29 error: TypeError: e.includes is not a function`

This was caused when getting the downloadDir for all searchees without a .path. The error happens when the torrent is not in the client. It could also be that infoHash wasn't defined as I was checking for that.


zakkarry:
Did you see some error when this occurred that clued you in to this bugs existence?

Can you put the error in your PR body?

ShanaryS:
I was working on something else that called this code. The error was not handled so the error was only outputted to terminal.

mmgoodnow:
yes, would like to know more about root cause here.

PR #646, opened 2024-04-20
fix(client/qbittorrent): missing message property in error.includes

ShanaryS:
This was called from a feature I'm testing

`2024-04-19 22:01:29 error: TypeError: e.includes is not a function`

This was caused when getting the downloadDir for all searchees without a .path.

It successfully finds the info hash in client via `isInfoHashInClient` but fails when getting torrentInfo from `getTorrentConfiguration`. The root cause seems to be the /torrents/info from getTorrentConfiguration sometimes failing:

`error: TypeError: fetch failed
at node:internal/deps/undici/undici:12618:11
at async QBittorrent.request (file:///usr/src/cross-seed/dist/clients/QBittorrent.js:47:26)
at async QBittorrent.getTorrentConfiguration (file:///usr/src/cross-seed/dist/clients/QBittorrent.js:122:30)
at async QBittorrent.getDownloadDir (file:///usr/src/cross-seed/dist/clients/QBittorrent.js:105:23)
at async findSearchableTorrents (file:///usr/src/cross-seed/dist/pipeline.js:178:39)
at async Job.main [as exec] (file:///usr/src/cross-seed/dist/pipeline.js:202:42)
at async Job.run (file:///usr/src/cross-seed/dist/jobs.js:26:17)`

Qbit requests failing is nothing new even in v5 so I think .includes just needs to be updated to be on e.message.

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -241,7 +241,7 @@ export default class QBittorrent implements TorrentClient {
 				}
 			}
 		} catch (e) {
-			if (e.includes("retrieve")) {
+			if (e.message.includes("retrieve")) {
 				return resultOfErr("NOT_FOUND");
 			}
 			return resultOfErr("UNKNOWN_ERROR");


zakkarry:
Does updating it to message resolve the issue? Does it always fail otherwise?

ShanaryS:
> Does updating it to message resolve the issue?

The request on /torrents/info still fails but the error is handled properly. So it's treated like it doesn't exist.

> Does it always fail otherwise?

No. I think the request failing is outside of cross seed. Maybe a try catch for all qbit requests and include in for retries?

https://github.com/cross-seed/cross-seed/blob/52ec1c509bc3168350fc53f91ec8b8c154fae53f/src/clients/QBittorrent.ts#L152-L164

zakkarry:
I was seeing something I think was similar in my id-search branch, I tagged you in a comment on what I changed. I think that was network related perhaps, or failure in deluge.

Either way, seems like an issue that should be resolved as much as we can.

mmgoodnow (approved):
It sounds to me that you don't think it's a reproducible error, just a networking blip of sorts. In any case, this change LGTM

PR #648, opened 2024-04-21
fix(client/qbittorrent): unhandled branch causing undefined property access

ShanaryS:
The problem stems from the fact that we are using infoHash in searchee. If the torrent in qbit is using infoHashv2, then it won't return any results.

Calling `isInfoHashInClient` was redundant as it was handled in `getTorrentConfiguration`. This reduces the api calls to qbit and handles the case of missing infoHash.


zakkarry (approved):
lgtm pending @mmgoodnow approval/merge

PR #649, opened 2024-04-21
feat(torznab/indexer): filter out indexers without mediatype category

zakkarry:
filters out indexers per search who dont have the media type being searched for, the result ends up requiring dual messaging for timestamp and category filtering. I'm not sure if this matters, i just merged the message.

currently based off id-search branch, so targeting that for now.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -21,7 +21,7 @@ import { logger } from "./logger.js";
 import { Metafile } from "./parseTorrent.js";
 import { Result, resultOf, resultOfErr } from "./Result.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
-import { Searchee, hasVideo } from "./searchee.js";
+import { Searchee } from "./searchee.js";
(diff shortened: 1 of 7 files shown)


ShanaryS:
Could we leverage this for music, books, and audiobooks? Not full support but to not search them if the indexers don't support it. These can work similar to hasVideo where checking for the presence of .epub/.pdf/etc for books. For music and audiobooks would just have to go under audio as they use the same formats.

zakkarry:
> Could we leverage this for music, books, and audiobooks? Not full support but to not search them if the indexers don't support it. It can work similar to hasVideo where checking for the presence of .epub/.pdf/etc for books. For music and audiobooks would just have to go under audio as they use the same formats.

That's an interesting thought, but we'd have to be able to either identify those based on the other regex's failing, or initially with a percentage of files that matches (but would be 50% match if 1 .pdf and 1 txt file for example) but it is possible. 

We'd just have to work out the methodology.

PR #650, opened 2024-04-21
feat(clear-indexer): adds cli argument for clearing indexer stats

zakkarry:
this clears all values in the indexer table relevant to enabling the indexer.

i don't think we should advertise this command in docs, but i think it would benefit support when a user doesn't have to clear their entire db to get around a rate limiting from torznab that would otherwise last 24hours but was really only necessary to back-off for an hour.

has happened enough times that i wish it was available for support.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -301,6 +301,14 @@ program
 	.command("clear-indexer-failures")
 	.description("Clear the cached details of indexers (failures and caps)")
 	.action(async () => {
+		console.log(
+			"\nIf you've received a '429' (rate-limiting), continuing to hammer",
+			"your indexers may result in negative consequences.",
+		);
+		console.log(
+			"If you have to do this more than once in a short",
+			"period of time, you have bigger issue that need to be addressed.",
+		);
 		await db("indexer").update({
 			active: true,
 			status: null,
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -297,7 +297,20 @@ program
 		});
 		await db.destroy();
 	});
-
+program
+	.command("clear-indexer-failures")
+	.description("Clear the cached details of indexers (failures and caps)")
+	.action(async () => {
+		await db("indexer").update({
+			active: true,
+			status: null,
+			retry_after: null,
+			search_cap: null,
+			tv_search_cap: null,
+			movie_search_cap: null,
+		});
+		await db.destroy();
+	});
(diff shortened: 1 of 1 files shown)


src/cmd.ts
```diff
@@ -297,7 +297,25 @@ program
...
 		await db.destroy();
 	});
-
+program
+	.command("clear-indexer-failures")
+	.description("Clear the cached details of indexers (failures and caps)")
+	.action(async () => {
+		console.log(
+			"\nIf you've received a '429' (rate-limiting), continuing to hammer",
+			"your indexers may result in negative consequences.",
```
mmgoodnow, line 306:
```suggestion
			"If you've received a '429' (rate-limiting), continuing to hammer",
			"your indexers may result in negative consequences.",
```

src/cmd.ts
```diff
@@ -297,7 +297,25 @@ program
...
+		);
+		console.log(
+			"If you have to do this more than once in a short",
+			"period of time, you have bigger issue that need to be addressed.",
+		);
+		await db("indexer").update({
+			active: true,
```
mmgoodnow, line 313: should we really set active to true? we don't do it in line 291

zakkarry: oh, now that you mention it i guess removed trackers would become active then.....

ill fix this.

src/cmd.ts
zakkarry: It appears like on the very next line and i wanted it to stand out more so I added the \n

not a big deal though.

src/cmd.ts
```diff
@@ -297,7 +297,24 @@ program
...
+	.command("clear-indexer-failures")
+	.description("Clear the cached details of indexers (failures and caps)")
+	.action(async () => {
+		console.log(
+			"If you've received a '429' (rate-limiting), continuing to hammer",
+			"your indexers may result in negative consequences.",
+		);
+		console.log(
+			"If you have to do this more than once in a short",
+			"period of time, you have bigger issue that need to be addressed.",
```
mmgoodnow, line 310:
```suggestion
			"period of time, you have bigger issues that need to be addressed.",
```

mmgoodnow (approved):

src/cmd.ts
zakkarry: lol oops, eagle eye 👁️

PR #651, opened 2024-04-21
fix(qbit/injection): correctly assign categories and autotmm

zakkarry:
After the update to the linking, there's now an issue with the logic used in qbittorrent inject() function which leaves dataCategory being assigned for all torrents if linking is enabled.

This addresses that, and fixes autotmm assignment as well.

After this is merged, both `legacyLinking` as true and false will operate with qbit as we intend, and even if qbitmanage is set `force_auto_tmm: true` and forces `autotmm` on - it won't change the path unless the category or savepath is changed (this could be done by cat_change in qbitmanage and needs to be addressed in qbitmanage. there is nothing more we can do from our side)

- [x] test changing cat with autotmm disabled after injection 
- [ ] see if @StuffAnThings can do `cat_change` operations before enabling `autotmm` via `force_autotmm` for torrents that have a save_path in `linkDir` by having `linkDir` path provided in qbm yaml config

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -296,7 +296,7 @@ export default class QBittorrent implements TorrentClient {
 			| Decision.MATCH_PARTIAL,
 		path?: string,
 	): Promise<InjectionResult> {
-		const { duplicateCategories, skipRecheck, dataCategory } =
+		const { duplicateCategories, skipRecheck, dataCategory, linkDir } =
 			getRuntimeConfig();
 		try {
 			if (await this.isInfoHashInClient(newTorrent.infoHash)) {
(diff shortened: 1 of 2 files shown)


zakkarry:
> should we rename it to linkCategory or something?

I really need to discuss what I'm doing about this, I have local changes that are pretty significant.

zakkarry:
> should we rename it to linkCategory or something?

No, this PR fixes this qbit bug, but with wbm when cat_change is used on .cross-seed categories and force_autotmm is enabled it will change the save paths (and 90% of the time leave the old file orphaned since it conflicts) - leaving the cross-seed pointed at the original file unless the cross-seed was named differently i guess.

No issues with Deluge though :O

zakkarry:
Do not delete this branch.

PR #652, opened 2024-04-21
fix(clients): refactor for v6 matching and linking features

zakkarry:
- rename dataCategory to linkCategory
- renamed legacyLinking to flatLinking
- refactored inject() in qbittorrent to tag duplicateCat when linking
- qbittorrent injection now should use flatLinking if using any autotmm
- refactored performAction() in action.ts
- updated config.template.js for new options
- updated configSchema.ts
- change partial match recheck to always
- refactor deluge labeling ternary
- refactor deluge to use Result class
- lots of logic changes for qbittorrent (inject()) formData
- extract recheck logic into a helper function
- #648


mmgoodnow:
can we split this PR into one with just renames and another with logic changes? Alternatively rewriting the history on this branch to do the same thing but with commits

src/utils.ts
```diff
@@ -57,7 +59,14 @@ export function getTag(name: string, isVideo: boolean): MediaType {
...
 }
-
+export function determineSkipRecheck(decision: Decision): boolean {
+	const { skipRecheck } = getRuntimeConfig();
+	return decision === Decision.MATCH_PARTIAL
+		? false
+		: decision === Decision.MATCH
+			? true
+			: skipRecheck;
+}
```
mmgoodnow, line 69: should we just get rid of skipRecheck being configurable at all?

src/decide.ts
```diff
@@ -225,20 +226,9 @@ async function assessCandidateHelper(
...
-		searchee,
-	);
-	if (!partialSizeMatch && matchMode === MatchMode.PARTIAL) {
-		return { decision: Decision.SIZE_MISMATCH };
-	}
 	const sizeMatch = compareFileTreesIgnoringNames(candidateMeta, searchee);
-	if (!sizeMatch && matchMode !== MatchMode.PARTIAL) {
-		return { decision: Decision.SIZE_MISMATCH };
-	}
-
```
mmgoodnow, line 240: why did we need to change the matching logic?

src/clients/QBittorrent.ts
```diff
@@ -329,38 +327,55 @@ export default class QBittorrent implements TorrentClient {
...
+			);
 			// for some reason the parser parses the last kv pair incorrectly
 			// it concats the value and the sentinel
 			formData.append("foo", "bar");
 
 			await this.request("/torrents/add", formData);
-
-			if (path && !skipRecheck) {
+			//if we have a linked file and skiprecheck is false
+			if (determineSkipRecheck(decision)) {
```
mmgoodnow, line 378: why are we no longer checking for path here?

src/clients/QBittorrent.ts
```diff
@@ -329,38 +327,55 @@ export default class QBittorrent implements TorrentClient {
...
 			}
-
+			formData.append("contentLayout", path ? "Original" : contentLayout);
+			formData.append(
+				"skip_checking",
+				determineSkipRecheck(decision).toString(),
+			);
+			formData.append(
+				"paused",
+				!determineSkipRecheck(decision).toString(),
```
mmgoodnow, line 370: these calls to determineSkipRecheck can be extracted into a separate variable

src/clients/QBittorrent.ts
```diff
@@ -234,11 +234,9 @@ export default class QBittorrent implements TorrentClient {
...
 		let torrentInfo: TorrentConfiguration;
 		try {
-			if (await this.isInfoHashInClient(searchee.infoHash!)) {
-				torrentInfo = await this.getTorrentConfiguration(searchee);
-				if (torrentInfo.save_path === undefined) {
-					return resultOfErr("NOT_FOUND");
-				}
+			torrentInfo = await this.getTorrentConfiguration(searchee);
+			if (torrentInfo.save_path === undefined) {
+				return resultOfErr("NOT_FOUND");
```
mmgoodnow, line 239: so was the isInfoHashInClient call redundant here? (looks it otherwise would blow up on the `return resultOf(torrentInfo!.save_path)`? 

can we move that return line inside the try/catch?

src/clients/Deluge.ts
```diff
@@ -326,21 +333,17 @@ export default class Deluge implements TorrentClient {
...
-		const skipRecheckTorrent =
-			decision === Decision.MATCH_PARTIAL ? skipRecheck : true;
 		return [
 			filename,
 			filedump,
 			{
-				add_paused: isTorrent ? !skipRecheckTorrent : !skipRecheck,
-				seed_mode: isTorrent ? skipRecheckTorrent : skipRecheck,
+				add_paused: !determineSkipRecheck(decision),
+				seed_mode: determineSkipRecheck(decision),
```
mmgoodnow, line 346: same thing about extracting a var here

src/utils.ts
zakkarry: In favor of always rechecking risky?

I use `true,` I've had maybe a handful of false positives due to trackers uploading corrupted stuff out of many thousands of matches.

I don't imagine people's experiences would necessarily align with mine given the different level of consistencies amongst tracker's content though.

src/clients/QBittorrent.ts
zakkarry: my thought was that because path was previously exclusive to data-based matching, which we always honored skipRecheck with, but was now solely dependent on whether or not linking was enabled and configured properly to link the searchee, we would just depend on the logic in `determineSkipRecheck`.

Even if it is torrent-backed, we would still be provided a `path` in some (or most if configured) . There is no reason, in my opinion, to calculate rechecking conditional on linking, since it could be a linked torrent-backed searchee we would previously have skipped rechecking on in v5.

I altered the logic to depend solely on the `matchMode`, except for `risky` matches, in which case it would honor `skipRecheck`.

Thoughts?

src/clients/QBittorrent.ts
zakkarry: @ShanaryS pointed out that the logic in `getTorrentConfiguration()` pulls the infohash, and throws if it isn't in the client, which would provide the same logic with fewer API queries to qBittorrent.

src/clients/Deluge.ts
```diff
@@ -276,23 +274,32 @@ export default class Deluge implements TorrentClient {
...
+				const ogLabel =
+					searchee.infoHash && torrentInfo!.label
+						? torrentInfo!.label
+						: this.delugeLabel;
+
+				const shouldSuffixLabel =
+					searchee.infoHash &&
+					duplicateCategories &&
+					!ogLabel.endsWith(this.delugeLabelSuffix) &&
+					ogLabel !== linkCategory;
```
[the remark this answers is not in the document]
zakkarry: `path` is not a metric by which we can determine anything other than a file having been linked now - given that we link everything possible.

src/decide.ts
zakkarry: It seemed more appropriate to check for other matching before checking for partial, and it still makes sense to me. On a prior iteration where I was trying to revert the entire v6 behavior for qbittorrent, it also kept that branch of logic isolated at the end, preferring the other match types based on the order of "best-match-scenario" (and matches that qbittorrent supported would return before doing anything that it wasn't supporting) without the necessity of extra CPU cycles.

@ShanaryS helped with the partial flow that was moved to the end, so if there are questions concerning the more specific ordering in MATCH_PARTIAL returns, they'd be able to shed some light.

src/clients/Deluge.ts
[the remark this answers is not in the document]
mmgoodnow: exactly - we only want to suffix if the file was not linked

src/clients/QBittorrent.ts
```diff
@@ -329,38 +327,51 @@ export default class QBittorrent implements TorrentClient {
...
+					"tags",
+					searchee.infoHash
+						? `${TORRENT_TAG},${newCategoryName}`
+						: category,
+				);
 			} else {
-				formData.append("contentLayout", contentLayout);
-				formData.append("skip_checking", skipRecheckTorrent.toString());
-				formData.append("paused", (!skipRecheckTorrent).toString());
+				// torrent-backed
```
mmgoodnow, line 350:
```suggestion
				// perfect match
```

src/clients/QBittorrent.ts
```diff
@@ -329,38 +327,51 @@ export default class QBittorrent implements TorrentClient {
...
+			// searchee was linked
 			if (path) {
-				formData.append("contentLayout", "Original");
-				formData.append("skip_checking", skipRecheck.toString());
-				formData.append("paused", (!skipRecheck).toString());
+				//because it was a linked file, turn off autotmm
+				//and set a save path and datacat
+				formData.append(
+					"autoTMM",
+					flatLinking && searchee.infoHash ? autoTMM : "false",
```
mmgoodnow, line 338: why does flatLinking matter here?

src/clients/QBittorrent.ts
```diff
@@ -329,38 +327,51 @@ export default class QBittorrent implements TorrentClient {
...
+				// use searchee's autoTMM
+				// we infer this based on no path (linking) being done
+				formData.append("autoTMM", autoTMM.toString());
+				//set category and tags like v5 perfectMatch
+				formData.append("category", newCategoryName);
+				formData.append(
+					"tags",
+					searchee.infoHash
+						? `${TORRENT_TAG},${newCategoryName}`
+						: category,
```
mmgoodnow, line 360: this can be extracted

src/utils.ts
mmgoodnow: always rechecking risky, or never rechecking risky. I don't really care. I just think that the program understands whether to recheck better than the user does.

src/clients/Deluge.ts
```diff
@@ -136,10 +135,7 @@ export default class Deluge implements TorrentClient {
 		const id = Math.floor(Math.random() * 0x7fffffff);
 		const abortController = new AbortController();
 
-		setTimeout(
-			() => void abortController.abort(),
-			ms("10 seconds"),
-		).unref();
+		setTimeout(() => void abortController.abort(), 10000).unref();
```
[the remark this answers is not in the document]
zakkarry: lol, cause after i broke injection for deluge with my label ternary refactor, i removed abort to see if it was your PR that broke it, it wasn't it was me, but when i put it back i just didn't add the import back that i had removed.

I'll put it back.

src/decide.ts
mmgoodnow: I remember this being in the original draft of the partial matching PR. Seems fine to me

src/clients/QBittorrent.ts
mmgoodnow: good change, and well reasoned.

src/clients/QBittorrent.ts
zakkarry: @ShanaryS said that partial can directly point to the original data...so this isn't necessarily a perfect match in the sense that they are identical torrent -> torrent, but data -> torrent in a similar way as a data match would be minus the linking?

file names would be perfect matches, just wasn't trying to be ambiguous about how we refer to perfect matches in other areas.

src/clients/QBittorrent.ts
zakkarry: If flatLinking isn't enabled, we'd have tracker directories, which would then necessitate autotmm be off for that torrent or it'd be moved immediately. they could have it on and then we should use true or false, since they files would be flat., but if we use autotmm that would move it to the flatlining folder anyway and leave empty tracker dirs..


it's annoying.

src/clients/QBittorrent.ts
zakkarry: ill update the comment accordingly though

src/decide.ts
zakkarry: i'm curious if checking something like the piece size or some other meta data to determine rechecking would be a way for us to determine if rechecking risky was necessary.

src/clients/QBittorrent.ts
zakkarry: We warn them if they have flatLinking disabled on qbit, so I think this is proper behavior

autotmm is ignored completely and its disabled on the injected torrent if it would be in a subfolder (tracker dir) of the linkDir....because it would otherwise move if it was autotmm true

if flatlining is enabled, because we inject to one cat, it is irrelevant....let them decide based on their original torrent

src/utils.ts
ShanaryS: If we are to remove the option this is what I think:
* MATCH: Skip rechecking and start unpaused
* MATCH_SIZE_ONLY: Skip rechecking and start unpaused (should error if something is wrong? If not recheck and pause)
* MATCH_PARTIAL: Recheck and pause

src/decide.ts
ShanaryS: The most complete solution would probably be to recheck non perfectMatches and unpause if at 100% when done.

src/utils.ts
zakkarry: >(should error if something is wrong? If not recheck and pause)

What do you mean if something is wrong, what would we measure this by?

src/decide.ts
zakkarry: I think the most easy way to address this is to offer two PARTIAL decisions, SAFE_PARTIAL where linking is disabled or not possible for that data - and a perfect/direct match to the original data can be made, and normal PARTIAL.

but then we would probably have to introduce that as a `matchMode` as well.

You did say that that setup works if linking fails with partial, right? I'm not misunderstanding?

.gitignore
```diff
@@ -41,3 +41,5 @@ cache.json
 docker-compose.yml
 utils.zsh
 test.db
+.idea/discord.xml
+.idea/dictionaries/
```
mmgoodnow, line 45: https://stackoverflow.com/questions/28147162/how-to-create-a-shared-editable-dictionary-for-intellij-idea-spellchecking

src/clients/Deluge.ts
```diff
@@ -262,44 +284,34 @@ export default class Deluge implements TorrentClient {
...
 				"core.add_torrent_file",
 				params,
 			);
-			if (addResult.result) {
-				const { linkingCategory } = getRuntimeConfig();
+
+			const addResponse =
+				typeof addResult.result === "string"
+					? addResult.result
+					: addResult.error;
```
mmgoodnow, line 301: imo this `addResponse` variable that represents multiple values just obscures things. The extraction into this.calculateLabel() is good though

zakkarry: It might be better to review qbit-tags or we merge now - this is outdated severely.

This one was mainly supposed to be the one we rebase and merge into master, so to keep the commits clean enough that that is possible, we branched - qbit has been a mess and therefore we've iterated through probably double digit attempts to nail down this category/tagging/autotmm - that's in qbit-tags. id-search is rebased off qbit-tags presently, and ensemble is rebased off id-search.

gitception.

zakkarry: > imo this `addResponse` variable that represents multiple values just obscures things. The extraction into this.calculateLabel() is good though

Not sure how to do this though, given the different properties and their types..

src/clients/QBittorrent.ts
```diff
@@ -312,61 +380,55 @@ export default class QBittorrent implements TorrentClient {
...
-				formData.append("paused", (!skipRecheckTorrent).toString());
+				// we were provided a path, so set it
+				formData.append("savepath", save_path);
 			}
 
+			formData.append(
+				"autoTMM",
+				flatLinking && searchee.infoHash ? autoTMM.toString() : "false",
+			);
+			formData.append("contentLayout", path ? "Original" : contentLayout);
```
mmgoodnow, line 417: looks like this could be simplified as we're already checking for path higher up
```suggestion
			formData.append(contentLayout);
```

src/clients/QBittorrent.ts
```diff
@@ -215,76 +221,140 @@ export default class QBittorrent implements TorrentClient {
...
+			(torrent) =>
+				hash === torrent.hash ||
+				hash === torrent.infohash_v1 ||
+				hash === torrent.infohash_v2,
 		);
-		const searchResult = JSON.parse(responseText).find(
-			(e) => e.hash === searchee.infoHash,
-		) as TorrentInfo;
-		if (searchResult === undefined) {
+	}
```
mmgoodnow, line 314: is this really the only way to handle info hash v2? I see it gets called in several places now, and fetching the entire torrent list every time feels like it'll be slow, especially on large qbit instances (maybe also slower on instances using fastresume files). would it be worth adding an infoHashV2 field into Metafile?

Also, is there ever a situation we expect multiple values in the returned array? If not, we could use `.find` instead of `.filter` and then return `TorrentInfo | undefined`

src/clients/QBittorrent.ts
```diff
@@ -215,76 +221,140 @@ export default class QBittorrent implements TorrentClient {
...
 			}
 			return resultOfErr("UNKNOWN_ERROR");
 		}
-		return resultOf(torrentInfo!.save_path);
 	}
 
-	async getTorrentConfiguration(
+	/*
+	@param searchee the Searchee we are generating off (in client)
+	@param subfolderLayout whether it is subfolder layout or not
```
mmgoodnow, line 277: is this param still correct? looks like it's a TorrentInfo

src/clients/QBittorrent.ts
```diff
@@ -215,76 +221,140 @@ export default class QBittorrent implements TorrentClient {
...
+	@return string absolute location from client with content layout considered
+	 */
+	async generateCorrectSavePath(
```
mmgoodnow, line 280: looks like this one doesn't need to be async either. also, a naming nit - maybe `getCorrectSavePath`? we're not really generating/producing anything here

src/clients/QBittorrent.ts
```diff
@@ -142,68 +138,78 @@ export default class QBittorrent implements TorrentClient {
...
-		if (maybeNewCategory?.savePath === ogCategory.savePath) {
-			// setup is already complete
-		} else if (maybeNewCategory) {
-			await this.request(
-				"/torrents/editCategory",
-				`category=${newCategoryName}&savePath=${ogCategory.savePath}`,
-				X_WWW_FORM_URLENCODED,
-			);
+		if (searchee.path) {
+			return `${TORRENT_TAG},${TORRENT_TAG}-data`;
```
mmgoodnow, line 207: we're still using cross-seed-data?

src/clients/QBittorrent.ts
```diff
@@ -215,76 +221,140 @@ export default class QBittorrent implements TorrentClient {
...
+	async recheckTorrent(infoHash: string): Promise<void> {
+		const torrent = await this.getTorrentInfo(infoHash);
+		if (torrent.length === 0) {
+			throw new Error("Torrent not found in client");
+		}
+		await this.request(
+			"/torrents/recheck",
+			`hashes=${torrent[0].hash}`,
 			X_WWW_FORM_URLENCODED,
 		);
```
mmgoodnow, line 241: these helpers are great! they make substantially more calls to `this.getTorrentInfo` though. is there a way to minimize them?

src/clients/QBittorrent.ts
```diff
@@ -215,76 +221,140 @@ export default class QBittorrent implements TorrentClient {
...
+
+	async addTorrent(formData: FormData): Promise<void> {
+		await this.request("/torrents/add", formData);
+	}
+
+	async recheckTorrent(infoHash: string): Promise<void> {
+		const torrent = await this.getTorrentInfo(infoHash);
+		if (torrent.length === 0) {
+			throw new Error("Torrent not found in client");
+		}
```
mmgoodnow, line 236: given that this check requires an additional api call, maybe it's better to beg for forgiveness than ask for permission here by checking the result of the `recheck` api call?

mmgoodnow (commented):
ok finally was able to finish a full review. overall I think it looks pretty good.

 at this point I don't totally understand the qbit tag/category behavior but I'm also not a qbit user. As long as we're not painting ourselves into a corner I'm pretty much fine with whatever

src/action.ts
```diff
@@ -134,7 +145,12 @@ async function linkAllFilesInMetafile(
 				e === "NOT_FOUND" ? "TORRENT_NOT_FOUND" : e,
 			);
 		}
-		sourceRoot = join(downloadDirResult.unwrapOrThrow(), searchee.name);
+		sourceRoot = join(
+			downloadDirResult.unwrapOrThrow(),
+			searchee.files.length === 1
+				? searchee.files[0].path
+				: searchee.name,
+		);
```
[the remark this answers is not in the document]
zakkarry: We initially did have this as isSingleFileTorrent, I believe we changed it to accommodate the counter-example you listed if memory serves.

@ShanaryS will need to confirm that my memory is not failing me. There were dozens of iterations of some things here :P

src/configSchema.ts
```diff
@@ -85,7 +87,10 @@ function transformDurationString(durationStr: string, ctx: RefinementCtx) {
  * @return path if valid formatting
  */
 function checkValidPathFormat(path: string, ctx: RefinementCtx) {
-	if (sep === "\\" && !path.includes("\\") && !path.includes("/")) {
+	if (
+		(sep === "\\" && !path.includes(`\x5C`) && !path.includes("/")) ||
+		path === "."
+	) {
```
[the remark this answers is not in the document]
zakkarry: For some reason. previously. using escaped slash was acting strange for me when I was testing why this function didn't work properly, and I got a little frustrated with testing and switched to it to test where the failure point was, I don't remember what my thought process was, but I can change it back now as it wasn't the issue. 

Just a remnant.

.github/ISSUE_TEMPLATE/bug_report.yml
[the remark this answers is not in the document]
zakkarry: cherry picked and merged, when double-checking i noticed I missed two, so there were two merged PR's. oops.

src/action.ts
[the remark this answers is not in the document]
ShanaryS: This check was added because of your counterexample. There would still only be a single file in that just with a path of `Movie/Movie.mkv`. We could add it on searchee but it's a very simple check.

src/clients/Deluge.ts
```diff
@@ -263,43 +285,48 @@ export default class Deluge implements TorrentClient {
...
-									? torrentInfo!.label
-									: `${torrentInfo!.label}${
-											this.delugeLabelSuffix
-										}`
-								: torrentInfo!.label
-							: this.delugeLabel,
+					this.calculateLabel(searchee, torrentInfo!),
 				);
+
+				if (!determineSkipRecheck(decision)) {
```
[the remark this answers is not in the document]
ShanaryS: Right now it's only used for MATCH_SIZE_ONLY, which I think we can just recheck and pause. So only MATCH would not recheck and start unpaused.

src/clients/QBittorrent.ts
```diff
@@ -142,68 +138,78 @@ export default class QBittorrent implements TorrentClient {
...
 			logger.verbose({
 				label: Label.QBITTORRENT,
-				message: "received 403 from API. Logging in again and retrying",
+				message: `Request failed: ${e.message}`,
 			});
-			await this.login();
-			return this.request(path, body, headers, retries - 1);
+			if (retries > 0) {
+				return this.request(path, body, headers, retries - 1);
+			}
```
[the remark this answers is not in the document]
ShanaryS: This was added as `fetch()` sometimes threw an error. The qbit status codes are always 200 for most endpoints and the others don't matter to us.

src/clients/QBittorrent.ts
ShanaryS: Yeah there requests from it to preserve all the filtering options in v5.

src/clients/QBittorrent.ts
ShanaryS: Do you mean to first check with the provided hash? Yeah that would work in most scenarios, only an issue for v2 which should be rare.

src/clients/QBittorrent.ts
ShanaryS: Yeah we could just pass torrent info as an arg and let the caller handle it.

src/clients/QBittorrent.ts
ShanaryS: > is this really the only way to handle info hash v2?

Yes, if we use the v1 infohash, it won't return anything. The only other solution is like you said to calculate v2. But first I'll try to see if I can pass around torrentInfo so we can only call it once.

> Also, is there ever a situation we expect multiple values in the returned array?

I don't think there would be multiple values. qbit uses `torrent.hash` for everything and that can't be duplicated. `torrent.infohash_v2` is more characters than v1 (but gets tructated to 40 for `torrent.hash`) so those would never be duplicated.

src/clients/QBittorrent.ts
zakkarry: I don't know of any PT's that are planning to implement v2 in any sort of haste, mostly no plans at all from what I've seen, however it is somewhat prevalent on certain public.

Given our limited support specifically for public, I'm not sure how much this is beneficial, however given that the check for v1 initially should catch most of these I think it isn't a bad idea to include compatibility.

ShanaryS: The check for all torrents is perfectly fine. It takes about 1 second with 6k torrents using `.fastresume`. With the refactoring I did, this is only called 3 times per inject:
1. Check if new torrent already exists
2. Confirm new torrent is added
3. If searchee has infoHash

src/clients/QBittorrent.ts
ShanaryS: See: https://github.com/cross-seed/cross-seed/pull/652/files#r1586504034

src/clients/Deluge.ts
zakkarry: It seems logical to do the logic off of one variable that we can either be reliant on success based on its type/properties being a string, or of ErrorType in case it fails.

We couple the 4 states this way, it allows the LSP to show what state was eliminated after each check

src/action.ts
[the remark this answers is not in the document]
mmgoodnow: But a multi file torrent with one file wouldn't be changed by subfolder layout since it's already in a subfolder, right? I'm not sure I understand.

src/clients/QBittorrent.ts
mmgoodnow: It shouldn't be hard to also keep track of v2 info hashes. But this PR is already big enough, I think this solution is fine for now.

src/clients/QBittorrent.ts
mmgoodnow: What is the use case?

zakkarry: Wanting to distinguish data specific filtering by tag since all torrents go to the same category now.

src/configSchema.ts
```diff
@@ -15,13 +15,13 @@ const ZodErrorMessages = {
...
 	injectUrl:
 		"You need to specify rtorrentRpcUrl, transmissionRpcUrl, qbittorrentUrl, or delugeRpcUrl when using 'inject'",
-	recheckWarn:
-		"It is strongly recommended to not skip rechecking for risky or partial matching mode.",
-	windowsPath: `Your path is not formatted properly for Windows. Please use "\\\\" or "/" for directory separators.`,
-	qBitFlatLinking:
-		"Using Automatic Torrent Management in qBittorrent without flatLinking enabled can result in injection path failures.",
+	//TODO for some reason this is not tabbed right? needs \t to align. - I think it's because its refined outside of a primitive
+	// I KNOW YOU HATE THIS IM SORRY MG :(
+	windowsPath: `\t\t\tYour path is not formatted properly for Windows. \n\t\t\t\tPlease use "\\\\" or "/" for directory separators.`,
```
mmgoodnow, line 20: what does "this is not tabbed right" mean though 🤨

as far as I'm concerned, "tabbed right" means left-aligned

src/clients/QBittorrent.ts
mmgoodnow: but why do they want to distinguish data specific filtering? do they need to separate the torrents for some reason?

zakkarry: I have no idea these qbit people are nuts.

mmgoodnow: @bakerboy448 could you shed some light here? AFAIK, other than this, we've tried pretty hard to make hard links behave the same way whether the original files are data based or torrent based. why do people need to separate them?

zakkarry: I believe it was @yammes08 and @bakerboy448 both who said that they thought this was the way to preserve ALL of the potential filtering and behaviors of v5.

Why this would be necessary to do, I don't know, but it was possible and that seemed to be a thing they wanted to do.

I don't know the use-case beyond "it was possible" lol

yammes08: This was to preserve the same granularity for filtering that exists for v5. So for example, with the existing configuration you can filter on:
- all cross-seeds (cross-seed tag)
- cross-seed data matches (cross-seed-data category)
- cross-seeds from a specific category (eg. movies.cross-seed)

Now that in in v6 all torrents, if linking is successful, will move into the same category, the additional tags provide an alternative mechanism to perform the same granularity of filters as is possible in v5.

It's not that it is something we/I specifically wanted, but there will be users out there who make use of this level of filtering. The proposed configuration retains that, whilst still reaping the benefits of linking.

bakerboy448: defer to yammes and davo on this as they've been closer in testing / identifying functionality than I.

mmgoodnow: So far nobody has actually said they need this, so I'm inclined to leave it out. it'll be very easy to add back in if someone has an important use case for it.

src/configSchema.ts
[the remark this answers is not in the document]
mmgoodnow: I think the second half of my comment there is still outstanding - what do you think about it?

src/clients/Deluge.ts
```diff
@@ -263,43 +285,48 @@ export default class Deluge implements TorrentClient {
...
 			);
-			if (addResult.result) {
-				const { linkingCategory } = getRuntimeConfig();
+
+			const addResponse =
+				typeof addResult.error?.message === "string"
+					? addResult.error
+					: addResult.result;
+
+			if (typeof addResponse === "string") {
```
mmgoodnow, line 308: I think the part that worries me about this specifically is that `typeof addResponse === "string"` really semantically means "the operation succeeded", but when you read `typeof addResponse === "string"`, nothing in the expression itself indicates any notion of success/failure.

src/clients/QBittorrent.ts
```diff
@@ -142,68 +120,89 @@ export default class QBittorrent implements TorrentClient {
...
-			);
+		if (category.endsWith(TORRENT_CATEGORY_SUFFIX)) {
+			if (duplicateCategories) {
+				return `${TORRENT_TAG},${categoryForTagging}`;
+			} else {
+				return TORRENT_TAG;
+			}
+		}
+		if (searchee.path) {
+			return `${TORRENT_TAG},${TORRENT_TAG}-data`;
```
mmgoodnow, line 200:
```suggestion
			return TORRENT_TAG;
```

src/clients/QBittorrent.ts
```diff
@@ -296,84 +326,88 @@ export default class QBittorrent implements TorrentClient {
...
-				await new Promise((resolve) => setTimeout(resolve, 100));
-				await this.request(
-					"/torrents/recheck",
-					`hashes=${newTorrent.infoHash}`,
-					X_WWW_FORM_URLENCODED,
+			await wait(1000);
+			const newInfo = await this.getTorrentInfo(newTorrent.infoHash);
+			if (!newInfo) {
+				throw new CrossSeedError(
+					`Failed to retrieve torrent after adding`,
```
mmgoodnow, line 399: CrossSeedError is admittedly very poorly named, but is intended to represent errors that we are certain are misconfigurations. In this case we should just use Error.

src/configSchema.ts
mmgoodnow: ok, just delete the commentary and I'll make my peace with the tabs.

mmgoodnow (commented):
Things are looking good here, I think it just needs one more cleanup round and it should be good to go.

.gitignore
mmgoodnow: this is still outstanding

src/configSchema.ts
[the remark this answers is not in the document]
zakkarry: I can look at it, but given that its part of the transform refining, some other things would need to be changed or moved around.

Currently, checkConfigPaths - IIRC - checks permissions and stuff too, so our refinement for that is outside the primitive for formatting.

src/clients/Deluge.ts
zakkarry: the result-deluge branch solves this, so it's kind of irrelevant for this PR to readdress it that way when there's already working code using Result...

src/configSchema.ts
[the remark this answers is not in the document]
mmgoodnow: it doesn't look like this `checkValidPathFormat` actually does any _transformations_, it just makes a check that either passes or fails. So if we just remove all references to it, that should be all we need on this end. 

On the `checkConfigPaths` end, I think as long as whatever user-provided path resolves correctly and is `fs.access`ible, , there's no reason for `cross-seed` to make potentially incorrect assumptions about the path not having the correct separators.

> so our refinement for that is outside the primitive for formatting.

what do you mean by this?

src/clients/Deluge.ts
mmgoodnow: where is that branch? can it be merged into this one?

zakkarry: it could be... its currently the last in the chain of like 6 lol

mmgoodnow: ah never mind then.

zakkarry: So they only place where Deluge.ts is modified after this branch in the "chain" before the result branch is in the "Ensemble matching" commit/branch. 

It could easily target this with some changes, but it would need some minor changes - specifically getDownloadDir() and getAllDownloadDirs() and a labeling return that was incompatible with virtual searchee's.

It is not major, and I could see the value in getting that code in sooner rather than later.

If you want me to do this, I can figure it out rather easily. We just need to make sure that it's handled properly in the ensemble branch which is downstream from this one (this branch -> arr-ids -> ensemble -> inject-saved-torrents -> filter-search-logging -> result-deluge)

Being that it's entirely focused in one commit, it shouldn't be a big deal.

Let me know.

src/configSchema.ts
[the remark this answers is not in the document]
zakkarry: > > so our refinement for that is outside the primitive for formatting.
> 
> what do you mean by this?

Inside the primitive, we transform and addZodIssue() for bad formatting of paths - this is done inside the primitive area (as opposed to the specific refinement we do)

For example...

```js
outputDir: z.string().transform(checkValidPathFormat),
```

Following the initial validation (`validateAndSetRuntimeConfig()`) which checks formatting/syntax - we validate the paths specifically on the fs in `doStartupValidation()`.

I suppose the difference is whether they are being checked for formatting versus permissions/validity on the file system itself. Which would result in a different error "type" - formatting vs exist/perm - and that's more or less why they were separated when I implemented Zod and expanded the startup validation process.

It does make sense to me, and I believe your comment that prompted this was to leave `fs` calls out of Zod given its scope of being schema related only.

If you'd like this to be changed now, I'm just a bit lost on whether you want to take these errors entirely outside of Zod to rely on `fs` calls or if you are saying to put the `fs` calls back into Zod's check...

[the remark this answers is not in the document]
zakkarry: Just to further the logic i held doing this as granularly as I did, some people might not understand why the path without the proper escaped `\` is "wrong" for the config, but I don't think you are advocating for removing that check altogether, right? Just potentially relocating it?

If nothing else that would either 1. move fs calls into Zod - or 2. remove the Zod error'ing format from these types of config mistakes.

src/utils.ts
ShanaryS: If it throws MISSING_FILES or some other error. But are doing recheck and pause of this.

zakkarry: I'm of the mind that, and haven't ever used `skipRecheck: false`, that rechecking even risky at this point is generally unnecessary; that we should only be rechecking partials.

Given the filtering we are doing in decide it should be _VERY_ edge-case that we get a match injected that is not correct. The one caveat being if a tracker uploads corrupted files.

ShanaryS:
All are resolved except for:
1. https://github.com/cross-seed/cross-seed/pull/652#discussion_r1583990369
2. https://github.com/cross-seed/cross-seed/pull/652#discussion_r1590339336

src/utils.ts
mmgoodnow: The failure mode I'm worried about here is rechecking and starting a torrent that turns out to not actually be a match and incurring a large DL on that torrent. As long as whenever we recheck we also pause (which I think we're doing), the impact of this decision is minimized and leaves the door open to changes later. I'm fine either way here.

src/configSchema.ts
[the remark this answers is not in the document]
mmgoodnow: these issues aren't guaranteed to be errors. Relative paths like `foo` are valid paths. Because of that, I don't think we should be treating them as if they _are_ errors. 

> I'm just a bit lost on whether you want to take these errors entirely outside of Zod to rely on fs calls or if you are saying to put the fs calls back into Zod's check...

I think these checks don't belong in the zod layer, but should be performed later on, only once making `fs` calls reveals that a provided path isn't accessible.

src/utils.ts
zakkarry: Yes, anytime rechecks are done in deluge or qbittorrent it is left paused, which prevents any incurred download.

The one edge-case is if pieces do not match in qbittorrent, the behavior I remember in that case is that individually, when pieces are requested and determined to be different, it will request the pieces and subsequently download them and replace the originals.

This is generally only going to be a problem in the case I mentioned, corrupted uploads of a torrent that is otherwise the same size and passes through decide.

PR #654, opened 2024-04-22
feat(torznab/indexer): filter out indexers without audio and book category

ShanaryS:
Since we are restricting searches based on category, we might as well add these in to further limit. Not official support for searching (not even sure if we can make improvements).

I put music and audiobooks into audio since they have the exact same extensions. I also preferred audio over book so if a release contains a book and audio book, it gets marked as audio.

I also put comic/manga with books. As there is some overlap in .pdf but mainly just to reduce clutter.

getTag now just takes a searchee rather than isVideo as the extension checking is done for all media. Should we put getTag in searchee.js? Only uses are with searchees and causes imports to utils.js.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -173,7 +173,7 @@ export async function performAction(
 	const { action, linkDir } = getRuntimeConfig();
 
 	if (action === Action.SAVE) {
-		await saveTorrentFile(tracker, getTag(searchee, searchee), newMeta);
+		await saveTorrentFile(tracker, getTag(searchee), newMeta);
 		const styledName = chalk.green.bold(newMeta.name);
 		const styledTracker = chalk.bold(tracker);
 		logger.info(
@@ -207,7 +207,7 @@ export async function performAction(
 			);
 			await saveTorrentFile(
 				tracker,
-				getTag(searchee, searchee),
+				getTag(searchee),
 				newMeta,
 			);
 			return InjectionResult.FAILURE;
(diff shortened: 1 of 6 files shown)


src/action.ts
```diff
@@ -171,10 +171,9 @@ export async function performAction(
 	tracker: string,
 ): Promise<ActionResult> {
 	const { action, linkDir } = getRuntimeConfig();
-	const isVideo = hasVideo(searchee);
 
 	if (action === Action.SAVE) {
-		await saveTorrentFile(tracker, getTag(searchee.name, isVideo), newMeta);
+		await saveTorrentFile(tracker, getTag(searchee.name, searchee), newMeta);
```
zakkarry, line 176:
```suggestion
		await saveTorrentFile(tracker, getTag(searchee), newMeta);
```
redundant to pass searchee.name and searchee both, just send the searches and pull the .name out of it in getTag...

src/action.ts
```diff
@@ -208,7 +207,7 @@ export async function performAction(
 			);
 			await saveTorrentFile(
 				tracker,
-				getTag(searchee.name, isVideo),
+				getTag(searchee.name, searchee),
```
zakkarry, line 210: same here and anywhere else....

zakkarry (changes requested):
redundancies....

src/action.ts
ShanaryS: I'm dumb. I thought about that when I first added isVideo but forgot about it here...

PR #655, opened 2024-04-22
feat(torznab/indexer): filter out indexers without audio and book category

ShanaryS:
Since we are restricting searches based on category, we might as well add these in to further limit. Not official support for searching (not even sure if we can make improvements).

I put music and audiobooks into audio since they have the exact same extensions. I also preferred audio over book so if a release contains a book and audio book, it gets marked as audio.

I also put comic/manga with books. As there is some overlap in .pdf but mainly just to reduce clutter.

getTag now just takes a searchee rather than isVideo as the extension checking is done for all media. Should we put getTag in searchee.js? Only uses are with searchees and causes imports to utils.js.

### DIFF
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -168,11 +168,7 @@ async function createTorznabSearchQueries(
 		parseInt(str.match(/\d+/)![0]);
 	const relevantIds = await getRelevantArrIds(searchee, ids, caps);
 	const mediaType = getTag(searchee);
-	if (
-		mediaType === MediaType.EPISODE &&
-		caps.tvSearch &&
-		shouldSearchIndexer(mediaType, caps.categories)
-	) {
+	if (mediaType === MediaType.EPISODE && caps.tvSearch) {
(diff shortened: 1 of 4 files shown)


src/torznab.ts
```diff
@@ -249,6 +238,10 @@ function shouldSearchIndexer(mediaType: MediaType, caps: TorznabCats) {
 			return caps.movie;
 		case MediaType.ANIME:
 			return caps.anime;
+		case MediaType.AUDIO:
+			return caps.audio;
+		case MediaType.BOOK:
+			return caps.book;
 		case MediaType.OTHER:
 			return true;
```
zakkarry, line 246: Given our expansion on these types, I think default behavior (when media type is assigned other) should be false

zakkarry, line 246:
```suggestion
			return false;
```
I think we should change this to false now...

zakkarry (changes requested):

zakkarry (approved):

PR #657, opened 2024-04-23
fix(client/rtorrent): send content-type text/xml

zakkarry:
hotio reported that certain builds of rtorrent (particularly jesec's rflood) requires the text/xml for requests. if not provided, rtorrent will potentially return empty strings

rtorrent is the only client implementation in cross-seed that does not currently provide a content-type in the requests.

this adds the header "Content-Type: text/xml" to the call to clientCreator() to fix this issue.

### DIFF
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -120,7 +120,7 @@ export default class RTorrent implements TorrentClient {
 			basic_auth: shouldUseAuth
 				? { user: username, pass: password }
 				: undefined,
-			headers: headers, // Include headers option
+			headers: headers,
 		});
 	}
 
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -111,11 +111,16 @@ export default class RTorrent implements TorrentClient {
 
 		const shouldUseAuth = Boolean(username && password);
 
+		const headers = {
+			"Content-Type": "text/xml",
+		};
+
 		this.client = clientCreator({
 			url: href,
 			basic_auth: shouldUseAuth
 				? { user: username, pass: password }
 				: undefined,
+			headers: headers, // Include headers option
 		});
 	}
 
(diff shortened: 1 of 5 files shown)


src/clients/RTorrent.ts
```diff
@@ -111,11 +111,16 @@ export default class RTorrent implements TorrentClient {
...
+		const headers = {
+			"Content-Type": "text/xml",
+		};
+
 		this.client = clientCreator({
 			url: href,
 			basic_auth: shouldUseAuth
 				? { user: username, pass: password }
 				: undefined,
+			headers: headers, // Include headers option
```
mmgoodnow, line 123: this comment isn't really doing anything
```suggestion
			headers: headers,
```

mmgoodnow (commented):
I think it's already being set so I'm not sure if this is the problem here

https://github.com/baalexander/node-xmlrpc/blob/master/lib/client.js#L49

mrhotio:
False alarm then I guess, I’ll wait and see if any more reports come in… it’s been a while and a quick look at the code didn’t immediately made it obvious the header being set and assumed it could be the same issue as with NZB360… guess not.

zakkarry:
Ok, it wasnt apparent to me and I more or less took your word for it, having had it be an issue in other applications it seemed a reasonable cause then (and the headers were not set, but implicitly set) - so that's a little bit on me too for not actually digging into what was being done.

Thanks for being proactive about it though, seems like something might be going on in rflood or something then?

mrhotio:
Nothing going on with rflood, I should have just verified it again since it's been so long that the issue popped up...but cross-seed isn't exactly the app you can setup in a quick 5 minutes when you aren't knee deep in torrents. Since I was in the #cross-seed channel and it popped into my head, I thought might aswell ask/mention it... You know what they say, assumptions is the mother of all f ups :-)

PR #658, opened 2024-04-23
fix(client/deluge): incremental id with json-rpc requests

zakkarry:
since every id is independent to a connection, using random will create the potential for a bad seed to produce duplicates.

starting at 0, we will now increment through the id last used.

this change was due to new information i was previously unaware of (independent per connection) given to me by a developer on the team

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -50,6 +50,7 @@ export default class Deluge implements TorrentClient {
 	private delugeLabel = TORRENT_TAG;
 	private delugeLabelSuffix = TORRENT_CATEGORY_SUFFIX;
 	private isLabelEnabled: boolean;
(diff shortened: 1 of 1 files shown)


mmgoodnow (approved):
was this causing issues?

zakkarry:
> was this causing issues?

the thing is, it could potential given cross-seed generates the same number twice - i aimed for the upper threshold of Math and modulo...but I don't know.

I was speaking with one of the main developers on Deluge about a bug today and explained some potential causes and this was a theory i said could maybe be connected..

I showed him the id randomize thing, and he said no, its per-connection and that wasn't something i knew or could verify reliably before without tons and tons of digging through source.

I therefore just went with rand, assuming it would probably never cause issues.

Given the new information, i figured it was two lines, just fix it.

PR #662, opened 2024-04-24
bug(client/qbit): Check for subfolder when linking and properly set category

ShanaryS:
`isSubfolderContentLayout()` was basically always returning false. Updated the check and is correct from my testing. Also simplified `getDownloadDir()` as it was making unnecessary calls and checks.

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -224,15 +224,10 @@ export default class QBittorrent implements TorrentClient {
 			if (torrentInfo.save_path === undefined) {
 				return resultOfErr("NOT_FOUND");
 			}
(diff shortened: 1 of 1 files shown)


zakkarry:
we couldn't isolate this logic into getDownloadDir?

ShanaryS:
> we couldn't isolate this logic into getDownloadDir?

Yes... I didn't see the new logic

zakkarry:
> > we couldn't isolate this logic into getDownloadDir?
> 
> Yes... I didn't see the new logic

I extracted a bunch of shit, getDownloadDir should be able to handle this, just fix the referencing/branches a bit.

PR #665, opened 2024-04-26
bug(clients/qbit): Properly handle subfolder layout

ShanaryS:
Did some digging into how subfolder layout is handled. It should be completely fixed now. Problem was if a torrent with a single file that naturally had a folder, it would get the incorrect path.

Also getting torrent info in a way that doesn't rely on infohash versions.

Potentially fixed bug with rechecking searchee instead of newTorrent in deluge.

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -184,14 +184,19 @@ export default class QBittorrent implements TorrentClient {
 			return TORRENT_TAG;
 		}
 		if (category.endsWith(TORRENT_CATEGORY_SUFFIX)) {
-			return `${category},${TORRENT_TAG}`;
+			if (duplicateCategories) {
+				return `${TORRENT_TAG},${category}`;
+			} else {
+				return TORRENT_TAG;
+			}
 		}
 
 		if (searchee.path) {
-			return `${TORRENT_TAG}-data,${TORRENT_TAG}`;
+			return `${TORRENT_TAG},${TORRENT_TAG}-data`;
+		} else if (duplicateCategories) {
+			return `${TORRENT_TAG},${category}${TORRENT_CATEGORY_SUFFIX}`;
 		} else {
-			const suffix = duplicateCategories ? TORRENT_CATEGORY_SUFFIX : "";
-			return `${category}${suffix},${TORRENT_TAG}`;
+			return TORRENT_TAG;
 		}
 	}
 
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -70,7 +70,7 @@ interface TorrentInfo {
 	uploaded_session: number;
 	upspeed: number;
 }
-
+/* Unused variable - removed for build.
 interface TorrentFiles {
 	availability: number;
 	index: number;
@@ -81,7 +81,7 @@ interface TorrentFiles {
 	progress: number;
 	size: number;
 }
-
+*/
 interface TorrentConfiguration {
 	save_path: string;
 	isComplete: boolean;
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -10,10 +10,7 @@ import { Label, logger } from "../logger.js";
 import { Metafile } from "../parseTorrent.js";
 import { getRuntimeConfig } from "../runtimeConfig.js";
 import { Searchee, SearcheeWithInfoHash } from "../searchee.js";
-import {
-	determineSkipRecheck,
-	extractCredentialsFromUrl,
-} from "../utils.js";
+import { determineSkipRecheck, extractCredentialsFromUrl } from "../utils.js";
 import { TorrentClient } from "./TorrentClient.js";
 import { Result, resultOf, resultOfErr } from "../Result.js";
 import { BodyInit } from "undici-types";
@@ -245,8 +242,10 @@ export default class QBittorrent implements TorrentClient {
 		searchee: Searchee,
 		torrentInfo: TorrentInfo,
 	): Promise<string> {
-		const subfolderContentLayout =
-			await this.isSubfolderContentLayout(searchee, torrentInfo);
+		const subfolderContentLayout = await this.isSubfolderContentLayout(
+			searchee,
+			torrentInfo,
+		);
 		if (subfolderContentLayout) {
 			return dirname(torrentInfo.content_path);
 		}
@@ -261,9 +260,11 @@ export default class QBittorrent implements TorrentClient {
 		const responseText = await this.request("/torrents/info", "");
 		const torrents = JSON.parse(responseText);
 		if (hash) {
-			return torrents.filter((torrent) => 
-				hash === torrent.infoHash_v1 ||
-				hash === torrent.infoHash_v2);
+			return torrents.filter(
+				(torrent) =>
+					hash === torrent.infoHash_v1 ||
+					hash === torrent.infoHash_v2,
+			);
 		}
 		return torrents;
 	}
@@ -291,7 +292,10 @@ export default class QBittorrent implements TorrentClient {
 		};
 	}
 
-	async isSubfolderContentLayout(searchee: Searchee, torrentInfo: TorrentInfo): Promise<boolean> {
+	async isSubfolderContentLayout(
+		searchee: Searchee,
+		torrentInfo: TorrentInfo,
+	): Promise<boolean> {
 		if (searchee.files.length > 1) return false;
 		if (dirname(searchee.files[0].path) !== ".") return false;
 		return dirname(torrentInfo.content_path) !== torrentInfo.save_path;
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -401,7 +401,7 @@ export default class QBittorrent implements TorrentClient {
 
 			await this.addTorrent(formData);
 			//if we have a linked file and skiprecheck is false
-			if (skipRecheck) {
+			if (!skipRecheck) {
 				await new Promise((resolve) => setTimeout(resolve, 100));
 				await this.recheckTorrent(newTorrent.infoHash);
 			}
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -1,4 +1,4 @@
-import { join, posix } from "path";
+import { dirname } from "path";
 import {
 	Decision,
 	InjectionResult,
@@ -9,11 +9,10 @@ import { CrossSeedError } from "../errors.js";
 import { Label, logger } from "../logger.js";
 import { Metafile } from "../parseTorrent.js";
 import { getRuntimeConfig } from "../runtimeConfig.js";
-import { Searchee } from "../searchee.js";
+import { Searchee, SearcheeWithInfoHash } from "../searchee.js";
 import {
 	determineSkipRecheck,
 	extractCredentialsFromUrl,
-	stripExtension,
 } from "../utils.js";
 import { TorrentClient } from "./TorrentClient.js";
 import { Result, resultOf, resultOfErr } from "../Result.js";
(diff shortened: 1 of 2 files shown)


src/clients/Deluge.ts
```diff
@@ -316,7 +316,7 @@ export default class Deluge implements TorrentClient {
 					// when paused, libtorrent doesnt start rechecking
 					// leaves torrent ready to download - ~99%
 					await this.call<string>("core.force_recheck", [
-						searchee.infoHash,
+						newTorrent.infoHash,
```
zakkarry, line 319: good catch, i blame it on sleep deprivation lol

src/clients/QBittorrent.ts
```diff
@@ -244,39 +240,53 @@ export default class QBittorrent implements TorrentClient {
...
+	): Promise<string> {
+		const subfolderContentLayout = await this.isSubfolderContentLayout(
+			searchee,
+			torrentInfo,
+		);
 		if (subfolderContentLayout) {
-			return {
-				save_path: join(save_path, stripExtension(searchee.name)),
-			};
+			return dirname(torrentInfo.content_path);
```
zakkarry, line 250: When i was doing research on the subfolder content type, I read that if you are using a incomplete folder, content path will be the temp folder prior to the files reaching their final destination - which is the save_path property- so I'm curious what the result is if the files are moving (or pre finished moving?) and whether that is considered "complete" but the files are moving and what the content_path actually would be in transit?

src/clients/QBittorrent.ts
```diff
@@ -244,39 +240,53 @@ export default class QBittorrent implements TorrentClient {
...
 		}
-		return { save_path };
+		return torrentInfo.save_path;
+	}
+
+	/*
+	@return array of query results
+	 */
+	async getAllTorrentInfo(): Promise<TorrentInfo[]> {
```
zakkarry, line 258: Can we leverage some of this in the future for #471 ?

The function names seem to imply some of that functionality would be available?

src/clients/QBittorrent.ts
```diff
@@ -244,39 +240,53 @@ export default class QBittorrent implements TorrentClient {
...
+
+	/*
+	@param hash the hash of the torrent or undefined for all torrents
+	@return array of query results
+	 */
+	async getTorrentInfo(hash: string): Promise<TorrentInfo[]> {
+		const torrents = await this.getAllTorrentInfo();
+		return torrents.filter(
+			(torrent) =>
+				hash === torrent.infohash_v1 || hash === torrent.infohash_v2,
```
zakkarry: Is this just to future proof or what? I'm not entirely sure we have anything for v2 that would necessitate this?

src/clients/QBittorrent.ts
```diff
@@ -244,39 +240,55 @@ export default class QBittorrent implements TorrentClient {
...
+	@param hash the hash of the torrent or undefined for all torrents
+	@return array of query results
+	 */
+	async getTorrentInfo(hash: string): Promise<TorrentInfo[]> {
+		const torrents = await this.getAllTorrentInfo();
+		return torrents.filter(
+			(torrent) =>
+				hash === torrent.hash ||
+				hash === torrent.infohash_v1 ||
+				hash === torrent.infohash_v2,
```
zakkarry, line 273: is v2 infohash ever the same as v1?

our searchee's don't presently support v2 and it seems a long way off from adoption that would be relevant to us let?

src/clients/QBittorrent.ts
```diff
@@ -244,39 +240,55 @@ export default class QBittorrent implements TorrentClient {
...
-			"/torrents/info",
-			`hashes=${searchee.infoHash}`,
-			X_WWW_FORM_URLENCODED,
-		);
-		const searchResult = JSON.parse(responseText).find(
-			(e) => e.hash === searchee.infoHash,
-		) as TorrentInfo;
-		if (searchResult === undefined) {
+		const searchResult = await this.getTorrentInfo(searchee.infoHash);
+		if (searchResult.length === 0) {
```
zakkarry, line 285: is this undefined?

src/clients/QBittorrent.ts
```diff
@@ -285,16 +297,13 @@ export default class QBittorrent implements TorrentClient {
...
-		const files: TorrentFiles[] = JSON.parse(response);
-		const [{ name }] = files;
-		return files.length === 1 && posix.dirname(name) !== ".";
+	async isSubfolderContentLayout(
+		searchee: Searchee,
+		torrentInfo: TorrentInfo,
+	): Promise<boolean> {
+		if (searchee.files.length > 1) return false;
+		if (dirname(searchee.files[0].path) !== ".") return false;
+		return dirname(torrentInfo.content_path) !== torrentInfo.save_path;
```
zakkarry, line 306: same comment from the above about content path while files are in the "incomplete" directory or being moved FROM the incomplete directory?

zakkarry (commented):
just some thoughts on this new flow, some to get an idea of how things have changed and others just to verify understanding of things in the state they are when this PR would be merged I guess.

src/clients/QBittorrent.ts
zakkarry: Possibly irrelevant depending on when the actual completion callback is made from libtorrent, just a thought I had so i figured I'd comment to discuss since this changes things if not handled the way we are assuming.

ShanaryS:
> I'm confused, from the testers they said subfolder was handled after the change I made? was there some problem we didn't find?

I found problems with the old one myself. I needed to get all download dirs and it kept getting wrong paths until these changes. I also found the implementation in qbit: https://github.com/qbittorrent/qBittorrent/blob/321d7e5b17b6e4cacf713b4f4409364a5e65873e/src/gui/addnewtorrentdialog.cpp#L213

I also think this discord issue could have been caused by it: https://discord.com/channels/764440599066574859/1231604119593156672/1233189222102601821

Previously it wasn't properly account for single fine torrents that naturally had a folder.

src/clients/QBittorrent.ts
ShanaryS: This is calling the same /torrents/info endpoint but it gets all torrents in client. It was done this way as since we are only calculating v1 hash, if a torrent has v2 it won't return anything. So you get all torrents and check all the different hashes on it for a match.

As for that feature. I can look into it.

src/clients/QBittorrent.ts
ShanaryS: No it's needed right now. If a torrent has a v2 hash, it will truncate it to the first 40 characters (to match v1) and use that as the "hash" that gets checked for api calls and in BT_backup. But all torrents have a "hash" (always 40 characters), "infohash_v1" (naturally 40 characters) and "infohash_v2" (this is the full v2 hash, greater than 40 characters)".

I don't know how to allow the other endpoints to correctly get v2 torrents since they usually require a hash. Would probably get fixed by #471

src/clients/QBittorrent.ts
ShanaryS: No I changed it to require SearcheeWithInfoHash. Either way this returns an array of results so it should be handled gracefully.

src/clients/QBittorrent.ts
zakkarry: We check `progress` which I'm unsure of it's value considering the torrent's percent complete would be 100% `( progress === 1)` despite the "downloading/moving" operation being incomplete. There might need to be further checks done to validate isComplete given the operations taking place in this manner.

src/clients/QBittorrent.ts
zakkarry: Let's continue this conversation on [this thread/review comment](https://github.com/cross-seed/cross-seed/pull/665#discussion_r1581031095)

zakkarry:
> > I'm confused, from the testers they said subfolder was handled after the change I made? was there some problem we didn't find?
> 
> I found problems with the old one myself. I needed to get all download dirs and it kept getting wrong paths until these changes. I also found the implementation in qbit: https://github.com/qbittorrent/qBittorrent/blob/321d7e5b17b6e4cacf713b4f4409364a5e65873e/src/gui/addnewtorrentdialog.cpp#L213
> 
> I also think this discord issue could have been caused by it: https://discord.com/channels/764440599066574859/1231604119593156672/1233189222102601821
> 
> Previously it wasn't properly account for single file torrents that naturally had a folder.

I thought about that scenario when reviewing, the methodology seems sound now with the exception of [this thread/review comment](https://github.com/cross-seed/cross-seed/pull/665#discussion_r1581031095)

zakkarry:
> > I'm confused, from the testers they said subfolder was handled after the change I made? was there some problem we didn't find?
> 
> I found problems with the old one myself. I needed to get all download dirs and it kept getting wrong paths until these changes. I also found the implementation in qbit: https://github.com/qbittorrent/qBittorrent/blob/321d7e5b17b6e4cacf713b4f4409364a5e65873e/src/gui/addnewtorrentdialog.cpp#L213
> 
> I also think this discord issue could have been caused by it: https://discord.com/channels/764440599066574859/1231604119593156672/1233189222102601821
> 
> Previously it wasn't properly account for single file torrents that naturally had a folder.

I thought about that scenario when reviewing, the methodology seems sound now except [this thread/review comment](https://github.com/cross-seed/cross-seed/pull/665#discussion_r1581031095) needing some investigation.

src/clients/QBittorrent.ts
ShanaryS: We could also check the status for moving/rechecking or any other scenario as a part of that completeness check. Or just require it to be `uploading/pausedUP/queuedUP/stalledUP/forcedUP`

src/clients/QBittorrent.ts
```diff
@@ -244,57 +259,78 @@ export default class QBittorrent implements TorrentClient {
...
 
-		const { progress, save_path, auto_tmm, category } = searchResult;
+		const { save_path, state, auto_tmm, category } = searchResult[0];
+		const isComplete = [
+			"uploading",
+			"pausedUP",
+			"queuedUP",
+			"stalledUP",
+			"checkingUP",
+			"forcedUP",
```
bakerboy448, line 317: Proactively cover v5?

In qBittorrent 5.0, pausedUP and pausedDL have been renamed to stoppedUP and stoppedDL.

ShanaryS: I didn't know that. Will fix.

ShanaryS:
Additions:
1. Support v2 hash in all requests. This is done by getting torrents by `getTorrentsInfo()` to ensure we check all the hashes
2. Retrying failed requests. Sometimes `fetch()` would fail and throw an error, this will now retry for all requests.
3. Using state to check if completed. This better models what we consider "complete". I used "checkingUP" as complete, thoughts?
4. skipRecheck wasn't used properly for rechecking.

ShanaryS:
I'll add this here for future reference. Currently only "Original" and "Create Subfolder" layouts are supported. The "Remove Subfolder" is relatively new and is not planned right now. It would require a pretty decent refactor of how getDownloadDir works (or at least used) as well as how files are linked. We would basically have to completely ignore the file structure provided in the .torrent file.

Qbit's implementation of all options can be found here in `applyContentLayout`: https://github.com/qbittorrent/qBittorrent/blob/321d7e5b17b6e4cacf713b4f4409364a5e65873e/src/gui/addnewtorrentdialog.cpp#L194

This problem would be easier to solve with #471

zakkarry:
> I'll add this here for future reference. Currently only "Original" and "Create Subfolder" layouts are supported. The "Remove Subfolder" is relatively new and is not planned right now. It would require a pretty decent refactor of how getDownloadDir works (or at least used) as well as how files are linked. We would basically have to completely ignore the file structure provided in the .torrent file.
> 
> Qbit's implementation of all options can be found here in `applyContentLayout`: https://github.com/qbittorrent/qBittorrent/blob/321d7e5b17b6e4cacf713b4f4409364a5e65873e/src/gui/addnewtorrentdialog.cpp#L194
> 
> This problem would be easier to solve with #471

Pretty sure given the massive changes to 5 they are making that this is going to be something of a headache.

What I don't get is why they would change property/key names in an API that are linked to the same thing.

All this does is break shit.

PR #666, opened 2024-04-26
feat(decide): adds subsequent releases by same group filtering

zakkarry:
if the release group does a repack or propers, this release is certainly not the same as the previous and not cross-seedable.

this filters (last in the chain of filtering) whether the searchee and candidate are of the same "type" - repack and proper as well as repack# are supported.

considered adding internal, but not all trackers tag internals as such and could result in false-negatives

risky/partial will still operate on the basis of if there is a proper or repack in the name it returns true regardless of if they match to work with their aggregating of REPACK and PROPER into a PROPER.

if searchee matches arr's naming then it will act based on risky/partial being set.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -18,8 +18,9 @@ export const ANIME_REGEX =
 	/^(?:\[(?<group>.*?)\][_.\s-]?)?(?:\[?(?<title>.+?)[_.\s-]?(?:\(?(?:\d{1,2}(?:st|nd|rd|th))?\s?Season)?[_.\s-]?\]?)(?:[([~/|-]\s?(?!\d{4})(?<altTitle>.+?)[)\]~-]?\s?)?[_.\s-]?(?:\(?(?<year>(?:19|20)\d{2})\)?)?[[_.\s-](?:S\d{1,2})?[_.\s-]{0,3}(?:#|EP?|(?:SP))?[_.\s-]{0,3}(?<release>\d{1,4})(?:v\d)?[\])_.\s-]/i;
 export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w ]+?)(?:\[.+\])?(?:\))?(?=(?:\.\w{1,5})?$)/i;
+
 export const RP_REGEX =
-	/\b(?<type>(REPACK|PROPER)\d?)|(?<arrtype>(Proper)\d?)\b/;
+	/(\b(?<type>(REPACK|PROPER|\d\v\d)\d?))|(?<arrtype>(Proper|v\d))\b/;
 
 export const VIDEO_EXTENSIONS = [".mkv", ".mp4", ".avi", ".ts"];
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -18,7 +18,8 @@ export const ANIME_REGEX =
 	/^(?:\[(?<group>.*?)\][_.\s-]?)?(?:\[?(?<title>.+?)[_.\s-]?(?:\(?(?:\d{1,2}(?:st|nd|rd|th))?\s?Season)?[_.\s-]?\]?)(?:[([~/|-]\s?(?!\d{4})(?<altTitle>.+?)[)\]~-]?\s?)?[_.\s-]?(?:\(?(?<year>(?:19|20)\d{2})\)?)?[[_.\s-](?:S\d{1,2})?[_.\s-]{0,3}(?:#|EP?|(?:SP))?[_.\s-]{0,3}(?<release>\d{1,4})(?:v\d)?[\])_.\s-]/i;
 export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w ]+?)(?:\[.+\])?(?:\))?(?=(?:\.\w{1,5})?$)/i;
-export const RP_REGEX = /\b(?<type>(REPACK|PROPER)\d?)\b/i;
+export const RP_REGEX =
+	/\b(?<type>(REPACK|PROPER)\d?)|(?<arrtype>(Proper)\d?)\b/;
 
 export const VIDEO_EXTENSIONS = [".mkv", ".mp4", ".avi", ".ts"];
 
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -18,6 +18,7 @@ export const ANIME_REGEX =
 	/^(?:\[(?<group>.*?)\][_.\s-]?)?(?:\[?(?<title>.+?)[_.\s-]?(?:\(?(?:\d{1,2}(?:st|nd|rd|th))?\s?Season)?[_.\s-]?\]?)(?:[([~/|-]\s?(?!\d{4})(?<altTitle>.+?)[)\]~-]?\s?)?[_.\s-]?(?:\(?(?<year>(?:19|20)\d{2})\)?)?[[_.\s-](?:S\d{1,2})?[_.\s-]{0,3}(?:#|EP?|(?:SP))?[_.\s-]{0,3}(?<release>\d{1,4})(?:v\d)?[\])_.\s-]/i;
 export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w ]+?)(?:\[.+\])?(?:\))?(?=(?:\.\w{1,5})?$)/i;
+export const RP_REGEX = /\b(?<type>(REPACK|PROPER)\d?)\b/i;
```
ShanaryS, line 21:
```suggestion
export const RP_REGEX = /(?:\b|\d)(?<type>(REPACK\d?|PROPER\d?|v\d))\b/i;
```
I think we should also support anime style of vX. This is sometimes vX.Y but vX should be enough. Preceding character can be digit like `S01E05v2`.

src/decide.ts
```diff
@@ -169,7 +175,13 @@ function sizeDoesMatch(resultSize, searchee) {
 	const upperBound = length + fuzzySizeThreshold * length;
 	return resultSize >= lowerBound && resultSize <= upperBound;
 }
+function isRepackOrProper(searcheeName: string, candidateName: string) {
+	console.log(
+		`${searcheeName.match(RP_REGEX)} ${candidateName.match(RP_REGEX)}`,
+	);
 
+	return searcheeName.match(RP_REGEX) === candidateName.match(RP_REGEX);
```
ShanaryS, line 183: The name of the function does not match this check. Function should probably be renamed to something like `releaseVersionDoesMatch`

src/decide.ts
```diff
@@ -212,7 +224,9 @@ async function assessCandidateHelper(
 	if (!releaseGroupDoesMatch(searchee.name, name, matchMode)) {
 		return { decision: Decision.RELEASE_GROUP_MISMATCH };
 	}
-
+	if (!isRepackOrProper(searchee.name, name)) {
```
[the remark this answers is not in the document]
zakkarry: I'm adding the same condition for release group...so it would be fine.

src/decide.ts
```diff
@@ -169,7 +175,13 @@ function sizeDoesMatch(resultSize, searchee) {
 	const upperBound = length + fuzzySizeThreshold * length;
 	return resultSize >= lowerBound && resultSize <= upperBound;
 }
+function isRepackOrProper(searcheeName: string, candidateName: string) {
+	console.log(
```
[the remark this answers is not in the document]
zakkarry: forgot to control S ont he changes, and didn't double check that this was not updated to local IDE

src/constants.ts
zakkarry: Did you test this? it looks fine to me but we'd want to cover conservatively rather than too much which could lead to false-negatives.

zakkarry:
give me abit after i force push next there will be some new things to account for in it.

src/constants.ts
zakkarry: Let me know if this works...
```ts
export const RP_REGEX =
	/(\b(?<type>(REPACK|PROPER)\d?))|(?<arrtype>(Proper|v\d)\d?)\b/;
```
Given the context of the logic changes I've made since you reviewed....

zakkarry: Like this?

```ts
export const RP_REGEX =
	/(\b(?<type>(REPACK|PROPER|\d\v\d)\d?))|(?<arrtype>(Proper|\sv\d)\d?)\b/;```

check within the context of the latest push

File: src/decide.ts [hunk 5]
```diff
@@ -169,7 +175,24 @@ function sizeDoesMatch(resultSize, searchee) {
...
+	const searcheeVersionType = searcheeName.match(RP_REGEX);
+	const candidateVersionType = candidateName.match(RP_REGEX);
+	//sets either match arrtype or type to corresponding VersionType
 
+	if (
+		searcheeVersionType?.[0].toLowerCase() !==
+			candidateVersionType?.[0].toLowerCase() ||
+		searcheeVersionType?.groups?.arrtype
+	) {
+		return matchMode !== MatchMode.SAFE;
```
ShanaryS, line 192: This seems to ignore the check when not using `safe`. Also seems to reject release that maches `arrtype`.

zakkarry: if arrtype is true, it means that it is defaulted away from PROPER and REPACK and aggregated into Proper...I see very very few releases that use Proper vs all caps. We could also match if arrtype and candidate is proper or repack instead, but if arrtype exists it necessitate risky - because its obviously a renamed file.

for risky concern, maybe we should add something to the logic on the return ? I only kind of understand what you are meaning...it seems like you think their should be more or a different value returned? Maybe I'm missing something.

ShanaryS: It seems if matchMode is risky or partial, then this check won't apply at all. Which is fine just clarifying.

I don't see why we should check for arrtype. Since it's case sensitive won't it be rejected anyways?

```ts
export const RP_REGEX =
	/(?:\b|\d)(?<type>(?:REPACK\d?|PROPER\d?|v\d))\b/;
```
https://regex101.com/r/qSkVuj/1

The `\b` covers `\s`. Also can't have both `\b` and `\d` before `vX` so `\b|\d` is required.

src/constants.ts
ShanaryS: Moved to: https://github.com/cross-seed/cross-seed/pull/666#discussion_r1581380270

src/decide.ts
zakkarry: because arrs label repacks and proper as just proper universally

zakkarry: I guess also checking if candidateVersionType exists at all would be smart.

zakkarry:
```ts
	if (
		searcheeVersionType?.[0].toLowerCase() !==
			candidateVersionType?.[0].toLowerCase() ||
		(searcheeVersionType?.groups?.arrtype && candidateVersionType)
	) {
		return matchMode !== MatchMode.SAFE;
	}
```

zakkarry: https://regex101.com/r/6kE85o/1


test cases for all variants

PR #668, opened 2024-04-27
feat(match): ensemble matching for episodes

ShanaryS:
This feature allows grabbing rss/announce and searching of seasons from individual episodes. Along with the changes in v6, this feature completes all possible matches from cross-seed aside from tweaks.

This works by introducing a new type of searchee, a virtual searchee. This is needed since the episodes are sourced from multiple torrents and folders. As such, it hash no `infoHash` or `path`. All areas of the application where these were used for logic has been updated to reflect this new possiblity. In general, this should be treated as a data based searchee if searchee.path is not need. So most checks went from `if (searchee.path)` to `if (!searchee.infoHash)` so that virtual searchees are included.

Virtual searchees also have absolute file paths instead of relative to make linking easier. This required some small changes when linking as well as changes in perfect matches. For a perfect match, when the item is a virtual searchee it only requires a name match on each file rather than path.

To support rss/announce, the information needed to be cached. I opted to use an in memory database as I did not want to add to the existing which would need migrations. For 2k entries, it takes about ~3s on startup and about 1MB of memory.

This feature works best with search. Due to the nature of the problem, it's easier to find matches by searching rather than from rss/announce, especially for anime.

Also, I added a check to see if the size changed for data based searches and report `TORRENT_NOT_COMPLETE` if it hash. A similar check was done for ensemble but on each file.

---

The implementation is to parse titles and put them in a standard format `cleanTitle.season.resolution.source-RlsGrp` called key. As such two new regexs were added for resolution and source. These are necessary to restrict searching for mixed resolutions and sources. Once there is more than 2 distinct episodes that all have the same key, it's eligible to be searched. The searchee name will be the key while the files will be the absolute path to the largest file in each episode torrent.

Some additional checks are put into place. The highest episode number is used as the total number in the season. We then check for the presence of all episodes down to E01 and if more than the `seasonFromEpisodes` is missing, it won't be created. It also checks to ensure that the file with the newest modified time is at least 8 days old since most episodes are released weekly. Of course this will search if there is a mid season break and get no results but the alternative would be to use a sonarr (likely will later). To account for this, if the age of the newest episode is after its last search time, we will search it again immediately (assuming it's more than 8 days old).

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -34,11 +34,15 @@ const ZodErrorMessages = {
 		"excludeRecentSearch must be at least 3x searchCadence.",
 	excludeRecentOlder:
 		"excludeOlder and excludeRecentSearch must be defined for searching. excludeOlder must be 2-5x excludeRecentSearch.",
-	fuzzySizeThreshold:
-		"fuzzySizeThreshold must be between 0 and 1 with a maximum of 0.1 when using searchCadence or rssCadence",
 	injectNeedsInjectMode: "`cross-seed inject` requires the 'inject' action.",
 	autoResumeMaxDownloadUnsupported:
 		"autoResumeMaxDownload must be an integer of bytes between between 0 and 52428800 (50 MiB).",
+	numberMustBeRatio:
+		"fuzzySizeThreshold and seasonFromEpisodes must be between 0 and 1.",
+	fuzzySizeThresholdMax:
+		"fuzzySizeThreshold cannot be greater than 0.1 when using searchCadence or rssCadence.",
+	seasonFromEpisodesMin:
+		"seasonFromEpisodes cannot be less than 0.5 when using searchCadence",
 	injectUrl:
 		"You need to specify rtorrentRpcUrl, transmissionRpcUrl, qbittorrentUrl, or delugeRpcUrl when using 'inject'",
 	qBitAutoTMM:
@@ -49,9 +53,10 @@ const ZodErrorMessages = {
 		"outputDir should only contain .torrent files, cross-seed will populate and manage (https://www.cross-seed.org/docs/basics/options#outputdir)",
 	needsTorrentDir:
 		"You need to set torrentDir for rss and announce matching to work.",
-	needsInject: "You need to use the 'inject' action for partial matching.",
+	needsInject:
+		"You need to use the 'inject' action for partial matching or seasonFromEpisodes.",
 	needsLinkDir:
-		"You need to set a linkDir (and have your data accessible) for risky or partial matching to work.",
+		"You need to set a linkDir (and have your data accessible) for risky/partial matching and seasonFromEpisodes to work.",
 	linkDirInOtherDirs:
 		"You cannot have your linkDir inside of your torrentDir/dataDirs/outputDir. Please adjust your paths to correct this.",
 	dataDirsInOtherDirs:
@@ -62,6 +67,8 @@ const ZodErrorMessages = {
 		"You cannot have your outputDir inside of your torrentDir/dataDirs/linkDir. Please adjust your paths to correct this.",
 	relativePaths:
 		"Absolute paths for torrentDir, linkDir, dataDirs, and outputDir are recommended.",
+	needsPartial:
+		"seasonFromEpisodes requires matchMode partial if enabled and value is below 1.",
 };
 
 /**
@@ -267,7 +274,12 @@ export const VALIDATION_SCHEMA = z
 		fuzzySizeThreshold: z
 			.number()
 			.positive()
-			.lte(1, ZodErrorMessages.fuzzySizeThreshold),
+			.lte(1, ZodErrorMessages.numberMustBeRatio),
+		seasonFromEpisodes: z
+			.number()
+			.positive()
+			.lte(1, ZodErrorMessages.numberMustBeRatio)
+			.nullish(),
 		excludeOlder: z
 			.string()
 			.min(1, ZodErrorMessages.emptyString)
@@ -377,7 +389,14 @@ export const VALIDATION_SCHEMA = z
 		(config) =>
 			config.fuzzySizeThreshold <= 0.1 ||
 			(!config.searchCadence && !config.rssCadence),
-		ZodErrorMessages.fuzzySizeThreshold,
+		ZodErrorMessages.fuzzySizeThresholdMax,
+	)
+	.refine(
+		(config) =>
+			(!config.searchCadence && !config.rssCadence) ||
+			!config.seasonFromEpisodes ||
+			config.seasonFromEpisodes >= 0.5,
+		ZodErrorMessages.seasonFromEpisodesMin,
 	)
 	.refine((config) => {
 		if (
@@ -415,7 +434,8 @@ export const VALIDATION_SCHEMA = z
 		(config) =>
 			process.env.DEV ||
 			config.action === Action.INJECT ||
-			config.matchMode !== MatchMode.PARTIAL,
+			(config.matchMode !== MatchMode.PARTIAL &&
+				!config.seasonFromEpisodes),
 		ZodErrorMessages.needsInject,
 	)
 	.refine(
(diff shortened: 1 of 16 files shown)


src/configSchema.ts
```diff
@@ -222,6 +223,9 @@ export const VALIDATION_SCHEMA = z
 		return true;
 	})
 	.refine(
-		(config) => config.matchMode === MatchMode.SAFE || config.linkDir,
+		(config) =>
+			(config.matchMode === MatchMode.SAFE &&
+				!config.seasonFromEpisodes) ||
+			config.linkDir,
```
zakkarry, line 229:
```suggestion
		(config) =>
			(config.matchMode === MatchMode.SAFE) &&
				(config.seasonFromEpisodes &&
			!config.linkDir,
```
Can you explain this for me, the logic isn't readily apparent to me on why this isn't grouped differently....

ShanaryS: This says that linkDir is optional if you are in matchMode safe and seasonFromEpisodes is false.

So basically you need linkDir if you aren't on safe OR you have this feature enabled. The part about matchMode is not for this feature but for risky/partial needing linkDir.

src/clients/Deluge.ts
```diff
@@ -375,32 +375,51 @@ export default class Deluge implements TorrentClient {
...
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		data: SearcheeWithInfoHash[] | Metafile[],
+	): Promise<Map<string, string>> {
+		const response = await this.call<TorrentStatus>("web.update_ui", [
+			["save_path"],
+			{},
+		]);
+		const dirs = new Map<string, string>();
+		if (response.result?.torrents) {
+			for (const [hash, torrent] of Object.entries(
```
zakkarry, line 414: verified that this parses properly, so 2 clients out of 4 checked

zakkarry:
I'm curious to have the discussion of making partial/ensemble a matchMode like "aggressive" or something....rather than an additional option or potentially a 4th matchMode or something.

We typically are pretty conservative about adding options and its usually only when very specific functionality is added, and I think this could be boiled down to a matchMode.

ShanaryS:
> I'm curious to have the discussion of making partial/ensemble a matchMode like "aggressive" or something

Yeah that would work. But people may want ensemble without partial since it downloads. I would rather just always have it on... Is there a reason anyone wouldn't want ensemble? It's no different than just matching a season with another.

src/utils.ts
```diff
@@ -362,14 +365,18 @@ export function extractInt(str: string): number {
 	return parseInt(str.match(/\d+/)![0]);
 }
 
-export function getFuzzySizeFactor(): number {
-	const { fuzzySizeThreshold } = getRuntimeConfig();
-	return fuzzySizeThreshold;
+export function getFuzzySizeFactor(searchee: Searchee): number {
```
mmgoodnow, line 368: does this obviate fuzzySizeThreshold when you have seasonFromEpisodes turned on? is there a way to consolidate the options?

src/searchee.ts
```diff
@@ -95,19 +104,41 @@ export function getLargestFile(files: File[]): File {
...
+	searchee: SearcheeWithoutInfoHash,
+): Promise<number> {
+	if (!searchee.path) {
+		return getNewestFileAge(searchee.files.map((file) => file.path));
+	}
+	const pathStat = statSync(searchee.path);
+	if (pathStat.isFile()) return pathStat.mtimeMs;
+	return getNewestFileAge(
+		searchee.files.map((file) =>
+			getAbsoluteFilePath(searchee.path!, file.path, false),
```
mmgoodnow, line 137: TS would ordinarily narrow this to non-undefined automatically, but I think it's worried that mutation could remove `path` before the lambda gets called. 

You should be able to trick it into understanding that searchee.path is defined by rebinding `path` above the `.map` call with `const { path } = searchee;`

src/searchee.ts
```diff
@@ -360,3 +394,251 @@ export function getReleaseGroup(stem: string): string | null {
 		? null
 		: parsedGroupMatchString;
 }
+
+const logReason = (reason: string, options: { useFilters: boolean }): void => {
```
mmgoodnow, line 398: logReason could probably use a rename here

src/pipeline.ts
```diff
@@ -335,28 +339,108 @@ export async function checkNewCandidateMatch(
...
+	candidate: Candidate,
+	searcheeLabel: SearcheeLabel,
+): Promise<{
+	decision: DecisionAnyMatch | Decision.INFO_HASH_ALREADY_EXISTS | null;
+	actionResult: ActionResult | null;
+}> {
+	const searchees: SearcheeWithLabel[] = [];
+
+	const methods = [
+		await pushSearcheesForCandidate(candidate, searchees, searcheeLabel),
```
mmgoodnow, line 436: is there a hard reason we need to push into this array of searchees rather than having these functions return arrays themselves and concatenating them together into a new array? I find it obfuscates the data flow especially when the creation and mutation are in separate functions.

src/pipeline.ts
```diff
@@ -335,28 +339,108 @@ export async function checkNewCandidateMatch(
...
+	if (files.length === 0) {
+		logger.verbose({
+			label: searcheeLabel,
+			message: `Did not find any files for ensemble ${ensembleTitle} for ${candidateLog}: sources may be incomplete or missing`,
+		});
+		return null;
 	}
+	// Get searchee.length by total size of each episode (average if multiple files for episode)
+	const uniqueElements = new Set(ensemble.map((e) => e.element));
+	const totalLength = [...uniqueElements].reduce((acc, cur) => {
```
mmgoodnow, line 401: should we even bother with length checks for ensemble searchees? guessing the number of searchees feels like an easy way to get false negatives. I guess a lot of the time you'll have all of the episode files but you might also have all but one episode, and the finale might be twice as long. Would source/resolution and release group be enough?

src/searchee.ts
```diff
@@ -360,3 +394,251 @@ export function getReleaseGroup(stem: string): string | null {
...
+	episode: number | string,
+	existingSeasonMap: Map<string, SearcheeWithLabel[]>,
+	keyMap: Map<string, Map<number | string, SearcheeWithLabel[]>>,
+	ensembleTitleMap: Map<string, string>,
+): void {
+	for (let i = 0; i < keys.length; i++) {
+		const key = keys[i];
+		if (existingSeasonMap.has(key)) continue;
+		if (!ensembleTitleMap.has(key)) {
+			ensembleTitleMap.set(key, ensembleTitles[i]);
```
mmgoodnow, line 419: I noticed the same obfuscated mutation here, although it looks like this one's entire purpose is to share the mutation code across anime/standard so maybe it's less avoidable. Renaming it might help

mmgoodnow (commented):
I like it! I appreciate how you've broken off individually beneficial chunks from this PR over time to deliver it incrementally. Now that I've read it, I've realized how many of these recent features originally came from the ensemble matching work stream, and I'm glad they are all able to work together nicely.

src/pipeline.ts
ShanaryS: I think it's still helpful to reduce snatches. Since we are checking source, resolution, and release group, the only differences between episodes would be a couple bytes (so corrupted cross seeds or repacks). So the total size should still be close if the user has all.

If they are missing some, it's tricky but the lower limit for `seasonFromEpisodes` is `0.5` so there it should get caught. Different episode lengths and other things can reduce this effectiveness which can only be solved by sonarr. But I haven't seen any problems manifest and can always tweak it later on.

src/pipeline.ts
ShanaryS: I needed to return the string for logging, so I felt pushing to the array makes it easier. The alternative would be a more compilated return type.

src/searchee.ts
ShanaryS: There is a lot of mutation in this pipeline. I find it difficult to organize without it as the return types and reassignments would be a lot.

src/utils.ts
ShanaryS: Yes, `seasonFromEpisodes` completely replaces `fuzzySizeThreshold` for ensemble. `getFuzzySizeFactor(searchee)` and `getMinSizeRatio(searchee)` are used everywhere to account for this.

src/searchee.ts
ShanaryS: I prefer using ! as it a no-op.

mmgoodnow: Using ! puts the onus on the programmer to validate that they know better than the computer. We should rely on the computer wherever we can, and a variable binding is basically 0 cost.

src/pipeline.ts
mmgoodnow: Let's do the more complicated return type.

src/utils.ts
mmgoodnow: Hmm, I haven't really given it much thought but if you think the config interface makes the most sense like this I trust you.

PR #670, opened 2024-04-30
feat(action/inject): add inject job to retry saved torrents and more

ShanaryS:
This feature allows cross-seed to try to inject torrents it finds in outputDir. This means if something goes wrong and a match falls back to saving, it will later be injected. The job is set to run hourly and is instant when outputDir is empty. This can also be triggered by a new CLI option `cross-seed inject`. This uses all the settings from the user's config.js but if they want to inject torrents from another directory they can use `cross-seed inject --inject-dir <dir_name>`.

This also works for torrents never seen by cross-seed, so a user could add a torrent they got elsewhere and cross-seed will still inject it provided it matched. They should follow the cross-seed format of `[movie][Tracker]whatever.torrent`. This is because we use the tracker when linking. If the format is not valid, will link to `linkDir/UnknownTracker` if flatLinking is false.

This leverages existing functions so it shouldn't need to be updated if improvements or features or added to cross-seed. It essentially goes straight to the decide stage then action, never interacting with the database. I also made it so it did no filtering. Since this won't be requesting indexers, at the searchee creation stage and decide stage, only the necessary checks are performed. `MATCH_PARTIAL` decisions are ordered by torrent/data based over ensemble, then by most files to maximize the partial percentage.

Since there is no filtering, it's possible for false positives to happen. This won't apply to cross-seed retrying failed injections, as those already went through the full pipeline, it's only possible if the user supplied their own torrent to test. There is a warning on the help for `cross-seed inject` and a warn message each time it's used. It's too much detail to go into here, but I don't think any user will come across a false positive because of this, the one I saw is an edge case of an edge case. The no filtering does allow extra matches that cross-seed otherwise would of rejected. Biggest being two season packs that has a different number of episodes can be cross-seeded since we aren't checking total size fuzzy. Essentially if you think something is a cross-seed, `cross-seed inject` will definitively say so. We can use this fact for support or future development.

## Behavior Changes

Will now update database when if a `TORRENT_NOT_COMPLETE` injection result is found. Previously this wasn't done so that it will get retried next search but this is no longer necessary.

Updated the risky and partial algorithms to uniquely use a searchee file. So that multiple candidate files can't match with the same searchee file. I knew about this when first implementing partial, but it's only necessary now with the inject feature combined with the `ensemble-matching` branch.

Now returning 200 for failed injections from announce since we will be retrying those with the inject job. I'm also returning 200 if it already exists. This allows autobrr to mark an announce as successful in case rss or a webhook search injected between autobrr retries. Updated to logging to state `Injected`, `Saved`, `Already Exists` instead of `Added` along with status code.

Added a new decision `SAME_INFO_HASH`. In a few scenarios with the inject feature, this needs to be handled separately from already exists. It also helps prevent incorrect 200 response for announces if radarr/sonarr added the torrent to the client before `cross-seed` processed the announce.

Saved .torrent files also include info_hash so torrents with identical names can be saved. This is necessary since we will be using `outputDir` heavily.

We will no longer fall back to non-linking if it fails, we just saved instead for later retrying. This only was possible in the case of a match and the source wasn't found, pretty much always a user error.

I changed all `MATCH_PARTIAL` to save even on successful injection. Also, `cross-seed inject` will not delete processed .torrent files unless they are in client and complete or a week old. This was done mainly in the case of `ensemble-matching` where a user could resume a single torrent and remove the rest once that's complete. Then `cross-seed inject` will automatically re-inject the others this time at 100% using the newly completed torrent. This not only allows users to download the missing data once, but it gets linked between all so no extra storage is used.

## Special Abilities

The inject feature also can link files from multiple searchees and trigger a recheck in client to automate the `MATCH_PARTIAL` case above. So a user would only need to resume a single partial match and the next time inject is ran it will dynamically link the new files for the other torrents (already exists in client, no re-injection) and performs a recheck. No other user user input required aside from choosing which tracker they want to incur the partial download on.

I also added the ability to inject torrents if the source is stalled. These will always be treated as partial, so paused and rechecked. If all the sources are `TORRENT_NOT_COMPLETE` and it hasn't been injected for a day, it will be added with whatever files are present if linking is enabled. This allows the user to choose to resume this "cross seed" so they can potentially revive the stalled torrent. Of course this is not really cross seeding, but it has the potential to revive dead torrents so I think it's worth having.


mmgoodnow:
Starting a review.

src/pipeline.ts
```diff
@@ -270,17 +282,20 @@ export async function searchForLocalTorrentByCriteria(
 
 export async function checkNewCandidateMatch(
 	candidate: Candidate,
+	candidateLog: string,
```
mmgoodnow, line 285: it feels weird to pass around a log string rather than richer, structured, information. Is this because `candidate.tracker` doesn't exist?

src/pipeline.ts
```diff
@@ -317,30 +334,46 @@ export async function checkNewCandidateMatch(
...
-			candidate.tracker,
-		);
+		actionResult = (
+			await performAction(
+				assessment.metafile!,
+				assessment.decision,
+				searchee,
+				candidate.tracker,
+			)
+		).actionResult;
```
mmgoodnow, line 358: there's syntax sugar for this, but it's also a little weird—you can decide whether you like it better than this or not. 
```suggestion
		({ actionResult } = await performAction(
				assessment.metafile!,
				assessment.decision,
				searchee,
				candidate.tracker,
			))
```

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+
+	if (injectDir) {
+		// injectDir defined only by `cross-seed inject`
+		logger.warn({
+			label: Label.INJECT,
+			message: `Manually injecting torrents performs minimal filtering which slightly increases chances of false positives, see the docs for more info`,
+		});
+	}
+	if (!fs.existsSync(targetDir) || !fs.statSync(targetDir).isDirectory()) {
+		logger.info({
```
mmgoodnow, line 550: is this an error rather than info?

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+	const targetDirLog = chalk.bold.magenta(targetDir);
+
+	if (injectDir) {
+		// injectDir defined only by `cross-seed inject`
+		logger.warn({
+			label: Label.INJECT,
+			message: `Manually injecting torrents performs minimal filtering which slightly increases chances of false positives, see the docs for more info`,
+		});
+	}
+	if (!fs.existsSync(targetDir) || !fs.statSync(targetDir).isDirectory()) {
```
mmgoodnow, line 549: it'd be nice to consolidate this into one fs call where we catch ENOENTs rather than 2 separate calls. 

It may seem like this code protects against `ENOENT`, but if the file gets deleted in between the `existsSync` call and the `statSync` call, the program throws an unexpected error. By "begging for forgiveness" (catching ENOENT errors from `statSync`) rather than "asking for permission" (preflight `existsSync` calls), we harden the program against race conditions.

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+		useFilters: false,
+	});
+
+	const toDelete = new Set<string>();
+	const toRecheck = new Set<string>();
+
+	// Usually source got deleted or partial injection never completes
+	function shouldCleanUpTorrent(torrentFilePath: string) {
+		return fs.statSync(torrentFilePath).mtimeMs < Date.now() - ms("1 week");
+	}
```
mmgoodnow, line 579: I'm wondering if cleaning up torrents would make more sense as a separate job. What do you think?

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+		}
+		fullMatches.sort((a, b) => {
+			// Prefer torrent over data/virtual
+			if (a[0].infoHash && !b[0].infoHash) {
+				return -1;
+			}
+			if (!a[0].infoHash && b[0].infoHash) {
+				return 1;
+			}
+			return 0; // Should keep MATCH first within a searchee type
```
mmgoodnow, line 650: nit: with tuples, I think it's easier to read when we use destructuring to access its members rather than [0] or [1]

```suggestion
		fullMatches.sort(([searcheeA], [searcheeB]) => {
			// Prefer torrent over data/virtual
			if (searcheeA.infoHash && !searcheeB.infoHash) {
				return -1;
			}
			if (!searcheeA.infoHash && searcheeB.infoHash) {
				return 1;
			}
			return 0; // Should keep MATCH first within a searchee type
```

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+			if (decision === Decision.MATCH) {
+				fullMatches.unshift([searchee, decision]);
+			} else if (decision === Decision.MATCH_SIZE_ONLY) {
+				fullMatches.push([searchee, decision]);
+			} else if (decision === Decision.MATCH_PARTIAL) {
+				partialMatches.push([searchee, decision]);
+			} else if (decision === Decision.BLOCKED_RELEASE) {
+				foundBlocked = true;
+			}
+		}
```
mmgoodnow, line 641: I'm curious—could we run these through `checkNewCandidateMatch` to figure out what searchee they belong to, rather than checking them against every searchee? Or is there some reason we wouldn't want to do that?

src/pipeline.ts
```diff
@@ -482,14 +523,400 @@ export async function scanRssFeeds() {
...
+				message: `(${i + 1}/${candidatesSinceLastTime.length}) ${candidateLog}`,
 			});
-			await checkNewCandidateMatch(candidate, Label.RSS);
+			await checkNewCandidateMatch(candidate, candidateLog, Label.RSS);
 		}
 		logger.info({ label: Label.RSS, message: "Scan complete" });
 	}
 }
+
+export async function injectSavedTorrents() {
```
mmgoodnow, line 537: this is a beast of a function—I think it'd be good to break up into smaller pieces. Perhaps in a different file as well, though I'm not sure what the good division lines would be.

mmgoodnow (commented):
Overall things look good but the `injectSavedTorrents` function is too long to easily understand. I think applying some separation of concerns principles will help.

src/pipeline.ts
ShanaryS: We can't use `checkNewCandidateMatch` as we need information on every matched searchee. We also want to check data based searchees as well. We could the reverse lookup process though with some tweaking.

I think checking against every searchee is easiest and most complete. It takes just a couple seconds to process each torrent file as it is.

The only negative with the current implementation is that it could link an unrelated torrent since we are only checking size in risky and partial match modes. I haven't seen any issues with this yet (and it's extremely unlikely), but I'll just verify each successful match with a levenshtein check on the titles.

mmgoodnow: Why do we need information on every matched searchee?

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+	logger.info({ label: Label.INJECT, message: chalk.cyan(resultMsg) });
+
+	if (totalUnmatched > 0) {
+		logger.info({
+			label: Label.INJECT,
+			message: `Use "${chalk.bold.white("cross-seed diff")}" to get the reasons two torrents are not considered matches`,
+		});
+	}
+
+	if (foundBadFormat && !flatLinking) {
```
mmgoodnow, line 1017: there's a `logOnce` helper https://github.com/cross-seed/cross-seed/blob/4cb1f50e75644dfaf9e99b48eab40e16c06a2996/src/logger.ts#L87 that might allow us to "fire and forget" and not worry about storing this information in `foundBadFormat`.

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+	const toRecheck = new Set<string>();
+
+	// Usually source got deleted or partial injection never completes
+	function shouldCleanUpTorrent(torrentFilePath: string) {
+		return fs.statSync(torrentFilePath).mtimeMs < Date.now() - ms("1 week");
+	}
+	function getTorrentFilePathLog(torrentFilePath: string) {
+		return chalk.bold.magenta(
+			torrentFilePath.replace(
+				/(?:\[)([a-z0-9]{40})(?:].torrent$)/i,
```
mmgoodnow, line 824: are these non-capturing groups doing anything?
```suggestion
				/\[([a-z0-9]{40})].torrent$/i,
```

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+				(match, hash) => match.replace(hash, sanitizeInfoHash(hash)),
+			),
+		);
+	}
+
+	let totalInjected = 0;
```
mmgoodnow, line 830: these totals are a bit verbose, thoughts on something like 

```js
const totals = {
    INJECTED: 0,
    MATCH: 0,
    MATCH_PARTIAL: 0,
    BLOCKED_RELEASE: 0,
    ...
};
```

and then we can just do `totals[decision]++` or `totals[actionResult]++` in the relevant places. This would make the summary log at the end easier to extract into its own function as well.

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+		) {
+			const metaSearcheeRes = createSearcheeFromMetafile(meta);
+			if (metaSearcheeRes.isErr()) continue;
+			const metaSearchee = metaSearcheeRes.unwrap();
+			if (
+				!areMediaTitlesSimilar(searchee.title, metaSearchee.title) &&
+				!areMediaTitlesSimilar(searchee.name, metaSearchee.title)
+			) {
+				continue;
+			}
```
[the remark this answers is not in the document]
ShanaryS: I'm now just parsing the title when creating every metafile.

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+			(searcheeB.infoHash || searcheeB.path)
+		) {
+			return 1;
+		}
+		return searcheeB.files.length - searcheeA.files.length; // Prefer more files
+	});
+	return { fullMatches, partialMatches, foundBlocked };
+}
+
+async function injectInitialAction(
```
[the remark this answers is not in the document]
ShanaryS: Keeping track of what was injected to act on later in the pipeline.

src/pipeline.ts
```diff
@@ -489,14 +541,507 @@ export async function scanRssFeeds() {
...
+		totalInjected++;
+		if (matchedDecision! === Decision.MATCH_PARTIAL) {
+			totalPartialMatches++;
+		} else {
+			totalFullMatches++;
+		}
+	}
+
+	for (const infoHash of toRecheck) {
+		await getClient().recheckTorrent(infoHash);
```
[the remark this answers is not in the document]
ShanaryS: This was done because of the `resumePartials` branch as we might want to resume without rechecking. But this will be handled differently now.

src/pipeline.ts
ShanaryS: I want these to be logged at the end with all the summary, it would cause this to be lost in the middle of the logs.

ShanaryS:
> Thinking big-picture, it'd be really convenient for at least some parts of the existing reverse lookup process to handle this case. It doesn't really seem very different than the reverse lookup case from my POV - we have a torrent and we want to inject it because we think it's a cross-seed but need to figure out what to match against.

> What is the difference?

The reverselookup for rss and announce and retrying injection has some big differences in what they have access to and goals. The inject feature has already "snatched" the candidates so we can do comparisons on each individual file as well. We are also using the decide process to filter out bad matches, rather than doing the lookup first (the decide process is pretty much sufficient). The lookup is much loser than for candidates since we already know these matches are good since they were saved.

It also adds more flexibility for users who want to manually inject something (think the `pollinizer` or `crops` applications for music). The goals of this injection feature is to source from all possible searchees and link dynamically. It needs to know a lot of details at every stage from lookup, decide, and action, and it never saves anything to the database.

So even though some logic is similar at points, there is too many small or big differences in the process. Combining them while maintaining each properly would only make things more complicated and monolithic, rather than keeping them separate.

> > We also want to check data based searchees as well.

> This is one of the last remaining differences between data and torrent backed searchees—announce doesn't look at data-based searchees. Maybe this is the time to support it in rss/announce (as long as it's not too slow).

I no longer planned on working on this which all the improvements to torrent based. Watching all the data dirs or indexing each rss/announce is slower and more intensive that for torrentDir, I couldn't find a good solution for it. Maybe it can be implemented later, but I think it should be after v6 releases.

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+					tracker,
+					injectionResult,
+					progress,
+					filePathLog,
+				);
+			}
+			summary.INCOMPLETE_SEARCHEES++;
+			continue;
+		}
+		const result = await getClient().isTorrentComplete(meta.infoHash);
```
mmgoodnow, line 959: is there a reason we can't trust the `injectionResult` on hand and have to call the client again to check for completion?

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+
+	if (summary.FOUND_BAD_FORMAT && !summary.FLAT_LINKING) {
+		logger.warn({
+			label: Label.INJECT,
+			message: `Some torrents could be linked to linkDir/${UNKNOWN_TRACKER} - follow .torrent naming format in the docs to avoid this`,
+		});
+	}
+}
+
+export async function injectSavedTorrents() {
```
mmgoodnow, line 818: Good work breaking down this function so far. I'm still finding it somewhat hard to read because it's 215 lines long, there are many variables in scope, and there's a fair amount of mutation. Let's try to get it under 100. 

I would start by extracting the body of the `for (const torrentFilePath of dirContents)` loop into its own function. (this has precedent, with `performAction`/`performActions` and `findOnOtherSites`/`findMatchesBatch`). 

From there, I think extracting the bodies of the `TORRENT_NOT_COMPLETE` and `ALREADY_EXISTS` branches in the aftermath of `injectInitialAction` would make sense. We can pass around the `summary` object if needed.

Let's put all of these functions (really, all of the functions added to `pipeline.ts` in this PR) into a new file, `injectSavedTorrents.ts` (or `inject.ts`, if you prefer.) These are pretty independent codepaths from the other code in `pipeline.ts`.

Removing the `toDelete` builder code should help as well.

src/utils.ts
```diff
@@ -218,11 +274,15 @@ export function getLogString(
 			: `${color(searchee.title)} ${chalk.dim(`[${searchee.name}]`)}`;
 }
 
-export function formatAsList(strings: string[]) {
+export function formatAsList(
+	strings: string[],
+	options: { sort: boolean; type?: Intl.ListFormatType },
```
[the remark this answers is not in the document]
ShanaryS: It says `A type literal property cannot have an initializer. ts(1247)`

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+		}
+		if (injectionResult === InjectionResult.ALREADY_EXISTS) {
+			if (linkedNewFiles) {
+				logger.info({
+					label: Label.INJECT,
+					message: `${progress} Rechecking ${filePathLog} as new files were linked - ${chalk.green(injectionResult)}`,
+				});
+				await getClient().recheckTorrent(meta.infoHash);
+			} else if (fullMatches.length && !isComplete) {
+				logger.info({
```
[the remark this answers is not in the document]
ShanaryS: These messages describe what's happening to the saved torrent so I think it's important to have in info, similar to searching and action logs. I also like the different logging as the behaviors after rechecking are different.

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+					filePathLog,
+				);
+			}
+			summary.INCOMPLETE_SEARCHEES++;
+			continue;
+		}
+		const result = await getClient().isTorrentComplete(meta.infoHash);
+		let isComplete = result.isOk() ? result.unwrap() : false;
+		if (isComplete) {
+			toDelete.add(torrentFilePath);
```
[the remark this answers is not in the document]
ShanaryS: I'll implement this if it's not `outputDir` but otherwise we should keep it. Torrents saved long ago or have it's sources removed before the job would stay around forever unless the user manually intervenes. After this is merged, a user should never interact with `outputDir`.

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+			`Injected ${chalk.bold.green(summary.INJECTED)}/${chalk.bold.white(summary.TOTAL)} torrents`,
+			summary.FULL_MATCHES &&
+				`${chalk.bold.green(summary.FULL_MATCHES)} were full matches`,
+			summary.PARTIAL_MATCHES &&
+				`${chalk.bold.yellow(summary.PARTIAL_MATCHES)} were partial matches`,
+			summary.INCOMPLETE_SEARCHEES &&
+				`${chalk.bold.yellow(summary.INCOMPLETE_SEARCHEES)} had incomplete sources`,
+			summary.ALREADY_EXISTS && incompleteMsg,
+			summary.BLOCKED &&
+				`${chalk.bold.yellow(summary.BLOCKED)} were possibly blocklisted`,
```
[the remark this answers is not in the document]
ShanaryS: Since it's checking all searchees, the blocklist may be a false positive. It shouldn't be a big deal though as it's only displayed if a match couldn't be found.

src/pipeline.ts
```diff
@@ -489,14 +551,483 @@ export async function scanRssFeeds() {
...
+			});
+			continue;
+		}
+
+		if (decision === Decision.MATCH) {
+			fullMatches.unshift([searchee, decision]);
+		} else if (decision === Decision.MATCH_SIZE_ONLY) {
+			fullMatches.push([searchee, decision]);
+		} else {
+			partialMatches.push([searchee, decision]);
```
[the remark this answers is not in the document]
ShanaryS: We need to check if there is a full match later on, but otherwise I've implemented the above.

src/pipeline.ts
ShanaryS: If the torrent already exists, it's used for logging and behavior.

src/pipeline.ts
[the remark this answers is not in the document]
mmgoodnow: > Torrents saved long ago or have its sources removed before the job would stay around forever unless the user manually intervenes. 

why is it important to deal with those? We also keep torrents in the torrent_cache. 

I don't think deleting old torrents is in the critical path of this feature delivering value to users.

src/pipeline.ts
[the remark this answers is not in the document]
mmgoodnow: later on where?

src/pipeline.ts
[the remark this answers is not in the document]
ShanaryS: I don't see why we shouldn't be deleting them. If they are in `outputDir` for more than a week, then it means the searchee has been deleted and will never be injected. There is no valid use for these torrents. Every time the job is ran, there will be hundreds or thousands of attempts constantly only getting worse with time.

The docs for this feature makes it clear that users should no longer be interacting with `outputDir`, I don't see any reason to keep these torrents.

src/pipeline.ts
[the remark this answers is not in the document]
ShanaryS: If this torrent was a corrupted cross seed, meaning it never rechecks to 100%. Then it's possible for `cross-seed` to trigger a recheck everytime the job is ran until it's deleted after a week. A check is done to see if `fullMatches.length > 0` which allows us to delete the .torrent file safely after linking.

src/pipeline.ts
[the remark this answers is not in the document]
zakkarry: I tend to agree on this, and assuming that if action is not set to `save` there is no reason why a user would typically need those files past any particular date after a week of failing to inject.

Am i missing something?

src/inject.ts
```diff
@@ -0,0 +1,620 @@
...
+					areMediaTitlesSimilar(searcheeFile.name, metaFile.name),
+				),
+			)
+		) {
+			logger.warn({
+				label: Label.INJECT,
+				message: `Skipping potential false positive for ${getLogString(meta, chalk.bold.white)} from ${getLogString(searchee, chalk.bold.white)}`,
+			});
+			continue;
+		}
```
mmgoodnow, line 139: What does this protect us from? MATCH_SIZE_ONLY matches that may be incorrect?

src/inject.ts
```diff
@@ -0,0 +1,620 @@
...
+
+		const {
+			injectionResult,
+			matchedSearchee,
+			matchedDecision,
+			linkedNewFiles,
+		} = await injectInitialAction(meta, matches, tracker);
+		if (injectionResult === InjectionResult.FAILURE) {
+			injectionFailed(progress, filePathLog, injectionResult, summary);
+			continue; // Never delete failed torrent files
```
mmgoodnow, line 493: why never delete failed torrent files?

zakkarry: If I recall, this failure is potentially due to incomplete torrents and there is a cleanup done independently after a week of age.

Don't quote me, this is just what I think I remember.

src/inject.ts
zakkarry: Would it not be prudent to check `matchMode !== MatchMode.SAFE` prior to checking the titles if this is the case? Otherwise, it could do this unnecessarily?

src/inject.ts
ShanaryS: I added this a few days ago as injection failures is always transient or user error.

src/inject.ts
ShanaryS: It protects against bad matches from risky or partial. Think of it similar to how torznab search or reverselookup requires titles to be similar.

This is largely unnecessary for inject though as every .torrent in outputDir already passed the above filters and more, but it could technically match with another unrelated torrent a user has.

I don't think it's necessary to check for matchMode as the first check passes in 100% of cases.

ShanaryS:
I put the action logging as verbose for inject so there should only be 1 line per torrent.

src/pipeline.ts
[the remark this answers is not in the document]
mmgoodnow: > there will be hundreds or thousands of attempts constantly only getting worse with time.

These attempts are cheap. They don't do any tracker IO, and "hundreds or thousands" doesn't really pass the sniff test to me. I am the oldest user of cross-seed and my outputDir has 17 torrents in it. (they look like they're all from v6 as well, did the old ones get rewritten to the new format or is v6 just less reliable?)

There's no guarantee that `cross-seed` owns every torrent in `outputDir`, and it **cannot** make destructive assumptions about torrents it doesn't own. We experienced this with Sonarr and qBittorrent - Sonarr assumes that it owns every torrent with that category, even though we _have_ to use categories to store the files in the right place. That forced us to implement a hacky workaround. 

Here's what I'll propose—we delete files if they look like cross-seed created them. That is, they follow the `[mediaType][tracker]torrentname[infoHash].torrent` format. I pushed it to a separate branch because I'm not sure if I got all the logic right. https://github.com/cross-seed/cross-seed/pull/770/files

src/inject.ts
mmgoodnow: shouldn't this code live in `decide` then?

ShanaryS: `areMediaTitlesSimilar` could go in there but I think it's fine in utils. And the use of it is specific and straightforward, I don't think a wrapper function is needed.

src/pipeline.ts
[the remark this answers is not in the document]
ShanaryS: The only scenarios deletion will happen now is: if no matches are found, if it's blocklisted, or if it's in client and complete.

In the case of a stalled injection or source, the user will now need to remove the stalled torrent before it gets deleted.

Injection failures won't cause deletion as it should be temporary or user error.

src/inject.ts
mmgoodnow: what I'm saying is that this sounds like a decision rejection. Why wouldn't this make sense at the end of assessCandidateHelper?

src/pipeline.ts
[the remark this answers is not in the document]
mmgoodnow: > if no matches are found, if it's blocklisted, or if it's in client and complete.

Can we just do the third one in this PR?

src/inject.ts
ShanaryS: I don't think it adds any benefit to any other method. Meaning if this check fails, then it wouldn't have matched reverse lookup or been in the search results. This is just here to bring the safety closer to those methods, but it's not as strict as search and rss/announce.

src/pipeline.ts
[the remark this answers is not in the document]
ShanaryS: Users would need to manually clean up outputDir after deleting stalled torrents or if they use a blocklist. Remember that these torrents only exist in outputDir because injection failed. In an ideal world, these torrents wouldn't have been saved in the first place. So it feels like creating more work for the user for no real gain.

mmgoodnow:
Okay, I think the code is in a pretty good place. My latest changes are mostly around standardizing the helper boundaries and making an abstraction for what happens after `injectInitialAction`. There should be no runtime changes, but I don't have a testing setup for this PR.

mmgoodnow (approved):

ShanaryS (commented):
Good to merge

mmgoodnow (approved):

PR #679, opened 2024-05-09
refactor(client/deluge): switch to using Result class for calls

zakkarry:
refactors deluge to use the Result class rather than simply throw'ing

also fixes errors deluge would throw when trying to set labels that were not present. deluge would throw RPC error's, and we would base adding off that. now we get the current labels and decide whether to add or only set the label.

added jsdocs for kicks.

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -78,7 +78,7 @@ export default class Deluge implements TorrentClient {
 		try {
 			const authResponse = (
 				await this.call<boolean>("auth.login", [password], 0)
-			).unwrapOrThrow();
+			).unwrapOrThrow(new Error("failed to connect for authentication"));
 
 			if (!authResponse) {
 				throw new CrossSeedError(
@@ -102,7 +102,7 @@ export default class Deluge implements TorrentClient {
 			);
 			const webuiHostList = (
 				await this.call<WebHostList>("web.get_hosts", [], 0)
-			).unwrapOrThrow();
+			).unwrapOrThrow(new Error("failed to get host-list for reconnect"));
 			const connectResponse = await this.call<undefined>(
 				"web.connect",
 				[webuiHostList[0][0]],
@@ -188,7 +188,7 @@ export default class Deluge implements TorrentClient {
 		if (json.error) {
 			return resultOfErr(json.error);
 		}
-		return resultOf(json.result as ResultType);
+		return resultOf(json.result!);
 	}
 
 	/**
@@ -276,7 +276,7 @@ export default class Deluge implements TorrentClient {
 				]);
 			}
 			if (setResult.isErr()) {
-				throw new Error("Labeling Failure");
+				throw new Error(setResult.unwrapErrOrThrow().message);
 			}
 		} catch (e) {
 			logger.debug(e);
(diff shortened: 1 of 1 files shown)


src/clients/Deluge.ts
```diff
@@ -233,32 +247,53 @@ export default class Deluge implements TorrentClient {
...
-				label,
-			]);
-			if (setResult.error?.code === DelugeErrorCode.RPC_FAIL) {
+			const getCurrentLabels = await this.call<string[]>(
+				"label.get_labels",
+				[],
+			);
+			if (getCurrentLabels.isErr()) {
+				this.isLabelEnabled = false;
+				throw new Error("Labels have been disabled.");
```
mmgoodnow, line 263: can we just silently fail if labeling is not enabled? I don't think it needs to be required in order to use cross-seed

src/clients/Deluge.ts
```diff
@@ -233,32 +247,53 @@ export default class Deluge implements TorrentClient {
...
 				await this.call<void>("label.add", [label]);
-				await this.call<void>("label.set_torrent", [infoHash, label]);
+				await wait(300);
+				setResult = await this.call<void>("label.set_torrent", [
+					newTorrent.infoHash,
+					label,
+				]);
+			}
+			if (setResult.isErr()) {
+				throw new Error("Labeling Failure");
```
mmgoodnow, line 279: might help to expose the error object here?

src/clients/Deluge.ts
```diff
@@ -295,20 +330,31 @@ export default class Deluge implements TorrentClient {
...
-
-			const addResponse =
-				typeof addResult.error?.message === "string"
-					? addResult.error
-					: addResult.result;
-
-			if (typeof addResponse === "string") {
+			if (addResponse.isErr()) {
+				const addResponseError = addResponse.unwrapErrOrThrow();
+				if (addResponseError.message?.includes("already")) {
```
mmgoodnow, line 339: is there not always a message?

mmgoodnow (approved):
looks good! I have a couple nits.

src/clients/Deluge.ts
```diff
@@ -181,11 +184,16 @@ export default class Deluge implements TorrentClient {
 			}
 		}
 		this.handleResponseHeaders(response.headers);
-		return json;
+
+		if (json.error) {
+			return resultOfErr(json.error);
+		}
+		return resultOf(json.result as ResultType);
```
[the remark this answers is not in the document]
zakkarry: It was giving the possibility that its undefined, i added an assertion.

src/clients/Deluge.ts
zakkarry: I believe there is...i'll assert this

src/clients/Deluge.ts
zakkarry: This was added in the case that the label plugin was disabled or failed in an irrecoverable manner (for some reason) after cross-seed had started and determined that it was loaded.

src/clients/Deluge.ts
```diff
@@ -233,32 +247,53 @@ export default class Deluge implements TorrentClient {
...
-			const setResult = await this.call<void>("label.set_torrent", [
-				infoHash,
-				label,
-			]);
-			if (setResult.error?.code === DelugeErrorCode.RPC_FAIL) {
+			const getCurrentLabels = await this.call<string[]>(
+				"label.get_labels",
+				[],
+			);
+			if (getCurrentLabels.isErr()) {
```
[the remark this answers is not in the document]
zakkarry: It shouldn't be necessary, given the change I made to check which labels already exist. Previously we were relying on that response to be the way to determine if it needs to create the label.

It now does this correctly rather than dependence on a failure.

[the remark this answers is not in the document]
mmgoodnow: sounds good.

mmgoodnow (approved):

PR #681, opened 2024-05-10
enhancement(match): Display size diff in bytes if similar

ShanaryS:


src/decide.ts
```diff
@@ -51,9 +52,11 @@ const createReasonLogger =
 			case Decision.MATCH:
 				return;
 			case Decision.SIZE_MISMATCH:
-				reason = `its size does not match - (${humanReadableSize(
-					searchee.length,
-				)} -> ${humanReadableSize(candidate.size)})`;
+				reason = `its size does not match - (${
+					Math.abs(sizeDiff) > 10000000 // 10MB is the rounding error
+						? `${humanReadableSize(searchee.length)} -> ${humanReadableSize(candidate.size)}`
+						: `${sizeDiff > 0 ? "+" : ""}${sizeDiff} bytes`
```
zakkarry, line 58: Do you think adding a `-` instead of nothing would be clearer?

mmgoodnow (approved):
Lgtm pending the nit about `-`

src/decide.ts
ShanaryS: That's already covered by the number being negative.

PR #683, opened 2024-05-12
Refactor Result type

mmgoodnow:
The current Result type is okay, but it doesn't actually leverage Typescript's ability to catch bugs at build time. This PR does a few things:
- Result is now a **union type** of `OkResult | ErrResult`
- `unwrapOrThrow(err?: Error)` has been separated into two methods:
    - `unwrap()` this method ONLY exists on OkResult. In order to call it, you need to use `isOk` or `isErr` to do type narrowing so that TypeScript knows the method exists. 
    - `unwrapOrThrow(err: Error)` This is a convenience method to throw an error if the result is an ErrResult. It is like the old version of `unwrapOrThrow`, except the argument is now required. This exists on BOTH OkResult and ErrResult. 
    - I migrated all calls to `unwrapOrThrow()` with no arguments to use `unwrap()`, and kept calls of `unwrapOrThrow(new Error('error message'))` with one argument to use `unwrapOrThrow()`. 
- `unwrapErr()` this method ONLY exists on ErrResult. In order to call it, you need to use `isOk` or `isErr` to do type narrowing so that TypeScript knows the method exists. 
- `isOk` and `isErr` are now [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) which means they will do type narrowing. Once you call `result.isOk()`, typescript "grants you access" to calling the `unwrap()` method. 
- I added a first-class `isOk` function (not an instance method on either of the result classes) because TypeScript doesn't understand when you `.filter(r => r.isOk())`. It understands `filter(isOk)` though.

### DIFF
--- a/src/Result.ts
+++ b/src/Result.ts
@@ -32,10 +32,6 @@ class OkResult<T> {
 	orElse(): T {
 		return this.contents;
 	}
-
-	toArray(): [T] {
-		return [this.contents];
-	}
 }
 
 class ErrResult<U> {
@@ -72,10 +68,6 @@ class ErrResult<U> {
 	orElse<T>(t: T): T {
 		return t;
 	}
-
-	toArray(): [] {
-		return [];
-	}
 }
 
 export type Result<T, U> = OkResult<T> | ErrResult<U>;
@@ -86,4 +78,8 @@ export function resultOf<T, U>(value: T): Result<T, U> {
 
 export function resultOfErr<T, U>(value: U): Result<T, U> {
 	return new ErrResult(value);
+}
+
+export function isOk<T, U>(result: Result<T, U>): result is OkResult<T> {
+	return result.isOk();
 }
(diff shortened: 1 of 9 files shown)


src/Result.ts
```diff
@@ -1,83 +1,85 @@
...
 
-	mapOk<R>(mapper: (t: T) => R): Result<R, U> {
+	mapOk<R>(mapper: (t: T) => R): OkResult<R> {
 		return new OkResult(mapper(this.contents));
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	mapErr<R>(mapper: (u: U) => R): Result<T, R> {
-		return this as unknown as Result<T, R>;
+	mapErr(): OkResult<T> {
```
ShanaryS, line 20: Is the return type intentional? I'm not familiar with JS or TS internals so I can't comment much.

PR #689, opened 2024-05-17
enhancement(decide/match): support multi-file risky based matching

ShanaryS:
This mainly is to support multi-file risky based matching moving the 100% MATCH_PARTIAL matches to MATCH_SIZE_ONLY. I refactored decide.js to the state it will be downstream.

1st commit:
1. The priority of decisions 
2. Reporting more granular decisions
3. Caching all snatched .torrent files and saving infoHash with decision if present (might as well start doing this now for `cache-search-queries`). We can optionally cleanup based on modified time since it is updated each time it's used.
4. Better error message for decisions and snatch.

2nd commit:
1. Support multi-file risky based matches

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -19,6 +19,7 @@ export const ANIME_REGEX =
 export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w ]+?)(?:\[.+\])?(?:\))?(?=(?:\.\w{1,5})?$)/i;
 export const RESOLUTION_REGEX = /\b(?<res>\d{3,4}[ipx])\b/i;
+export const RES_STRICT_REGEX = /(?<res>(?:2160|1080|720)[pi])/;
 
 export const REPACK_PROPER_REGEX =
 	/(?:\b(?<type>(?:REPACK|PROPER|\d\v\d)\d?))|(?<arrtype>(?:Proper|v\d))\b/;
@@ -93,9 +94,12 @@ export enum Decision {
 	MATCH = "MATCH",
 	MATCH_SIZE_ONLY = "MATCH_SIZE_ONLY",
 	MATCH_PARTIAL = "MATCH_PARTIAL",
+	FUZZY_SIZE_MISMATCH = "FUZZY_SIZE_MISMATCH",
 	SIZE_MISMATCH = "SIZE_MISMATCH",
+	PARTIAL_SIZE_MISMATCH = "PARTIAL_SIZE_MISMATCH",
 	NO_DOWNLOAD_LINK = "NO_DOWNLOAD_LINK",
 	DOWNLOAD_FAILED = "DOWNLOAD_FAILED",
+	MAGNET_LINK = "MAGNET_LINK",
 	RATE_LIMITED = "RATE_LIMITED",
 	INFO_HASH_ALREADY_EXISTS = "INFO_HASH_ALREADY_EXISTS",
 	FILE_TREE_MISMATCH = "FILE_TREE_MISMATCH",
@@ -103,6 +107,19 @@ export enum Decision {
 	BLOCKED_RELEASE = "BLOCKED_RELEASE",
 	PROPER_REPACK_MISMATCH = "PROPER_REPACK_MISMATCH",
 	RESOLUTION_MISMATCH = "RESOLUTION_MISMATCH",
+}
+export type DecisionAnyMatch =
+	| Decision.MATCH
+	| Decision.MATCH_SIZE_ONLY
+	| Decision.MATCH_PARTIAL;
+export function isDecisionAnyMatch(
+	decision: Decision,
+): decision is DecisionAnyMatch {
+	return (
+		decision === Decision.MATCH ||
+		decision === Decision.MATCH_SIZE_ONLY ||
+		decision === Decision.MATCH_PARTIAL
+	);
 }
 
 export enum MatchMode {
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -127,13 +127,20 @@ export function compareFileTreesIgnoringNames(
 	candidate: Metafile,
 	searchee: Searchee,
 ): boolean {
-	const cmp = (candidate, searchee) => {
-		return searchee.length === candidate.length;
-	};
-
-	return candidate.files.every((elOfA) =>
-		searchee.files.some((elOfB) => cmp(elOfA, elOfB)),
-	);
+	for (const candidateFile of candidate.files) {
+		let matchedSearcheeFiles = searchee.files.filter(
+			(searcheeFile) => searcheeFile.length === candidateFile.length,
+		);
+		if (matchedSearcheeFiles.length > 1) {
+			matchedSearcheeFiles = matchedSearcheeFiles.filter(
+				(searcheeFile) => searcheeFile.name === candidateFile.name,
+			);
+		}
+		if (matchedSearcheeFiles.length === 0) {
+			return false;
+		}
+	}
+	return true;
 }
 
 export function comparePartialSizeOnly(
@@ -306,11 +313,7 @@ async function assessCandidateHelper(
 	}
 
 	const sizeMatch = compareFileTreesIgnoringNames(metafile, searchee);
-	if (
-		sizeMatch &&
-		matchMode !== MatchMode.SAFE &&
-		searchee.files.length === 1
-	) {
+	if (sizeMatch && matchMode !== MatchMode.SAFE) {
 		return { decision: Decision.MATCH_SIZE_ONLY, metafile };
 	}
 
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -1,12 +1,13 @@
-import { existsSync, writeFileSync } from "fs";
+import { existsSync, statSync, utimesSync, writeFileSync } from "fs";
 import path from "path";
 import { appDir } from "./configuration.js";
 import {
 	Decision,
+	isDecisionAnyMatch,
 	MatchMode,
 	RELEASE_GROUP_REGEX,
 	REPACK_PROPER_REGEX,
-	RESOLUTION_REGEX,
+	RES_STRICT_REGEX,
 	TORRENT_CACHE_FOLDER,
 } from "./constants.js";
 import { db } from "./db.js";
@@ -32,13 +33,13 @@ const createReasonLogger =
 		searchee: Searchee,
 		candidate: Candidate,
 	): void => {
+		if (cached) return;
 		function logReason(reason): void {
 			logger.verbose({
 				label: Label.DECIDE,
 				message: `${name} - no match for ${tracker} torrent ${Title} - ${reason}`,
 			});
 		}
-		const sizeDiff = candidate.size - searchee.length;
 
 		let reason;
 		switch (decision) {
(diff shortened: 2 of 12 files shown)


src/decide.ts
```diff
@@ -227,84 +267,89 @@ function releaseGroupDoesMatch(
...
 	}
 
-	return { decision: Decision.FILE_TREE_MISMATCH };
+	return { decision: Decision.FILE_TREE_MISMATCH, metafile };
 }
 
-function existsInTorrentCache(infoHash: string): boolean {
-	return existsSync(
-		path.join(appDir(), TORRENT_CACHE_FOLDER, `${infoHash}.cached.torrent`),
+function existsInTorrentCache(infoHash: string | undefined): boolean {
```
mmgoodnow, line 343: I think this function should take `string` rather than `string | undefined`. Let's push the null checks to the callers

src/decide.ts
```diff
@@ -227,84 +267,89 @@ function releaseGroupDoesMatch(
...
+	if (!infoHash) return false;
+	const torrentPath = path.join(
+		appDir(),
+		TORRENT_CACHE_FOLDER,
+		`${infoHash}.cached.torrent`,
 	);
+	if (!existsSync(torrentPath)) return false;
+	utimesSync(torrentPath, new Date(), new Date());
```
mmgoodnow, line 351: this changes the modification time even though we're not modifying the file. would it be possible to just change the access time?

src/constants.ts
```diff
@@ -104,6 +108,15 @@ export enum Decision {
...
 }
+export type DecisionAllMatch =
+	| Decision.MATCH
+	| Decision.MATCH_SIZE_ONLY
+	| Decision.MATCH_PARTIAL;
+export const DECISION_ALL_MATCH = [
+	Decision.MATCH,
+	Decision.MATCH_SIZE_ONLY,
+	Decision.MATCH_PARTIAL,
+];
```
mmgoodnow, line 119: A more idiomatic way to express this in TypeScript would be to write a type predicate:
```ts
export function isDecisionAllMatch(decision: Decision): decision is DecisionAllMatch {
  return decision === Decision.MATCH || 
    decision === Decision.MATCH_SIZE_ONLY || 
    decision === Decision.MATCH_PARTIAL
}
```
`DECISION_ALL_MATCH.includes(decision)` can be replaced with `isDecisionAllMatch(decision)` and we won't lose type fidelity at the call site.

src/pipeline.ts
```diff
@@ -245,17 +244,13 @@ export async function checkNewCandidateMatch(
...
-		assessment.decision !== Decision.MATCH_PARTIAL
-	) {
+	if (!DECISION_ALL_MATCH.includes(assessment.decision)) {
 		return null;
 	}
 
 	const result = await performAction(
 		assessment.metafile!,
-		assessment.decision,
+		assessment.decision as DecisionAllMatch,
```
mmgoodnow, line 253: after making the `isDecisionAllMatch` change, this `as` shouldn't be necessary any more.

src/torrent.ts
```diff
@@ -64,35 +66,43 @@ export async function snatch(
...
-				`Unsupported: magnet link detected at ${tracker}: ${url}`,
+				`Unsupported: magnet link detected from ${tracker} for ${candidate.name}`,
 			);
+			logger.debug(`${candidate.name}: ${url}`);
 			return resultOfErr(SnatchError.MAGNET_LINK);
 		}
-		logger.error(`failed to access ${tracker}: ${url}`);
-		logger.debug(e);
+		logger.error(`Failed to access ${tracker} for ${candidate.name}`);
+		logger.debug(`${candidate.name}: ${url} - ${e}`);
```
mmgoodnow, line 82: this means we lose the `e` stack trace. can we split this into two lines?
```suggestion
		logger.debug(`${candidate.name}: ${url}`);
		logger.debug(e);
```

src/torrent.ts
```diff
@@ -64,35 +66,43 @@ export async function snatch(
...
 
 	if (response.status === 429) {
 		return resultOfErr(SnatchError.RATE_LIMITED);
 	} else if (!response.ok) {
 		logger.error(
-			`error downloading torrent from ${tracker} at ${url}: ${response.status} ${response.statusText}`,
+			`Error downloading torrent from ${tracker} for ${candidate.name}: ${response.status} ${response.statusText}`,
+		);
+		logger.debug(
+			`${candidate.name}: ${url} - Response: ${await response.text()}`,
```
mmgoodnow, line 93: I don't love the idea of logging the entire response text - they tend to be really large and murder log file usability IME. A common pattern I've adopted is logging the first 1000 characters, would that work here?

src/torrent.ts
```diff
@@ -64,35 +66,43 @@ export async function snatch(
...
 		);
-		logger.debug("response: %s", await response.text());
 		return resultOfErr(SnatchError.UNKNOWN_ERROR);
 	} else if (response.headers.get("Content-Type") === "application/rss+xml") {
 		const responseText = await response.clone().text();
 		if (responseText.includes("429")) {
 			return resultOfErr(SnatchError.RATE_LIMITED);
```
mmgoodnow, line 99: We can probably delete this. At one point Prowlarr was sending 429 but not as a status code, only as a field in the response XML. I'm pretty sure I got them to change it though (this was like 2 years ago)

src/torrent.ts
```diff
@@ -105,11 +115,14 @@ export async function snatch(
...
+		logger.error(
+			`Invalid torrent contents from ${tracker} for ${candidate.name}`,
+		);
 		const contentType = response.headers.get("Content-Type");
 		const contentLength = response.headers.get("Content-Length");
-		logger.debug(`Content-Type: ${contentType}`);
-		logger.debug(`Content-Length: ${contentLength}`);
+		logger.debug(
+			`${candidate.name}: ${url} - Content-Type: ${contentType} - Content-Length: ${contentLength}`,
+		);
```
mmgoodnow, line 125: might as well throw in a `logger.debug(e)` as well here

mmgoodnow (commented):
architecture looks good, I like the new organization of `decide.ts`. I have several comments but they're pretty minor for the most part

src/decide.ts
```diff
@@ -227,84 +267,89 @@ function releaseGroupDoesMatch(
...
 	}
-	const result = await snatch(link, tracker);
+	if (size) {
+		if (!fuzzySizeDoesMatch(size, searchee)) {
+			return { decision: Decision.FUZZY_SIZE_MISMATCH };
+		}
+	} else {
+		if (!resDoesMatch(searchee.name, name, matchMode)) {
+			return { decision: Decision.RESOLUTION_MISMATCH };
+		}
```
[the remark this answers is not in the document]
zakkarry: The main reason I put this in there was in the case of announces where size is not provided, resolution - when paired with the other indicators such as group and repack/proper - is the closest we can get to a size comparable check, given the limited information.

It is a somewhat cheap check, but I'm not entirely sure its necessary given the correlation size and resolution would have

src/decide.ts
```diff
@@ -48,17 +52,23 @@ const createReasonLogger =
...
+			case Decision.FUZZY_SIZE_MISMATCH ||
+				Decision.SIZE_MISMATCH ||
+				Decision.PARTIAL_SIZE_MISMATCH:
+				reason = `its size does not match by ${decision}${
+					sizeDiff === undefined
+						? ""
+						: ` - (${
+								Math.abs(sizeDiff) > 10000000 // 10MB is the rounding error
+									? `${humanReadableSize(searchee.length)} -> ${humanReadableSize(candidate.size)}`
+									: `${sizeDiff > 0 ? "+" : ""}${sizeDiff} bytes`
```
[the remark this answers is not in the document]
ShanaryS: The displaying of bytes was added as users kept asking why it said `SIZE_MISMATCH` when the sizes were equal due to rounding. We could always display bytes diff but I think using GB when possible is more readable.

src/decide.ts
```diff
@@ -227,84 +267,89 @@ function releaseGroupDoesMatch(
...
+			return { decision: Decision.RESOLUTION_MISMATCH };
+		}
+	}
+	if (!link) {
+		return { decision: Decision.NO_DOWNLOAD_LINK };
+	}
+
+	if (findBlockedStringInReleaseMaybe(searchee, blockList)) {
+		return { decision: Decision.BLOCKED_RELEASE };
+	}
```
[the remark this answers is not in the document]
ShanaryS: Due to how the decisions will be retried in the cache branch, it's better if we can reject it by the other reasons. Decisions like group, repack, and resolution will never be retried since they can't change. Fuzzy won't be unless the value is now higher.

I also think blocklist further down makes more sense to limit its occurrence. The group not matching means it will never be a cross seed so it should be reported first. So blocklist decisions can be interpreted as "passed all checks before snatch".

src/decide.ts
```diff
@@ -227,84 +267,89 @@ function releaseGroupDoesMatch(
...
+		return err === SnatchError.RATE_LIMITED
 			? { decision: Decision.RATE_LIMITED }
-			: { decision: Decision.DOWNLOAD_FAILED };
+			: err === SnatchError.MAGNET_LINK
+				? { decision: Decision.MAGNET_LINK }
+				: { decision: Decision.DOWNLOAD_FAILED };
 	}
+	const metafile = result.unwrap();
+	cacheTorrentFile(metafile);
+	candidate.size = metafile.length; // Trackers can be wrong
```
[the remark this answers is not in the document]
ShanaryS: This is for reporting the size difference when logging. The size is never used for logic after this.

src/decide.ts
ShanaryS: Yes but from my understanding, modified time is the only thing guaranteed to be stable across operating systems. It seems as we can rely on access time for our use cases though. But some user configurations (I think very unlikely) might not use access time.

This will only matter if we want to clean up cached torrent files. Say older than `Math.max(excludeOlder, ms("1y"))`.

src/decide.ts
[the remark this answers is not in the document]
mmgoodnow: let's keep things simple and just always show the diff as a human readable size. if the diff is big enough you can see it in the first two numbers but having the third number won't hurt.

src/decide.ts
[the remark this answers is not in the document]
mmgoodnow: good points 👍

src/decide.ts
mmgoodnow: what very unlikely user configurations would those be?

mmgoodnow (dismissed):
LGTM pending the resolution of these threads: 

https://github.com/cross-seed/cross-seed/pull/689#discussion_r1619187882
https://github.com/cross-seed/cross-seed/pull/689#discussion_r1619208403

src/decide.ts
ShanaryS: If the user sets `noatime` for their mount in linux, then it won't be tracked by the filesystem. We can only update atime for now but if we want to use it for clearing cache, modified avoids any potential issues.

src/decide.ts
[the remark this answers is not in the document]
ShanaryS: I combined both methods. It now displays the diff in a human readable format ` +232B` or `-1.23GB`

src/decide.ts
mmgoodnow: eh, that seems esoteric enough that I'm inclined not to worry about it. I'd rather have the semantics be right. The alternative would be to track them in the database but that feels heavier than necessary.

ShanaryS:
Good to merge. Had to update `humanReadable` to use absolute values for `Math.log`.

zakkarry:
> Good to merge. Had to update `humanReadable` to use absolute values for `Math.log`.

#695 ?

PR #693, opened 2024-05-22
fix(clients): set download path when adding torrent

ShanaryS:
This prevents hardlinks from breaking when a torrent is moved from/to the `save_path` after rechecking a match partial and completing the torrent. Only matters if the user's `downloadPath` is on a different drive. Tested and working in qBittorrent. Need to address for other clients as well.


mmgoodnow (approved):
I worry that not all clients will have the concept of per-torrent "download path". For qbit, this looks good.

PR #696, opened 2024-05-26
enhancement(arr/id): use ids for anime/unknown when possible

ShanaryS:
1. Using `ParsedMedia = ParsedMovie | ParsedSeries` instead of `ExternalIds` so we can track parsed episode information from sonarr
2. MediaType `ANIME` and `OTHER` will try all arrs for an id to use for searching.
3. If MediaType `ANIME`, search indexer if supports anime, tv, or movie categories instead of just anime.


src/arr.ts
```diff
@@ -211,53 +220,71 @@ function getRelevantArrInstances(mediaType: MediaType): string[] {
 			return sonarr ?? [];
 		case MediaType.MOVIE:
 			return radarr ?? [];
```
mmgoodnow, line 222: can we make `sonarr` and `radarr` transform to an empty array at the zod layer so we don't have to do these null checks everywhere?

zakkarry: yes, we can do that

src/configSchema.ts
```diff
@@ -176,8 +176,15 @@ export const VALIDATION_SCHEMA = z
 			.nullish()
 			.transform((value) => (Array.isArray(value) ? value : [])),
 		apiKey: z.string().min(24).nullish(),
-		radarr: z.array(z.string().url()).nullish(),
-		sonarr: z.array(z.string().url()).nullish(),
+		radarr: z
+			.array(z.string().url())
+			.nullish()
+			.transform((value) => (typeof value === "object" ? value : [])),
```
mmgoodnow, line 182: `typeof null === "object"` so this might be better off with `??`
```suggestion
			.transform(value => value ?? []),
```

mmgoodnow (approved):
lgtm pending comments

src/arr.ts
```diff
@@ -208,56 +217,74 @@ function getRelevantArrInstances(mediaType: MediaType): string[] {
...
+		let error = new Error(`No ids found for ${title} [${mediaType}]`);
 		for (const uArrL of uArrLs) {
-			const externalIds = await getExternalIdsFromArr(searchee, uArrL);
-			if (Object.values(externalIds).some(isTruthy)) {
-				logArrQueryResult(externalIds, searchee.name, mediaType);
-				return resultOf(externalIds);
+			const name =
+				mediaType === MediaType.OTHER &&
+				getRuntimeConfig().sonarr.includes(uArrL)
+					? `${title} S00E00` // Sonarr needs season or episode
```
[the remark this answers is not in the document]
zakkarry: It does not matter if it exists in the metadata or not, as it is only used to parse.

You can test this by finding a series without a special and entering 'Series.S00E00.' into the "Test Parsing" spot on the series index. 

It will need to be in your LIbrary (added) though.

src/arr.ts
```diff
@@ -208,56 +217,74 @@ function getRelevantArrInstances(mediaType: MediaType): string[] {
...
+			const ids = response.movie ?? response.series;
+			if (Object.values(ids).some(isTruthy)) {
+				const label = response.movie ? Label.RADARR : Label.SONARR;
+				logArrQueryResult(ids, name, label);
+				return resultOf(response);
 			}
 		}
-		throw new Error("fall through to catch");
+		throw error;
 	} catch (error) {
```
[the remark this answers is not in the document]
zakkarry: I think it's possible it could be changed to just if logic, at some point while writing this fetch would throw but that's been refactored a couple of times since...

I'll defer to @ShanaryS to handle this.

src/configSchema.ts
zakkarry: I was unaware, good to know.

PR #698, opened 2024-05-29
fix(logging/arr): check if corresponding arr is disabled before logging

zakkarry:
when there are no provided arr corresponding to the media type, it will still erroneously log that it is missing.

this adds a simple check early in the arr flow to make sure there is a corresponding URL before proceeding

rather than switch to a result for something so simple, and because of #696 being open right now, i kept this as a simple length check.

note: I'm not sure this is crucial for merging, but depending on the timeframe that #696 takes, i figured that getting something to demonstrate the problem might be best before i forget

### DIFF
--- a/src/arr.ts
+++ b/src/arr.ts
@@ -221,6 +221,9 @@ export async function scanAllArrsForExternalIds(
 	mediaType: MediaType,
 ): Promise<Result<ExternalIds, boolean>> {
 	const uArrLs = getRelevantArrInstances(mediaType);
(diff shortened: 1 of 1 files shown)


zakkarry:
i'm a idiot

https://github.com/cross-seed/cross-seed/pull/696/files#diff-9ccbad1d8c57aed9e2c6c42e9f699003b1bbaf8746f8aa9284f1e4ee4f62c57aR236-R238

PR #699, opened 2024-05-31
update(logging): Strip ansi from logs and general logging improvements

ShanaryS:
Logging improvements that were made across branches. Also added a new dependency strip-ansi from the same maintainers as chalk. This removes the ansi codes that clutters up the logs but keeps them in console. Also redacting more sensitive information.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,4 +1,5 @@
 import { join } from "path";
+import stripAnsi from "strip-ansi";
 import winston from "winston";
 import DailyRotateFile from "winston-daily-rotate-file";
 import { appDir, createAppDir } from "./configuration.js";
@@ -53,8 +54,11 @@ function redactMessage(message: string | unknown, options?: RuntimeConfig) {
 	const runtimeConfig = options ?? getRuntimeConfig();
 	let ret = message;
 
-	// redact torznab api keys
-	ret = ret.replace(/apikey=[a-zA-Z0-9]+/g, `apikey=${redactionMsg}`);
+	ret = ret.replace(/key=[a-zA-Z0-9]+/g, `key=${redactionMsg}`);
+	ret = ret.replace(/pass=[a-zA-Z0-9]+/g, `pass=${redactionMsg}`);
+	ret = ret.replace(/(?:auto\.\d+\.)([a-zA-Z0-9]+)/g, (match, key) =>
+		match.replace(key, redactionMsg),
+	);
 	ret = ret.replace(
 		/\/notification\/crossSeed\/[a-zA-Z-0-9_-]+/g,
 		`/notification/crossSeed/${redactionMsg}`,
@@ -65,6 +69,13 @@ function redactMessage(message: string | unknown, options?: RuntimeConfig) {
 		}
 	}
 	return ret;
+}
+
+function stripAnsiChars(message: string | unknown) {
+	if (typeof message !== "string") {
+		return message;
+	}
+	return stripAnsi(message);
 }
 
 const logOnceCache: string[] = [];
@@ -85,11 +96,10 @@ export function initializeLogger(options: RuntimeConfig): void {
 			}),
 			winston.format.errors({ stack: true }),
 			winston.format.splat(),
-			winston.format.colorize(),
 			winston.format.printf(({ level, message, label, timestamp }) => {
 				return `${timestamp} ${level}: ${
 					label ? `[${label}] ` : ""
-				}${redactMessage(message, options)}`;
+				}${stripAnsiChars(redactMessage(message, options))}`;
 			}),
 		),
 		transports: [
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -41,7 +41,7 @@ import {
 	TorrentLocator,
 } from "./torrent.js";
 import { queryRssFeeds, searchTorznab } from "./torznab.js";
-import { filterAsync, isTruthy, stripExtension } from "./utils.js";
+import { filterAsync, getLogString, isTruthy } from "./utils.js";
 
 export interface Candidate {
 	guid: string;
@@ -83,6 +83,7 @@ async function assessCandidates(
 async function findOnOtherSites(
 	searchee: Searchee,
 	hashesToExclude: string[],
+	progress: string,
 ): Promise<FoundOnOtherSites> {
 	// make sure searchee is in database
 	await db("searchee")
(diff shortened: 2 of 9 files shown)


src/pipeline.ts
```diff
@@ -161,16 +164,21 @@ async function findMatchesBatch(
 	const { delay } = getRuntimeConfig();
 
 	let totalFound = 0;
-	for (const [i, sample] of samples.entries()) {
+	for (const [i, searchee] of samples.entries()) {
+		const searcheeLog = searchee.infoHash
+			? `${chalk.bold.white(searchee.name)} ${chalk.dim(`[${searchee.infoHash.slice(0, 8)}...]`)}`
+			: searchee.path
+				? chalk.bold.white(searchee.path)
+				: chalk.bold.white(searchee.name);
```
mmgoodnow, line 172: can this be extracted into a helper?

src/utils.ts
```diff
@@ -129,6 +129,8 @@ export function getAnimeQueries(name: string): string[] {
 		animeQueries.push(cleanseSeparators(`${title} ${release}`));
 	}
 	if (altTitle) {
+		if (altTitle.toLowerCase() === "season") return animeQueries;
+		if (altTitle.toLowerCase() === "ep") return animeQueries;
```
mmgoodnow, line 133: why this change?

src/torznab.ts
```diff
@@ -681,3 +662,76 @@ async function makeRequests(
...
+		skippingIndexersCallout +=
+			indexersToUseByTimestamp.length === indexersToUse.length
+				? "timestamps)"
+				: "timestamps ";
+	}
+	if (indexersToUseByTimestamp.length > indexersToUse.length) {
+		skippingIndexersCallout +=
+			enabledIndexers.length === indexersToUseByTimestamp.length
+				? "supported category)"
+				: "and supported category)";
```
mmgoodnow, line 710: there's a util for this: https://github.com/cross-seed/cross-seed/blob/9558b31f908d731d184c2d2b17b983175aab1d69/src/utils.ts#L163

src/torznab.ts
```diff
@@ -681,3 +662,76 @@ async function makeRequests(
...
+	}
+	if (indexersToUse.length === 0) {
+		logger.info({
+			label: Label.TORZNAB,
+			message: `${skippingSearching(skippingIndexersCallout)}`,
+		});
+		throw new Error("SKIPPED");
```
mmgoodnow, line 717: I don't really like the use of throw/catch for control flow like this. is there a way to achieve this behavior without throw/catch?

src/torznab.ts
```diff
@@ -681,3 +662,76 @@ async function makeRequests(
...
+	const res: Result<ParsedMedia, boolean> =
+		indexersToUse.length > 0
+			? await scanAllArrsForMedia(searchee, mediaType)
+			: resultOfErr(false);
+	const parsedMedia = res.isOk() ? res.unwrap() : undefined;
+	const ids = parsedMedia?.movie ?? parsedMedia?.series;
+	const idsStr = ids ? `: ${formatFoundIds(ids)}` : "";
+
+	const skippingIndexers = `Searching ${indexersToUse.length} indexers for ${searchee.name}${skippingIndexersCallout} [${mediaTypeStr}${idsStr}]`;
+	logger.info({
```
mmgoodnow, line 730: this log line contains a lot of information. can some of it move to a followup verbose log?

src/utils.ts
ShanaryS: Some of my anime was parsing these as alternate titles due to weird naming. It would end up searching "season" or "ep". It's easier to reject these here instead of through the regex.

src/torznab.ts
ShanaryS: I like having these all in info so it's easier for users to catch something being wrong. I myself have caught bugs because something didn't parse as expected. If we are to split it up, I rather put `indexersToUse.length` and `skippingIndxersCallout` in verbose.

src/logger.ts
```diff
@@ -53,8 +54,8 @@ function redactMessage(message: string | unknown, options?: RuntimeConfig) {
 	const runtimeConfig = options ?? getRuntimeConfig();
 	let ret = message;
 
-	// redact torznab api keys
-	ret = ret.replace(/apikey=[a-zA-Z0-9]+/g, `apikey=${redactionMsg}`);
+	ret = ret.replace(/key=[a-zA-Z0-9]+/g, `key=${redactionMsg}`);
+	ret = ret.replace(/pass=[a-zA-Z0-9]+/g, `pass=${redactionMsg}`);
```
zakkarry, line 58:
```suggestion
	ret = ret.replace(/pass=[a-zA-Z0-9]+/g, `pass=${redactionMsg}`);
	ret = ret.replace(/(?:auto\.\d+\.)([a-zA-Z0-9]+)/g, (match, key) =>
		match.replace(key, redactionMsg),
	);
```
needed for a specific tracker's custom formatting. tested already.

zakkarry (changes requested):

src/logger.ts
mmgoodnow: what format does it take?

src/torznab.ts
mmgoodnow: sg

src/utils.ts
mmgoodnow: can you add a comment to that effect?

src/logger.ts
zakkarry: it appends the rss key to the URL in the format

tracker.com/download/auto.torrentid#.rsskey

zakkarry (approved):

ShanaryS:
Removed the logging for searching from pipeline.ts so only torznab.ts logs.

src/torznab.ts
```diff
@@ -681,3 +642,98 @@ async function makeRequests(
...
+				: "and supported category)",
+		);
+	}
+	const reasonStr = new Intl.ListFormat("en", {
+		style: "narrow",
+		type: "unit",
+	}).format(skippingIndexersCallout);
+	if (indexersToUse.length === 0) {
+		logger.info(`${skippingSearching(reasonStr)}`);
+		return { indexersToUse };
```
mmgoodnow, line 721: this is not really taking advantage of Intl.ListFormat, which if used idiomatically, will handle all of the things you're doing with ternaries.

Try this: 
```js
const filteringCauses = [
  areIndexersBeingFilteredByTimestamps,
  areIndexersBeingFilteredBySupportedCategory,
].filter(isTruthy);
const reasonStr = filteringCauses.length ? `(filtered by ${formatAsList(filteringCauses)})` ? '';
```

src/utils.ts
```diff
@@ -160,6 +166,17 @@ export function humanReadableDate(timestamp: number): string {
 	return new Date(timestamp).toLocaleString("sv");
 }
 
+export function getLogString(data: Metafile | Searchee, color: ChalkInstance) {
+	if (data instanceof Metafile) {
+		return `${color(data.name)} ${chalk.dim(`[${data.infoHash.slice(0, 8)}...]`)}`;
+	}
```
mmgoodnow, line 172: is this instanceof check actually necessary? I would think duck typing should just work here

ShanaryS: It won't handle data.path without this check. It still thinks it could be Searchee after.

mmgoodnow (dismissed):

PR #705, opened 2024-06-05
fix(search/delay): value over max signed int produces default to 1s

### LINKED ISSUE #704: TimeoutOverflowWarning when using invalid config values
### Have you checked for any existing open/closed tickets that are the same or similar to your issue?

### I have gone to the Discord server for support and/or acknowledge that this is a bug and not a support issue

- [X] I've tried to solve this already

### What's happening surrounding the bug? (context)

When setting delay to 3,000,000,000, then running `cross-seed search`, you get the error: `(node:23) TimeoutOverflowWarning: 3000000000 does not fit into a 32-bit signed integer.`

### Reproducing the bug

Set the delay in the config to `3_000_000_000`

### What should be happening?

This should fail at the config validation step (when launching the container). This seems to actually fail when the delay value is accessed, which happens once per search call.

Moreso, when this fails, the default value for delay is set to 1, which is an extremely dangerous value to use. If the value is overflowed, and you do _not_ wish to fail validation, then I would suggest setting the value to the max 32-bit integer value, and probably still printing a warning during the config validation step.

### Your setup

```markdown
- OS: Unraid
- Prowlarr: (not relevant)
- Docker
```

### config.js (REDACT SENSITIVE INFORMATION)

```
{
  ...
  delay: 3_000_000_000,
  ...
}
```

### verbose logs

```
info: Validating your configuration...
info: Your configuration is valid!
info: [search] Found 1384 torrents, 1040 suitable to search for matches
info: [1/1040] Searching for ---
info: [torznab] Searching 1 indexers for ---
(node:23) TimeoutOverflowWarning: 3000000000 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
(Use `node --trace-warnings ...` to show where the warning was created)
info: [2/1040] Searching for ---
info: [torznab] Searching 1 indexers for ---
(node:23) TimeoutOverflowWarning: 3000000000 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
info: [3/1040] Searching for ---
info: [torznab] Searching 1 indexers for ---
(node:23) TimeoutOverflowWarning: 3000000000 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

### Failure or refusal to provide the information we require to address bugs will result in potential closure of the issue.

- [X] I acknowledge that i have searched existing issues, read the documentation and faq, read the logs to see if the answer was provided in error messaging, and attempted to get support via the Discord prior to submitting this issue.

zakkarry:
when using a value over max signed int cross-seed will default back to 1s, producing searching behavior that essentially hammers your indexers rather than producing any semblance of a longer timeout

closes #704

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -8,6 +8,7 @@ import {
 	Action,
 	LinkType,
 	MatchMode,
+	NEWLINE_INDENT,
 	PROGRAM_NAME,
 	PROGRAM_VERSION,
 } from "./constants.js";
@@ -73,7 +74,7 @@ export async function validateAndSetRuntimeConfig(options: RuntimeConfig) {
 			logger.error(
 				`${
 					path.length > 0 ? `Option: ${optionLine}` : "Configuration:"
-				}\n\t\t\t\t${message}\n\t\t\t\t(https://www.cross-seed.org/docs/basics/options${
+				}${NEWLINE_INDENT}${message}${NEWLINE_INDENT}(https://www.cross-seed.org/docs/basics/options${
 					urlPath ? `#${urlPath.toLowerCase()}` : ""
 				})\n`,
 			);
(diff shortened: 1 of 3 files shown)


mmgoodnow (approved):
Thoughts on something like 1 hour, with maybe a pointer to using searchCadence? That should prevent the kind of misunderstanding that happened here

mmgoodnow (approved):

PR #708, opened 2024-06-09
bug(indexers): skip indexers with null caps and refresh with indexer job

ShanaryS:
1. `getEnabledIndexers()` no longer returns indexers with null caps. Which is possible if an indexer was down every time `cross-seed` was restarted since migrating to v6.
2. Caps are refreshed before `search` and `webhook`, not just on startup. This prevents the change above from being a silent failure if users don't restart `cross-seed` for a long time. Also benefits in the normal case as indexer caps will always be in sync without needing a restart.

### DIFF
--- a/src/jobs.ts
+++ b/src/jobs.ts
@@ -4,6 +4,7 @@ import { main, scanRssFeeds } from "./pipeline.js";
 import { exitOnCrossSeedErrors } from "./errors.js";
 import { Label, logger } from "./logger.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
+import { updateCaps } from "./torznab.js";
 
 class Job {
 	name: string;
@@ -41,6 +42,7 @@ function getJobs(): Job[] {
 	if (torznab.length > 0) {
 		if (rssCadence) jobs.push(new Job("rss", rssCadence, scanRssFeeds));
 		if (searchCadence) jobs.push(new Job("search", searchCadence, main));
+		jobs.push(new Job("updateIndexerCaps", ms("1 day"), updateCaps));
 	}
 	return jobs;
 }
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -498,9 +498,8 @@ function collateOutcomes<Correlator, SuccessReturnType>(
 	);
 }
 
-async function updateCaps(
-	indexers: { id: number; url: string; apikey: string }[],
-): Promise<void> {
+export async function updateCaps(): Promise<void> {
+	const indexers = await getAllIndexers();
 	const outcomes = await Promise.allSettled<Caps>(
 		indexers.map((indexer) => fetchCaps(indexer)),
 	);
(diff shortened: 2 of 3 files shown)


src/server.ts
```diff
@@ -108,6 +109,7 @@ async function search(
 		message: `Received search request: ${criteriaStr}`,
 	});
 
+	await refreshIndexerCaps();
```
mmgoodnow, line 112: this feels a bit heavy… let's keep an eye on it. Caps basically never change so I think we could probably get away with something like a weekly or even less frequent job to update caps

src/torznab.ts
```diff
@@ -522,6 +522,20 @@ async function updateCaps(
...
+			indexer.searchCap === null ||
+			indexer.tvSearchCap === null ||
+			indexer.movieSearchCap === null ||
+			indexer.tvIdCaps === null ||
+			indexer.movieIdCaps === null ||
+			indexer.categories === null,
+	);
+	updateCaps(indexersWithoutCaps); // Will display errors if any
+}
+
```
mmgoodnow, line 538: why do these separately?

mmgoodnow (approved):

src/server.ts
zakkarry: I was thinking the same thing, given they don't update or change.

src/torznab.ts
zakkarry: I'm curious if there is ever a situation where it wouldn't be an "all or nothing" - missing just one seems unlikely?

ShanaryS: `updateCaps(indexersWithoutCaps)` is what will fix the null caps once the indexer becomes available. This will log an error message for each indexer that can't be reached every time it's ran, which I feel is desirable.

`updateCaps(await getEnabledIndexers())` checks for the unlikely scenario that caps have been updated. No error message will ever come from this unless the indexer just failed.

If `getAllIndexers()` were to be used instead of splitting them up, any rate limited indexers would display the error, even if the caps aren't null. Which I feel is misleading.

src/server.ts
ShanaryS: Prowlarr checks daily for indexer updates. I think that would be a good rate if doing it by job.

zakkarry (approved):

src/server.ts
zakkarry: Agreed

src/jobs.ts
```diff
@@ -41,6 +42,7 @@ function getJobs(): Job[] {
 	if (torznab.length > 0) {
 		if (rssCadence) jobs.push(new Job("rss", rssCadence, scanRssFeeds));
 		if (searchCadence) jobs.push(new Job("search", searchCadence, main));
+		jobs.push(new Job("indexer", ms("1 day"), updateCaps));
```
mmgoodnow, line 45: can this job's name be more specific? updateIndexerCaps maybe?

zakkarry (approved):

mmgoodnow (approved):

Issue #711, opened 2024-06-10
Deluge API is temperamental resulting in incorrectly `skipping incomplete torrent` 


RogueOneEcho:
### Have you checked for any existing open/closed tickets that are the same or similar to your issue?

- [X] I've searched far and wide!

### I have gone to the Discord server for support and/or acknowledge that this is a bug and not a support issue

- [X] I've tried to solve this already

### What's happening surrounding the bug? (context)

When executing a search with cross-seed using **deluge** the logs are filled with warnings of `skipping incomplete torrent` despite these torrents being 100% downloaded and seeding.

This occurs for both `v5.9.2` and `v6.0.0-10`.

### Reproducing the bug

Following these steps to manually [inspect the deluge API response with CURL](https://deluge.readthedocs.io/en/latest/devguide/how-to/curl-jsonrpc.html)

```bash
curl -d '{"method": "web.update_ui", "params": [["state", "progress", "save_path", "label"], {"hash": "**TORRENT HASH**"}], "id": 1}' -K curl.cfg | jq '.result.torrents'
```

Shows the  API  will occasionally response with `state` and `progress` omitted.

```json
{
  "**TORRENT HASH**": {
    "save_path": "**PATH**",
    "label": "**LABEL**"
  }
}
```

Repeating the request will eventually respond:

```json
{
  "**TORRENT HASH**": {
    "save_path": "**PATH**",
    "label": "**LABEL**",
    "state": "Seeding",
    "progress": 100
  }
}
```

As these fields are absent this causes cross-seed to assume the torrent is incomplete.

https://github.com/cross-seed/cross-seed/blob/0ab8741a5fdbab23db8c743d6ba8a6152ae2c460/src/clients/Deluge.ts#L470-L471

### What should be happening?

I would suggest that if those fields are absent from the response then cross-seed could attempt to retry the API request a few times, if still no luck then it should report "unable to torrent status" or similar which would be more accurate than incorrectly warning `skipping incomplete torrent`

### Your setup

```markdown
- OS: Ubuntu 20.04 LTS
- Docker: Yes (ghcr.io/cross-seed/cross-seed:master)
- Deluge: v2.1.1 https://fleet.linuxserver.io/image?name=linuxserver/deluge
```

### config.js (REDACT SENSITIVE INFORMATION)

N/A

### verbose logs

No additional verbose logs other than repetitions of:

[...]

PR #715, opened 2024-06-11
update(client/deluge): broaden indicators determining completion in client

zakkarry:
this accepts paused torrents if either marked with progress 100% or no remaining download, as well as adds total_remaining to the acceptable completion indicators

this may not entirely solve #711 but may increase likelihood that it is not show-stopping


mmgoodnow (commented):
I'm suspicious that this fixes it, but it seems fine to try.

mmgoodnow (approved):

zakkarry:
> I'm suspicious that this fixes it, but it seems fine to try.

Yea, I'm not convinced it will, but it can't hurt. 

As I mentioned in #711 I'm currently testing a theory, but replicating the issue seems rather evasive.

PR #723, opened 2024-06-14
enhancement(clients/qbit): better handling of post-injection verification

ShanaryS:
With lots of torrents, qbit can sometimes be slow to respond. 3s should catch pretty much all cases but added a failsafe.

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -7,6 +7,7 @@ import {
 } from "../constants.js";
 import { CrossSeedError } from "../errors.js";
 import { Label, logger } from "../logger.js";
+import ms from "ms";
 import { Metafile } from "../parseTorrent.js";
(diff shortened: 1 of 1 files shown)


zakkarry (dismissed):

mmgoodnow (commented):
I don't love the idea of an unconditional 3 second delay. If you look at rTorrent it uses an exponential backoff approach - what would you think of that? 

Also, what is the purpose of waiting and checking for the torrent - is it just to make sure?

ShanaryS:
> I don't love the idea of an unconditional 3 second delay. If you look at rTorrent it uses an exponential backoff approach - what would you think of that?

I'll switch that

> Also, what is the purpose of waiting and checking for the torrent - is it just to make sure?
Qbit is very weird when things are added, just to be safe. Previously at a 1s wait I've had a few that threw the error.

mmgoodnow (approved):

PR #724, opened 2024-06-14
enhancement(logging/search): track searchee source using labels to clarify injections

ShanaryS:


zakkarry (dismissed):

src/pipeline.ts
```diff
@@ -254,11 +259,9 @@ export async function checkNewCandidateMatch(
 		searchee,
 		candidate.tracker,
 	);
-	await sendResultsNotification(
-		searchee,
-		[[assessment, candidate.tracker, result]],
-		Label.REVERSE_LOOKUP,
-	);
+	sendResultsNotification(searchee, [
```
zakkarry, line 262: curious as to why `await` was removed here

ShanaryS: The function is not async.

zakkarry (dismissed):

PR #725, opened 2024-06-14
enhancement(reverselookup): using regex to parse rss/announce names instead of Fuse

ShanaryS:
This replaces the Fuse method of reverse lookup to leveraging the existing regex. Only the title, season, episode is used. All other parts should be handled in the decide stage. The title is cleansed as much as possible then converted to lowercase and with symbols and spaces removed. This is already much better than the Fuse, but we then add a fuzzy using the levenshtein distance. The max distance is `Math.floor(cleanTitle.length / 4)` so short titles needs to be exact and longer titles get more leeway. I haven't come across a situation where we need this fuzzy, but I'm sure they exist.

This method vastly outperforms Fuse in all aspects.
1. It's much faster. It's always ~50ms to get all matching items from the database for me. Where Fuse took 150ms if nothing was found up to multiple seconds if items were.
2. It has far less false positives that Fuse. Fuse would erroneously move to decide on at least 1/20 announces for me, I haven't seen a single false positive for the new method. They do exist, for example I'm not using the year in the key for movies since it can be wrong, but of course it's still far better than before.
3. Better true positives. Due to all the parsing, it handles weird naming formats much better and doesn't get tripped up by spaces or dots like Fuse.
4. A not insignificant benefit is that this is much more "deterministic". You can easily make improvements and logically reason about the algorithm. Where as the previous method was essentially a black box with a few parameters to tweak.

Due to how much more robust this is, we now create searchees from all matches rather than just the first. These are ordered by most files to reduce the chance of partials. I'm also using the improved dupe check originally from the `cache-searche-queries` branch.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,11 +40,11 @@ const SOURCE_REGEXES = {
 	NF: /\b(nf|netflix(u?hd)?)\b/i,
 	HULU: /\b(hulu)\b/i,
 	ATVP: /\b(atvp|aptv)\b/i,
-	HBO: /\b(hbo)(?![ ._-]max)\b|\b(hmax|hbom|hbo[ ._-]max)\b|\b((?<!hbo[ ._-])max)\b/i,
-	PCOK: /\b(pcok|peacock)\b/i,
+	HBO: /\b(hbo)(?![ ._-]max)\b|\b(hmax|hbom|hbo[ ._-]max)\b/i,
+	PCOK: /\b(pcok)\b/i,
 	PMTP: /\b(pmtp|Paramount Plus)\b/i,
 };
-export function sourceRegexParse(title: string): string | null {
+export function parseSource(title: string): string | null {
 	for (const [source, regex] of Object.entries(SOURCE_REGEXES)) {
 		if (regex.test(title)) return source;
 	}
@@ -53,8 +53,8 @@ export function sourceRegexParse(title: string): string | null {
 export function sourceRegexRemove(title: string): string {
 	const originalLength = title.length;
 	for (const regex of Object.values(SOURCE_REGEXES)) {
-		title = title.replace(regex, "");
-		if (title.length !== originalLength) return title;
+		const newTitle = title.replace(regex, "");
+		if (newTitle.length !== originalLength) return newTitle;
 	}
 	return title;
 }
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -22,18 +22,42 @@ export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w .]+?)(?:\[.+\])?(?:\))?(?:\s\[.+\])?$/i;
 export const RESOLUTION_REGEX = /\b(?<res>\d{3,4}[pix](?:\d{3,4}[pi]?)?)\b/i;
 export const RES_STRICT_REGEX = /(?<res>(?:2160|1080|720)[pi])/;
-
+export const YEAR_REGEX = /(?<year>(?:19|20)\d{2})(?![pix])/i;
 export const REPACK_PROPER_REGEX =
 	/(?:\b(?<type>(?:REPACK|PROPER|\d\v\d)\d?))\b/i;
-
 export const ARR_PROPER_REGEX = /(?:\b(?<arrtype>(?:Proper|v\d)))\b/;
-
 export const SCENE_TITLE_REGEX = /^(?:[a-z0-9]{0,5}-)?(?<title>.*)/;
-
 export const ARR_DIR_REGEX =
 	/^(?<title>(?!.*(?:(\d{3,4}[ipx])|([xh.]+26[4-6])|(mpeg)|(xvid)|(?:(he)|a)vc))[\p{L}\s:\w'’!();.,&–+-]+(?:\(\d{4}\))?)(?<id>\s[{[](?:tm|tv|im)db(?:id)?-\w+?[}\]])?$/iu;
 export const SONARR_SUBFOLDERS_REGEX =
 	/^(?:S(?:eason )?(?<seasonNum>\d{1,4}))$/i;
+export const NON_UNICODE_ALPHANUM_REGEX = /[^\p{L}\p{N}]+/giu;
+
+// Needs to be hanlded through helper functions since there is varitions
+const SOURCE_REGEXES = {
+	AMZN: /\b(amzn|amazon(hd)?)\b[ ._-]web[ ._-]?(dl|rip)?\b/i,
+	DSNP: /\b(dsnp|dsny|disney)\b/i,
+	NF: /\b(nf|netflix(u?hd)?)\b/i,
+	HULU: /\b(hulu)\b/i,
+	ATVP: /\b(atvp|aptv)\b/i,
+	HBO: /\b(hbo)(?![ ._-]max)\b|\b(hmax|hbom|hbo[ ._-]max)\b|\b((?<!hbo[ ._-])max)\b/i,
+	PCOK: /\b(pcok|peacock)\b/i,
+	PMTP: /\b(pmtp|Paramount Plus)\b/i,
+};
+export function sourceRegexParse(title: string): string | null {
+	for (const [source, regex] of Object.entries(SOURCE_REGEXES)) {
+		if (regex.test(title)) return source;
+	}
+	return null;
+}
+export function sourceRegexRemove(title: string): string {
+	const originalLength = title.length;
+	for (const regex of Object.values(SOURCE_REGEXES)) {
+		title = title.replace(regex, "");
+		if (title.length !== originalLength) return title;
+	}
+	return title;
+}
 
 export const VIDEO_EXTENSIONS = [".mkv", ".mp4", ".avi", ".ts"];
 export const AUDIO_EXTENSIONS = [
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -33,7 +33,7 @@ export const SONARR_SUBFOLDERS_REGEX =
 	/^(?:S(?:eason )?(?<seasonNum>\d{1,4}))$/i;
 export const NON_UNICODE_ALPHANUM_REGEX = /[^\p{L}\p{N}]+/giu;
 
-// Needs to be hanlded through helper functions since there is varitions
+// Needs to be handled through helper functions since there are variations
 const SOURCE_REGEXES = {
 	AMZN: /\b(amzn|amazon(hd)?)\b[ ._-]web[ ._-]?(dl|rip)?\b/i,
 	DSNP: /\b(dsnp|dsny|disney)\b/i,
--- a/src/preFilter.ts
+++ b/src/preFilter.ts
@@ -106,11 +106,12 @@ export function findBlockedStringInReleaseMaybe(
 }
 
 /**
- * Filters duplicates from searchees that should be for different candidates.
+ * Filters duplicates from searchees that should be for different candidates,
+ * e.g. all searchees created by cross-seed.
  * @param searchees - An array of searchees to filter duplicates from.
- * @returns An array of searchees with duplicates removed.
+ * @returns An array of searchees with duplicates removed, preferring infoHash.
  */
-export function filterDupes(searchees: Searchee[]): Searchee[] {
+export function filterDupesByName<T extends Searchee>(searchees: T[]): T[] {
 	const duplicateMap = searchees.reduce((acc, cur) => {
 		const entry = acc.get(cur.name);
 		if (entry === undefined) {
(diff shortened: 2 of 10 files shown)


src/constants.ts
```diff
@@ -21,17 +21,41 @@ export const RELEASE_GROUP_REGEX =
 	/(?<=-)(?:\W|\b)(?!(?:\d{3,4}[ip]))(?!\d+\b)(?:\W|\b)(?<group>[\w .]+?)(?:\[.+\])?(?:\))?(?:\s\[.+\])?$/i;
 export const RESOLUTION_REGEX = /\b(?<res>\d{3,4}[pix](?:\d{3,4}[pi]?)?)\b/i;
 export const RES_STRICT_REGEX = /(?<res>(?:2160|1080|720)[pi])/;
-
+export const YEAR_REGEX = /(?<year>(?:19|20)\d{2})/;
```
zakkarry, line 24: Could this erroneously match anime that has "1920x1080" in the title? I believe you said this was something done in anime in a previous conversation...

src/constants.ts
```diff
@@ -21,17 +21,41 @@ export const RELEASE_GROUP_REGEX =
...
 export const SCENE_TITLE_REGEX = /^(?:\w{0,5}-)?(?<title>.*)/i;
-
 export const ARR_DIR_REGEX =
 	/^(?!.*(?:(\d{3,4}[ipx])|([xh]26[4-6])|(?:(he)|a)vc))[\p{L}\s:\w'’!().,&–+-]+(?:\(\d{4}\))?\s?(?:\{(?:tm|tv|im)db(?:id)?-\w+\})?$/iu;
 export const SONARR_SUBFOLDERS_REGEX = /^(?:S(?:eason )?(?<season>\d{1,4}))$/i;
+export const NON_UNICODE_LETTER_NUM_REGEX = /[^\p{L}\p{N}]+/giu;
+
+// Needs to be hanlded through helper functions since there is varitions
+const SOURCE_REGEXES = {
+	AMZN: /\b(amzn|amazon(hd)?)\b/i,
```
zakkarry, line 36: To avoid matching `amazon` in episode titles or series names, we've got an open PR with a fix on trash's repo....

https://github.com/TRaSH-Guides/Guides/pull/1950/files

src/constants.ts
```diff
@@ -21,17 +21,41 @@ export const RELEASE_GROUP_REGEX =
...
+	DSNP: /\b(dsnp|dsny|disney)\b/i,
+	NF: /\b(nf|netflix(u?hd)?)\b/i,
+	HULU: /\b(hulu)\b/i,
+	ATVP: /\b(atvp|aptv)\b/i,
+	HBO: /\b(hbo)(?![ ._-]max)\b|\b(hmax|hbom|hbo[ ._-]max)\b|\b((?<!hbo[ ._-])max)\b/i,
+	PCOK: /\b(pcok|peacock)\b/i,
```
zakkarry, line 42: same as above could potentially apply to peacock or max - dictionary words commonly found in episode or series titles...

src/constants.ts
```diff
@@ -21,17 +21,41 @@ export const RELEASE_GROUP_REGEX =
...
 	/(?:\b(?<type>(?:REPACK|PROPER|\d\v\d)\d?))\b/i;
-
 export const ARR_PROPER_REGEX = /(?:\b(?<arrtype>(?:Proper|v\d)))\b/;
-
 export const SCENE_TITLE_REGEX = /^(?:\w{0,5}-)?(?<title>.*)/i;
-
 export const ARR_DIR_REGEX =
 	/^(?!.*(?:(\d{3,4}[ipx])|([xh]26[4-6])|(?:(he)|a)vc))[\p{L}\s:\w'’!().,&–+-]+(?:\(\d{4}\))?\s?(?:\{(?:tm|tv|im)db(?:id)?-\w+\})?$/iu;
 export const SONARR_SUBFOLDERS_REGEX = /^(?:S(?:eason )?(?<season>\d{1,4}))$/i;
+export const NON_UNICODE_LETTER_NUM_REGEX = /[^\p{L}\p{N}]+/giu;
```
zakkarry, line 32: no real opinion on this, but maybe something like `NON_UNICODE_ALPHANUM_REGEX` would be easier to understand?

zakkarry (commented):
just a few comments right now

src/constants.ts
ShanaryS: It actually doesn't matter for our use case too much. I tried to order the safer ones first which should catch most. Otherwise the chance of it happening is low. Even if it does, in safe mode it won't matter since they will be both be wrong but equal. And it's unlikely for something to get renamed that would remove one of these words for risky/partial. Same idea extends to ensemble.

We could also remove the problematic ones since which might be the best case. Unlikely the same release group gets renamed their source renamed between trackers.

src/searchee.ts
```diff
@@ -160,3 +172,106 @@ export async function createSearcheeFromPath(
 		length: totalLength,
 	});
 }
+
+export function getKeyMetaInfo(stem: string, isAnime: boolean): string {
+	const resM = stem.match(RES_STRICT_REGEX)?.groups?.res;
+	const res = resM ? `.${resM}` : "";
+	const sourceM = sourceRegexParse(stem);
```
mmgoodnow, line 179: it looks like these are only going to find WEB-DLs - what about BluRays or other sources like HDTV?

src/searchee.ts
```diff
@@ -160,3 +172,106 @@ export async function createSearcheeFromPath(
...
+	if (!keyTitle.length) return null;
+	const seasonM = match!.groups!.season;
+	const season = seasonM
+		? `S${extractInt(seasonM)}`
+		: `S${match!.groups!.year}`;
+	const ensembleTitle = `${match!.groups!.title}.${season}`;
+	const episode = seasonM
+		? extractInt(match!.groups!.episode)
+		: `${match!.groups!.month}/${match!.groups!.day}`;
+	return { ensembleTitle, keyTitle, season, episode };
```
mmgoodnow, line 249: why would the ensembleTitle be the same for seasons and episodes?

src/searchee.ts
```diff
@@ -160,3 +172,106 @@ export async function createSearcheeFromPath(
...
+		.replace(YEAR_REGEX, "")
+		.replace(NON_UNICODE_ALPHANUM_REGEX, "")
+		.toLowerCase();
+	if (!keyTitle.length) return null;
+	const season = `S${extractInt(match.groups!.season)}`;
+	const ensembleTitle = `${match.groups!.title}.${season}`;
+	return { ensembleTitle, keyTitle, season };
+}
+
+export function getEpisodeAndKey(stem: string): {
```
mmgoodnow, line 228: episode and what?

src/searchee.ts
```diff
@@ -160,3 +172,106 @@ export async function createSearcheeFromPath(
...
+		? `S${extractInt(seasonM)}`
+		: `S${match!.groups!.year}`;
+	const ensembleTitle = `${match!.groups!.title}.${season}`;
+	const episode = seasonM
+		? extractInt(match!.groups!.episode)
+		: `${match!.groups!.month}/${match!.groups!.day}`;
+	return { ensembleTitle, keyTitle, season, episode };
+}
+
+export function getReleaseAndKeys(stem: string): {
```
mmgoodnow, line 252: is this only applicable to anime?

src/torrent.ts
```diff
@@ -219,22 +227,89 @@ export async function loadTorrentDirLight(
...
+	}
+	return { keyTitles: [], useFallback: true };
+}
+
+export async function getTorrentByName(
 	name: string,
-): Promise<null | Metafile> {
+): Promise<{ keys: string[]; metas: Metafile[] }> {
+	const { keyTitles, element, useFallback } = getKeysFromName(name);
+	const metas = useFallback ? await getTorrentByFuzzyName(name) : [];
```
mmgoodnow, line 266: how valuable is the fallback? it seems like this logic is pretty comprehensive. I'd be in favor of deleting the fuse code.

src/torrent.ts
```diff
@@ -219,22 +227,89 @@ export async function loadTorrentDirLight(
...
+	const filteredEntries = allEntries.filter((dbName) => {
+		const entry = getKeysFromName(dbName.name);
+		if (entry.element !== element) return false;
+		return entry.keyTitles.some((dbKeyTitle) => {
+			return keyTitles.some(
+				(keyTitle) => distance(keyTitle, dbKeyTitle) <= maxDistance,
+			);
+		});
+	});
+	if (filteredEntries.length === 0) return { keys, metas };
```
mmgoodnow, line 287: looks like this line doesn't do anything

mmgoodnow (commented):
The code looks pretty good. I'm happy to see a lot of domain-specific knowledge getting encoded as code to make these things more accurate.

src/preFilter.ts
```diff
@@ -105,6 +105,11 @@ export function findBlockedStringInReleaseMaybe(
 	});
 }
 
+/**
+ * Filters duplicates from searchees that should be for different candidates.
```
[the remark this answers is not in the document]
ShanaryS: `filterDupes` (now `filterDupesByName`) isn't used anymore after `cache-search-queries` is merged. But the use case is going over all searchees that `cross-seed` finds preferring infohash.

`filterDupesFromSimilar` is used when we know all the searchees are for the same media. So same movie, but they could be true duplicates (cross seeds) or be different resolutions, source, etc. This removes the searchees that are cross seeds of each other, leaving only one. This means that all that survive can have unique decisions.

I updated the docstring to make the uses clearer.

src/searchee.ts
ShanaryS: The main reason `SOURCE_REGEX` was added is for ensemble aggregating episodes. Individual episodes are unlikely to be bluray, that can could only be a negative. I think HDTV|SDTV could be added but those are rare and can be missing from names so I ignored them like we do for resolutions below 720p.

`SOURCE_REGEX` is being used to reject in `decide.js`, but that's just because it's available. I rather the extra snatches than risk false negatives. It will be caught by fuzzy size and release group anyways since these highly affected by the source.

src/searchee.ts
ShanaryS: It's hard to name. The key is the unique identifier to use if something matches. I separated the info as ensemble uses this too and parses it differently.

src/searchee.ts
ShanaryS: ensembleTitle is only used in the `ensemble-matching` branch to give a good name for the virtual searchee. From rss/announce, we get the title from `getSeasonKey(candidate.name)`. When searching, we get the title from `getEpisodeAndKey(searchee.name)` to name the virtual season searchee.

src/searchee.ts
ShanaryS: `ANIME_REGEX` matches pretty much anything since anime naming is very loose. If used on anime after the other regexes fail, release is the episode. This is also the only one that could have multiple titles.

src/torrent.ts
ShanaryS: The `ANIME_REGEX` cannot be as complete as the others due to anime naming. So we always use Fuse additionally.

Fuse is used exclusively if the media doesn't match one of our regexes. This is to support non-video matching from rss/announce.

src/utils.ts
```diff
@@ -131,22 +135,22 @@ export function getAnimeQueries(name: string): string[] {
...
-export function reformatTitleForSearching(name: string): string {
-	// use lazy regex evaluation
-	const fullMatch =
-		name.match(EP_REGEX)?.[0] ??
-		name.match(SEASON_REGEX)?.[0] ??
-		name.match(MOVIE_REGEX)?.[0] ??
-		name;
+export function reformatTitleForSearching(
+	name: string,
+	useRegex = true,
```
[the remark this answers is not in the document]
ShanaryS: I wanted to keep these here so that logic is not duplicated. Any changes we make here for non id searching should also apply to rss, announce, and ensemble.

src/searchee.ts
mmgoodnow: the ones above seem to follow a convention - `getSeasonKey` `getMovieKey` etc, so with this one being `getEpisodeAndKey` it seems like there should be another word in between `And` and `Key`. Are these ones just different from each other?

src/utils.ts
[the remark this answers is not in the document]
mmgoodnow: We can keep the logic in one place with a separate function too, by calling it inside this one.

src/constants.ts
```diff
@@ -22,18 +22,42 @@ export const RELEASE_GROUP_REGEX =
...
-
 export const SCENE_TITLE_REGEX = /^(?:[a-z0-9]{0,5}-)?(?<title>.*)/;
-
 export const ARR_DIR_REGEX =
 	/^(?<title>(?!.*(?:(\d{3,4}[ipx])|([xh.]+26[4-6])|(mpeg)|(xvid)|(?:(he)|a)vc))[\p{L}\s:\w'’!();.,&–+-]+(?:\(\d{4}\))?)(?<id>\s[{[](?:tm|tv|im)db(?:id)?-\w+?[}\]])?$/iu;
 export const SONARR_SUBFOLDERS_REGEX =
 	/^(?:S(?:eason )?(?<seasonNum>\d{1,4}))$/i;
+export const NON_UNICODE_ALPHANUM_REGEX = /[^\p{L}\p{N}]+/giu;
+
+// Needs to be hanlded through helper functions since there is varitions
```
mmgoodnow, line 36:
```suggestion
// Needs to be hanlded through helper functions since there are variations
```

src/searchee.ts
ShanaryS: I see it now. They were previously named like that before their logic changed.

src/searchee.ts
ShanaryS: Changed to `getAnimeKeys`

mmgoodnow (approved):
LGTM

PR #733, opened 2024-06-21
Add support for basic auth

frenchcutgreenbean:
Access auth protected instances of your torznab apps. Helpful if you already have prowlarr setup on a seedbox / server and want to quickly use cross-seed on a separate machine.

Tested with a remote instance of prowlarr to cross-seed on my local machine.

user:pass is extracted from the URL in the config then converted to btoa format and stored in the database under 'basicauth' column.

I couldn't figure out how to add the new column through code but running this in cross-seed.db worked
"ALTER TABLE indexer ADD COLUMN basicauth TEXT;"

### DIFF
--- a/src/indexers.ts
+++ b/src/indexers.ts
@@ -15,6 +15,7 @@ export interface Indexer {
 	id: number;
 	url: string;
 	apikey: string;
+	basicauth: string;
 	/**
 	 * Whether the indexer is currently specified in config
 	 */
@@ -34,6 +35,7 @@ export async function getAllIndexers(): Promise<Indexer[]> {
 		id: "id",
 		url: "url",
 		apikey: "apikey",
+		basicauth: "basicauth",
 		active: "active",
 		status: "status",
 		retryAfter: "retry_after",
@@ -67,6 +69,7 @@ export async function getEnabledIndexers(): Promise<Indexer[]> {
 			id: "id",
 			url: "url",
 			apikey: "apikey",
+			basicauth: "basicauth",
 			active: "active",
 			status: "status",
 			retryAfter: "retry_after",
--- a/src/migrations/02-timestamps.ts
+++ b/src/migrations/02-timestamps.ts
@@ -1,18 +1,8 @@
-import Knex from "knex";
+import { Knex } from "knex";
 import { getRuntimeConfig } from "../runtimeConfig.js";
+import { sanitizeUrl, getApikey, getBasicAuth } from "../utils.js";
 
-function sanitizeUrl(url: string | URL): string {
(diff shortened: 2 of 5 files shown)


bakerboy448:
This _really_ isn't needed as it's stupid for users to double auth their Torznab apis....the apikey of the Torznab app protects itself....no additional proxy should be setup in front of the api routes....believe there is only 1 seedbox provider who for several years refuses to fix their proxy configuration and allow external prowlarr usage

zakkarry:
I'd tend to agree, requiring basic auth for a path that is strictly API related and has authentication natively is sort of dumb.

Furthermore, the arr's as a whole require authentication prior to accessing outside of local networks...in fact I think the majority of all full stacks in this space require it.

I'm not entirely sure what benefit you gain that you wouldn't otherwise have, or what setup would _necessitate_ basic auth.

If you have a specific reason for this feature being needed, I'd be open to hearing it out though.

These are just my opinions, btw.

zakkarry:
Also you should add an additional migration, although I'd argue that the basic auth information should be kept out of the database entirely and read from the config - database shouldn't need to be modified for anything.

frenchcutgreenbean:
@bakerboy448 
I guess I am one of the ones with the bad seedbox provider 😭. 


This was created based on the assumption that all seedbox providers require basic auth to access arr APIs. 
If that is not the case I would agree that this probably is not something worth adding at all.

>I'm not entirely sure what benefit you gain that you wouldn't otherwise have, or what setup would necessitate basic auth.
If you have a specific reason for this feature being needed, I'd be open to hearing it out though.

My seedbox provider requires it, but if this isn't a common occurrence it's definitely not a needed feature.

bakerboy448:
Name and shame?

There's like ~2 providers that seemingly refuse to fix their reverse proxies since prowlarr was developed a few years ago.

zakkarry:
I'd like to know which providers, we could perhaps attempt to speak with them directly (again), however the premise seems flawed as I described in my previous post.

They are doing this for no apparent reason, and exempting the URI path seems pretty trivial....especially for someone running an entire seedbox provider.

frenchcutgreenbean:
hostingby.design is mine

I am not very knowledgeable but I would guess it comes down to how they chose to route? 
They expose the arrs like this https://shared-box.com/prowlarr which would be a common url between all users instead of something like this ...com/username/prowlarr

So I am guessing with a shared URL there is no other way to distinguish the instance.

bakerboy448:
It's been mentioned that HBD users can bypass the basic auth by accessing it on the subnet/port using Wireguard.

PR #749, opened 2024-07-18
update(config): add config restrictions when using daemon mode

ShanaryS:
These serves to prevent users from configuring `cross-seed` in a way harmful to themselves or the ecosystem.

Using webhook bypasses these restrictions and always ignores `excludeRecentSearch` and `excludeOlder`. Effectively, webhook searches are always valid from a timestamp perspective.

Users can use `cross-seed search` to ignore these restrictions as well. But this is discouraged and mainly exist for any one off circumstances.


src/configSchema.ts
```diff
@@ -155,12 +167,21 @@ export const VALIDATION_SCHEMA = z
 			.string()
 			.min(1, { message: ZodErrorMessages.emptyString })
 			.transform(transformDurationString)
-			.nullish(),
+			.nullish()
+			.refine(
+				(cadence) =>
+					!cadence || (cadence >= 600000 && cadence <= 7200000),
```
mmgoodnow, line 173: can you use `ms` for these? they're hard to read

mmgoodnow (dismissed):

PR #750, opened 2024-07-19
feat(action/inject): resume injections below size threshold and readd skipRecheck

ShanaryS:
This offers `cross-seed` the highly requested ability to resume partial matches without user intervention. Also, we will now recheck and resume all non torrent based searchees as well since we can't tell if a data/virtual searchee is completed.

Readded the skipRecheck option since we can now resume torrents after rechecking.

The maximum that we will resume is `50MB` remaining. This was chosen as it should resume all normal partial matches while never resuming those from ensemble and for the `inject-saved-torrents` feature to only download missing data once between partial matches on different trackers. We do not need to check for fuzzy ratio as the decide stage already has for partial matches.

In the worst case, say a user has 100 partial matches for a single tracker, the maximum download from this would be `5GB`. This will never happen however as 50% of my partial matches are below `1MB`, 80% under `10MB` and 95% under `20MB`. And that's from testing with an aggressive `fuzzySizeThreshold` of `0.1` and including ensemble matches. So a user would likely need at least 1,000+ partials on a single tracker for `5GB` of `cross-seed` download.

This was implemented in an extremely safe way:
1. After injecting a partial, start a background function to check the torrent status
2. Check every `15s`. If it's in a rechecking state, repeat. Timeout of 1 hour.
3. If it's not paused, early exit.
4. If size remaining greater than max, early exit.
5. Resume otherwise.

To cover the case where `cross-seed` has restarted or the timeout elapsed, we also check with the hourly inject job and can resume from there.

### DIFF
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -14,9 +14,19 @@ import { Metafile } from "../parseTorrent.js";
 import { Result, resultOf, resultOfErr } from "../Result.js";
 import { getRuntimeConfig } from "../runtimeConfig.js";
 import { File, Searchee, SearcheeWithInfoHash } from "../searchee.js";
-import { extractCredentialsFromUrl, isTruthy, wait } from "../utils.js";
+import {
+	extractCredentialsFromUrl,
+	humanReadableSize,
+	isTruthy,
+	sanitizeInfoHash,
+	wait,
+} from "../utils.js";
 import {
 	TorrentMetadataInClient,
+	getMaxRemainingBytes,
+	getResumeStopTime,
+	resumeErrSleepTime,
+	resumeSleepTime,
 	shouldRecheck,
 	TorrentClient,
 	validateSavePaths,
@@ -468,6 +478,59 @@ export default class RTorrent implements TorrentClient {
 		await this.methodCallP<void>("d.check_hash", [infoHash]);
 	}
 
+	async resumeInjection(
+		infoHash: string,
+		options: { checkOnce: boolean },
+	): Promise<void> {
+		let sleepTime = resumeSleepTime;
+		const maxRemainingBytes = getMaxRemainingBytes();
+		const stopTime = getResumeStopTime();
+		let stop = false;
+		while (Date.now() < stopTime) {
+			if (options.checkOnce) {
+				if (stop) return;
+				stop = true;
+			}
+			await wait(sleepTime);
+			const torrentInfoRes = await this.checkOriginalTorrent(
+				infoHash,
+				false,
+			);
+			if (torrentInfoRes.isErr()) {
+				sleepTime = resumeErrSleepTime; // Dropping connections or restart
+				continue;
+			}
+			const torrentInfo = torrentInfoRes.unwrap();
+			if (torrentInfo.hashing) {
+				continue;
+			}
+			const torrentLog = `${torrentInfo.name} [${sanitizeInfoHash(infoHash)}]`;
+			if (torrentInfo.isActive) {
+				logger.warn({
+					label: Label.RTORRENT,
+					message: `Will not resume torrent ${torrentLog}: active`,
+				});
+				return;
+			}
+			if (torrentInfo.bytesLeft > maxRemainingBytes) {
+				logger.warn({
+					label: Label.RTORRENT,
+					message: `Will not resume torrent ${torrentLog}: ${humanReadableSize(torrentInfo.bytesLeft, { binary: true })} remaining > ${humanReadableSize(maxRemainingBytes, { binary: true })}`,
+				});
+				return;
+			}
+			logger.info({
+				label: Label.RTORRENT,
+				message: `Resuming torrent ${torrentLog}`,
+			});
+			await this.methodCallP<void>("d.resume", [infoHash]);
+		}
+		logger.warn({
+			label: Label.RTORRENT,
+			message: `Will not resume torrent ${infoHash}: timeout`,
+		});
+	}
+
 	async inject(
 		meta: Metafile,
 		searchee: Searchee,
@@ -500,8 +563,8 @@ export default class RTorrent implements TorrentClient {
 
 		await saveWithLibTorrentResume(meta, torrentFilePath, basePath);
 
-		const recheck = shouldRecheck(searchee, decision);
-		const loadType = recheck ? "load.normal" : "load.start";
+		const toRecheck = shouldRecheck(searchee, decision);
+		const loadType = toRecheck ? "load.normal" : "load.start";
 
 		const retries = 5;
 		for (let i = 0; i < retries; i++) {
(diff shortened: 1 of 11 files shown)


zakkarry:
Based on discussions I've had, I forsee some people will inherently not trust our code and judgment on resuming. What do you think about the idea of having a way to disable this functionality?

The most obvious solution is an option like `resumePartials` or something that is boolean, or perhaps a threshold with an upper limit?

I'm not sure if this is entirely possible to implement with a simple if, but it would seem to me that this is going to ruffle some feathers for the people who micromanage the hell out of their cross-seeds.

ShanaryS:
If they don't want it to be automatically resumed, then they shouldn't be using partial.

But if we are going to give the option, the option should be the amount resumed with an upper limit of `50MB`.

src/clients/RTorrent.ts
```diff
@@ -451,6 +461,60 @@ export default class RTorrent implements TorrentClient {
...
+				return;
+			}
+			if (torrentInfo.bytesLeft > maxRemainingBytes) {
+				logger.warn({
+					label: Label.RTORRENT,
+					message: `Will not resume torrent ${torrentLog}: ${humanReadableSize(torrentInfo.bytesLeft, { binary: true })} remaining > ${humanReadableSize(maxRemainingBytes, { binary: true })}`,
+				});
+				return;
+			}
+			logger.info({
```
mmgoodnow, line 506: maybe verbose for these? also fine to wait and see what people say

src/clients/RTorrent.ts
```diff
@@ -451,6 +461,60 @@ export default class RTorrent implements TorrentClient {
...
+		let sleepTime = resumeSleepTime;
+		const maxRemainingBytes = getMaxRemainingBytes(decision);
+		const stopTime = getResumeStopTime();
+		let stop = false;
+		while (Date.now() < stopTime) {
+			if (options.checkOnce) {
+				if (stop) return;
+				stop = true;
+			}
+			await wait(sleepTime);
```
mmgoodnow, line 478: these waits are going to be a bit wonky on run-once entrypoints like `cross-seed search` because they will create "holds" on the event loop which will make the program look like it's hanging. We could get around that by adding `.unref()` to the setTimeout call, but it would mean either the last 0-~1 torrents in a `cross-seed search` run don't get resumed, or we would turn it off entirely for `cross-seed search`. 

Having thought about this as I was writing the above, that stuff doesn't seem like a huge deal. But we also have less-than-ideal behavior on restart because we lose some of the program's state. Together these concerns kind of make me think—what would it take to run this as a job?

src/cmd.ts
```diff
@@ -123,6 +123,17 @@ function createCommandWithSharedOptions(name: string, description: string) {
 				.choices(Object.values(MatchMode))
 				.makeOptionMandatory(),
 		)
+		.option(
+			"--no-skip-recheck",
+			"Recheck every torrent before resuming. Leave this off to only recheck those that are necessary.",
+			fallback(fileConfig.skipRecheck, true),
+		)
+		.option(
+			"--max-remaining-for-resume <decimal>",
```
mmgoodnow, line 132: This name also seems a bit clunky, although I have been thinking about it for a bit and haven't come up with anything super satisfying. Playing with ideas like `maxDlToIncur` `maxDownload`  `maxToDownload` `autoResumeMaxDownload` feels like the word "download" might help the user understand what it means. Or maybe something like `autoResumeUnder`

mmgoodnow (commented):
Implementation looks fairly straightforward for the problem it solves. we should solidify the config interface before merging, but I think the larger question of whether or not it should be a job can can be more of a "try it and see" thing and we can iterate later.

src/cmd.ts
```diff
@@ -123,6 +123,17 @@ function createCommandWithSharedOptions(name: string, description: string) {
 				.choices(Object.values(MatchMode))
 				.makeOptionMandatory(),
 		)
+		.option(
+			"--no-skip-recheck",
+			"Recheck every torrent before resuming. Leave this off to only recheck those that are necessary.",
+			fallback(fileConfig.skipRecheck, true),
+		)
```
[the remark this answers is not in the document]
ShanaryS: I'm not too confident with these either, but it worked from my testing iirc. The goal is to default to `skipRecheck: true`. There is no way to prevent resuming, at 0 complete matches will still be resumed.

This option just allows the user to also recheck these first. In the unlikely case where this torrent was corrupted, it won't have 0 remaining and will not resume. In the future, I would like to delete these torrents but I feel that would make this PR too complicated for how rare it is.

The idea is that a user can prevent failures later on in client once the corrupted piece is finally requested by using `skipRecheck: false`. At the cost of incuring tons of rechecking.

src/clients/TorrentClient.ts
```diff
@@ -124,7 +130,23 @@ export function shouldRecheck(
...
+	const { matchMode, maxRemainingForResume } = getRuntimeConfig();
+	if (decision !== Decision.MATCH_PARTIAL) return 0;
+	if (matchMode !== MatchMode.PARTIAL) return 0;
+	return maxRemainingForResume * 1024 * 1024;
+}
+export const resumeSleepTime = ms("15 seconds");
+export const resumeErrSleepTime = ms("5 minutes");
+export function getResumeStopTime() {
+	return Date.now() + ms("1 hour");
+}
```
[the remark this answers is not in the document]
ShanaryS: All torrents that return true from `shouldRecheck` will have their .torrent files saved in `outputDir`. So these torrents will get checked for resume each inject job cycle.

src/clients/RTorrent.ts
ShanaryS: In the case of partial, I think it's worth logging in info. Especially since this hangs non-daemon mode.

src/clients/RTorrent.ts
ShanaryS: I think it's unnecessary to run as a job since it's already covered by the inject job. For `cross-seed search` and `cross-seed rss`, it will seem like it hangs but it most cases it should recheck within a couple minutes. Unless there is tons of partial matches being rechecked.

For the inject job though, there is a log stating it's waiting for tasks to finish.

src/cmd.ts
ShanaryS: I think `autoResume` is better than `maxRemaining` as it's more specific, the latter could be misinterpreted to affect getting matches. I think `autoResumeMaxDownload` is best.

src/cmd.ts
```diff
@@ -123,6 +123,17 @@ function createCommandWithSharedOptions(name: string, description: string) {
...
+		.option(
+			"--max-remaining-for-resume <decimal>",
+			"The maximum size in MB remaining for a torrent to be resumed",
```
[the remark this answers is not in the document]
ShanaryS: I think bytes would be better then simply for fewer dependencies. It will default to `50 MiB` and I doubt users would lower it.

src/clients/TorrentClient.ts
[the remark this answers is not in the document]
ShanaryS: Note that it will only check once in the case of the inject job, it will not use the 1 hour timeout.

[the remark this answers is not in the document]
mmgoodnow: Ah, I see.

mmgoodnow (approved):

PR #751, opened 2024-07-19
update(config/template): descriptions for time-based options

zakkarry:
this PR clarifies the new restrictions implemented in the config validation and explains the ratio relationship between time-based values


ShanaryS (approved):

mmgoodnow (dismissed):
you used "needs to be", "should always be", and "should be" for the three different changes you made here. I'd recommend standardizing on language like MUST and SHOULD like they do in the torznab spec. I don't think you need to add a legend but just using clear words will help

 https://torznab.github.io/spec-1.3-draft/torznab/Specification-v1.3.html#terminology

mmgoodnow (approved):

PR #753, opened 2024-07-19
fix(action/linking): resolve symlinks prior to linking

zakkarry:
Users of `linkType: "symlink"` may have gotten themselves into a situation where they have the following, due to cross-seed picking different "representative searchees" of a group of matches over time. 

```
symlink A -> symlink B -> original
```

It is then confusing and hard to manage the files when they want to delete symlink B - even though the original still exists, symlink A is broken. 

This PR adds logic to resolve symlinks to the original file to prevent this. It's more similar to hardlinking behavior, except the original still can't be deleted.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -373,9 +373,7 @@ export async function performActions(searchee, matches) {
 function linkFile(oldPath: string, newPath: string) {
 	const { linkType } = getRuntimeConfig();
 	try {
-		const ogFileResolvedPath = lstatSync(oldPath).isSymbolicLink()
-			? findSymlinksRealFile(oldPath)
-			: oldPath;
+		const ogFileResolvedPath = unwrapSymlinks(oldPath);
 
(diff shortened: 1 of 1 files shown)


src/action.ts
```diff
@@ -371,12 +373,21 @@ export async function performActions(searchee, matches) {
...
 	const { linkType } = getRuntimeConfig();
 	try {
+		let originalResolvedPath;
+		const lstats = lstatSync(oldPath);
+
+		if (lstats.isSymbolicLink()) {
+			originalResolvedPath = realpathSync(oldPath);
+		} else {
+			originalResolvedPath = oldPath;
+		}
```
ShanaryS, line 383:
```suggestion
		const originalResolvedPath = lstatSync(oldPath).isSymbolicLink()
			? realpathSync(oldPath)
			: oldPath;
```

mmgoodnow (approved):

PR #755, opened 2024-07-23
update(logging/checks): adds timeout on cap fetching and validates webhook params

zakkarry:
there is no timeout set on fetchCaps - which depending on the number of indexers can delay config validation for what seems like indefinitely to a user with prowlarr/jackett not responding. this introduces a 10s timeout on cap fetching, which should be almost instant minus latency between cross-seed and prowlarr/jackett, as the indexer is not queried for this.

also added some sanity checking on webhook parameters, checks for correct length of infohash and the validity of a path on the fs before proceeding

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -11,6 +11,7 @@ import {
 } from "./pipeline.js";
 import { InjectionResult, SaveResult } from "./constants.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
+import { existsSync } from "fs";
 
 function getData(req: IncomingMessage): Promise<string> {
 	return new Promise((resolve) => {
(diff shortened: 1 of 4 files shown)


ShanaryS (dismissed):

src/server.ts
```diff
@@ -90,9 +91,16 @@ async function search(
...
 
-	if (!("infoHash" in criteria || "path" in criteria)) {
+	if (
+		!(
+			(criteria.infoHash && criteria.infoHash.length === 40) ||
+			(criteria.path &&
+				criteria.path.length > 0 &&
+				existsSync(criteria.path))
+		)
+	) {
```
mmgoodnow, line 101: should we just stick zod on this?

src/torznab.ts
```diff
@@ -523,8 +523,14 @@ async function fetchCaps(indexer: {
...
 	let response;
 	try {
+		const abortController = new AbortController();
+		setTimeout(
+			() => void abortController.abort(),
+			ms("10 seconds"),
+		).unref();
 		response = await fetch(
 			assembleUrl(indexer.url, indexer.apikey, { t: "caps" }),
+			{ signal: abortController.signal },
```
mmgoodnow, line 533: I learned about `AbortSignal.timeout()` recently. It's a lot simpler than the whole AbortController dance. 

```suggestion
			{ signal: AbortSignal.timeout(ms("10 seconds"))},
```

mmgoodnow (dismissed):

src/server.ts
zakkarry: I'm not quite sure what you mean?

PR #756, opened 2024-07-23
fix(torznab-validation): checks hosts for connectivity before validating

zakkarry:
this checks all unique hostnames from the validated torznab array for connectivity prior to validating each individual indexer.

if prowlarr/jackett is not available at all -> `CrossSeedError`

### DIFF
--- a/src/arr.ts
+++ b/src/arr.ts
@@ -84,10 +84,10 @@ export async function validateUArrLs() {
 	}
 }
 
-async function checkArrIsActive(uArrL: string, arrInstance: string) {
+export async function checkArrIsActive(uArrL: string, arrInstance: string) {
 	const arrPingCheck = await makeArrApiCall<{
 		current: string;
-	}>(uArrL, "/api");
+	}>(uArrL, "/api", undefined, arrInstance);
 
 	if (arrPingCheck.isOk()) {
 		const arrPingResponse = arrPingCheck.unwrap();
@@ -123,6 +123,7 @@ async function makeArrApiCall<ResponseType>(
 	uArrL: string,
 	resourcePath: string,
 	params = new URLSearchParams(),
+	arrInstance?: string,
 ): Promise<Result<ResponseType, Error>> {
 	const apikey = getApikey(uArrL)!;
 	const url = new URL(sanitizeUrl(uArrL));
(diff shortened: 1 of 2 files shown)


ShanaryS (dismissed):

src/arr.ts
```diff
@@ -124,6 +124,7 @@ async function makeArrApiCall<ResponseType>(
 	uArrL: string,
 	resourcePath: string,
 	params = new URLSearchParams(),
+	arrInstance?: string,
 ): Promise<Result<ResponseType, Error>> {
 	const apikey = getApikey(uArrL)!;
```
ShanaryS, line 129:
```suggestion
	options: { params?: URLSearchParams, arrInstance?: string }
): Promise<Result<ResponseType, Error>> {
	const urlParams = params ?? new URLSearchParams();
	const apikey = getApikey(uArrL)!;
```

ShanaryS: I think this makes calling it easier to understand

zakkarry: check out what I did....

zakkarry: yea, saw you edited it :P

ShanaryS (approved):

src/torznab.ts
```diff
@@ -631,6 +632,14 @@ export async function validateTorznabUrls() {
...
 		}
 	}
+
+	const uniqueUrls = new Set(
+		urls.map((url) => url.origin + "?" + url.searchParams.toString()),
+	);
+	for (const host of uniqueUrls) {
+		await checkArrIsActive(host, "Torznab");
+	}
+
```
mmgoodnow, line 642: Torznab servers aren't really arr instances, so I don't think we should use code from `arr.ts` for them. What does this do that fetching the caps doesn't already do?

zakkarry: We discussed the reason I did this the other day...because it relies on timing out of all indexers during cap fetching otherwise.

This prevents the need for that and checks the unique torznab hosts before fetching caps. The arr code works for this purpose perfectly.

zakkarry: Would you prefer a set timeout for fetchCaps over this approach?

mmgoodnow: It looks like this just moves the timeout to an earlier call. We do the fetchCaps calls in parallel (below) so I think the 10 second timeout on caps calls is a sufficient solution here. 

 https://github.com/cross-seed/cross-seed/blob/136010fe01e9d086892dffc09d78ed03d9be6786/src/torznab.ts#L596

PR #757, opened 2024-07-23
revert(includeSingleEpisode): remove exemption behavior with rss when false

zakkarry:
due to the inability to configure specific exclusions from RSS and autobrr's ability to exclude things from announce in filters, it seems prudent to remove the exclusion of RSS when set to false and only have announce be exempt.

for users wishing to exclude single episodes from their announce, [this is configurable from autobrr](https://autobrr.com/filters/examples#only-season-packs) and is not reliant on this setting in cross-seed to accomplish the same goal.

this allows for announce to be used to cross-seed single episodes, when searching for single episodes is not desired the option can still be set to false to relieve stress on API calls to indexers during searches.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -26,7 +26,7 @@ const ZodErrorMessages = {
 	qBitAutoTMM:
 		"If using Automatic Torrent Management in qBittorrent, please read: https://www.cross-seed.org/docs/v6-migration#qbittorrent",
 	includeSingleEpisodes:
-		"includeSingleEpisodes is not recommended when using rss or announce, please read: https://www.cross-seed.org/docs/v6-migration#updated-includesingleepisodes-behavior",
+		"includeSingleEpisodes is not recommended when using announce, please read: https://www.cross-seed.org/docs/v6-migration#updated-includesingleepisodes-behavior",
 	needsInject: "You need to use the 'inject' action for partial matching.",
 	needsLinkDir:
 		"You need to set a linkDir (and have your data accessible) for risky or partial matching to work.",
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -12,18 +12,23 @@ import {
 	parseSource,
 	TORRENT_CACHE_FOLDER,
 	ANIME_GROUP_REGEX,
+	SEASON_REGEX,
 } from "./constants.js";
 import { db } from "./db.js";
 import { Label, logger } from "./logger.js";
 import { Metafile } from "./parseTorrent.js";
 import { Candidate } from "./pipeline.js";
-import { findBlockedStringInReleaseMaybe } from "./preFilter.js";
+import {
+	findBlockedStringInReleaseMaybe,
+	isSingleEpisode,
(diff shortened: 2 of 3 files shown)


ShanaryS (approved):

mmgoodnow:
Let's step back a bit. Here are the use cases I know about:
1. I want to never cross-seed episode files.
2. I want to cross-seed as many episode files as I can. 
3. I want to cross-seed episode files if I can race on them.

In v5, we support use case 1 and use case 2. 

In https://github.com/cross-seed/cross-seed/commit/c4723561f370e2ef653a301caeda289858979267, I think we added support for use case 3 at the expense of use case 1.

Fundamentally, if there are 3 use cases, a boolean option can't cover them all. This PR seems to add back use case 1, but I feel like there's some weird hybrid compromise thing trying to cover both use case 2 and 3 at the same time. Would it make more sense to just explicitly drop support for use case 2?

zakkarry:
> Let's step back a bit. Here are the use cases I know about:
> 
>     1. I want to never cross-seed episode files.
> 
>     2. I want to cross-seed as many episode files as I can.
> 
>     3. I want to cross-seed episode files if I can race on them.
> 
> 
> In v5, we support use case 1 and use case 2.
> 
> In [c472356](https://github.com/cross-seed/cross-seed/commit/c4723561f370e2ef653a301caeda289858979267), I think we added support for use case 3 at the expense of use case 1.
> 
> Fundamentally, if there are 3 use cases, a boolean option can't cover them all. This PR seems to add back use case 1, but I feel like there's some weird hybrid compromise thing trying to cover both use case 2 and 3 at the same time. Would it make more sense to just explicitly drop support for use case 2?

I've always raced single episodes from announce, even on v5. I'm not sure what you mean in v5 not supporting racing them - do you mean that it wasn't exclusively racing only?

I think this is the best approach, given that filtering can be done in autobrr for announce - and the rest is determinate based on the config option, as it was in v5 and pre this change.

mmgoodnow:
> I've always raced single episodes from announce, even on v5. I'm not sure what you mean in v5 not supporting racing them - do you mean that it wasn't exclusively racing only?

I mean if you were using it for racing via announce, then you had no choice but to also incur all of the bulk searches—you could say v5 shoehorned the racing use case into use case 2.

zakkarry:
> > I've always raced single episodes from announce, even on v5. I'm not sure what you mean in v5 not supporting racing them - do you mean that it wasn't exclusively racing only?
> 
> I mean if you were using it for racing via announce, then you had no choice but to also incur all of the bulk searches—you could say v5 shoehorned the racing use case into use case 2.

I suppose so, but this seems like a more flexible solution - and one I frankly think is preferred.

Do you think there is a better approach than this one given that the filters in autobrr can omit the single episodes, and the rest is rejected based on config options?

mmgoodnow:
> this seems like a more flexible solution - and one I frankly think is preferred.

OK, sounds good to me

mmgoodnow (dismissed):

PR #758, opened 2024-07-24
enhancement(rss): support pagination for rss if gap longer than rssCadence

ShanaryS:
If the `lastRun` for the rss job is more than `rssCadence + ms("1 minute")`, will page back until it's caught up or a maximum of 12 pages. Each rss request has a delay of 5 minute. Will detect if an indexer does not support pagination by comparing the candidates to the previous "page".

This should cover at least 3 hours to a few days depending on how many results the tracker returns each time and frequency of uploads. But unfortunately, it seems quite a few indexers don't support this. In that scenario it will only cost a single extra rss query regardless of how long `cross-seed` was down for.

Added a migration as we are now using indexer limits for rss scan.

### DIFF
--- a/src/arr.ts
+++ b/src/arr.ts
@@ -135,7 +135,10 @@ async function makeArrApiCall<ResponseType>(
 			headers: { "X-Api-Key": apikey },
 		});
 	} catch (networkError) {
-		if (networkError.name === "AbortError") {
+		if (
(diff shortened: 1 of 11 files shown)


mmgoodnow:
@ShanaryS I am going to take over this PR.

PR #759, opened 2024-07-24
update(schema/cli): allow override with --no of config for exclude options

zakkarry:
for many users, overriding the exclude time-based variables via cli has been valuable for searching, with zod this is not possible due to requiring ms formatting.

this PR allows the `--no` prefix to be used in commander (cli) for the cli parameters

`--no-exclude-older`
`--no-exclude-recent-search`

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -191,6 +191,14 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			"Exclude torrents which have been searched more recently than n minutes ago. Bypasses the -a flag.",
 			fileConfig.excludeRecentSearch,
 		)
+		.option(
+			"--no-exclude-older",
+			"Don't Exclude torrents based on when they were first seen.",
+		)
+		.option(
+			"--no-exclude-recent-search",
+			"Don't Exclude torrents based on when they were last searched.",
+		)
 		.option("-v, --verbose", "Log verbose output", false)
 		.addOption(
 			new Option(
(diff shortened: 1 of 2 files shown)


mmgoodnow (commented):
ideally the solution doesn't involve `"undefined"`. Can you clarify the use case? You mentioned overriding values from the config file, but this seems more like under-riding them.

zakkarry:
> ideally the solution doesn't involve `"undefined"`. Can you clarify the use case? You mentioned overriding values from the config file, but this seems more like under-riding them.

There were users providing `undefined` in v5 with exclude-older and exclude-recent-search cli arguments, because of `ms`'s inability to translate them to milliseconds, it effectively returned NaN and provided a way to set undefined without changing the config.

This provides a way for that same functionality to effectively cancel settings when performing searches without having to edit the config file.

It works in this respect.

mmgoodnow:
can we use false internally / coerce false to null and expose it via CLI as `--no-exclude-recent-search`? 
https://github.com/cross-seed/cross-seed/blob/136010fe01e9d086892dffc09d78ed03d9be6786/src/cmd.ts#L373

zakkarry:
> can we use false internally/coerce false to null and expose it via CLI as `--no-exclude-recent-search`?
> 
> https://github.com/cross-seed/cross-seed/blob/136010fe01e9d086892dffc09d78ed03d9be6786/src/cmd.ts#L373
> 
> This is a feature of commander.js where creating an option like `-no-my-option` will show up in the program as `myOption: false`

I've done this now and tested it, didn't use literal false because it uses the type boolean and does not equal `false` in Zod. FYI.

mmgoodnow (approved):

PR #762, opened 2024-07-27
fix(searchee): use parsed titles for searchees

ShanaryS:
Also will now log the original title when source from torrent as `Title S7 [Season 7 [adfedfe7...]]`.

Now sanitizing infoHash in logs.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -12,6 +12,7 @@ import {
 import { InjectionResult, SaveResult } from "./constants.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
 import { existsSync } from "fs";
+import { sanitizeInfoHash } from "./utils.js";
 
 function getData(req: IncomingMessage): Promise<string> {
 	return new Promise((resolve) => {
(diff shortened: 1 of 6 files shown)


src/server.ts
```diff
@@ -107,7 +108,12 @@ async function search(
 		return;
 	}
 
-	const criteriaStr = inspect(criteria);
+	const criteriaStr = criteria.infoHash
+		? inspect(criteria).replace(
+				/infoHash: '.*'/i,
+				`infoHash: '${sanitizeInfoHash(criteria.infoHash)}'`,
+			)
+		: inspect(criteria);
```
mmgoodnow, line 116: why do info hashes need to be sanitized? also, can this be done by just replacing infoHash?

```js
inspect(criteria).replace(
  criteria.infoHash, 
  sanitizeInfoHash(criteria.infoHash)
)
```

zakkarry: When logs are shared, full infohashes are not needed and add what is usually unnecessary length to the logging strings.

Trackers and users both have voiced a desire to not have them shared in full, as well.

The first bit of characters generally provides the distinction that would be required to find exactly which torrent, my preference was the first and last 5 with an ellipse, but @ShanaryS thought this was fine to keep as the sanitization.

PR #763, opened 2024-07-27
logging(decide): clarify logging around size mismatch

ShanaryS:
Lots of users were confused by the original message, this should make it unambiguous as to why it was rejected.

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -42,92 +42,88 @@ export interface ResultAssessment {
 	metafile?: Metafile;
 }
 
-const createReasonLogger =
-	(searcheeLog: string, candidateName: string, tracker: string) =>
-	(
-		decision: Decision,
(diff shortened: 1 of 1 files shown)


mmgoodnow:
can you show some before/after screenshots?

ShanaryS:
@mmgoodnow
Before:
`- its size does not match by FUZZY_SIZE_MISMATCH: -2.62 GB`
`- its size does not match by PARTIAL_SIZE_MISMATCH: -42.03 KB`

After:
`- the sizes are outside of the fuzzySizeThreshold range: 0.740 > 0.1`
`- too many files are missing or have different sizes: torrent progress would be 0.89%`

Need to multiply the percentage one.

mmgoodnow:
I really like the `torrent progress would be 0.89%`

PR #768, opened 2024-08-03
fix(startup/perm): check config_dir exists before permissions check

### LINKED ISSUE #766: Vanilla install of V6 on vanilla windows fails to pinpoint correct directory
### Have you checked for any existing open/closed tickets that are the same or similar to your issue?

### I have gone to the Discord server for support and/or acknowledge that this is a bug and not a support issue

- [X] I've tried to solve this already

### What's happening surrounding the bug? (context)

cross-seed is attempting to access an incorrect directory.

### Reproducing the bug

Vanilla everything.
Install per-requisites.
Install cross-seed V6
run cross-seed

### What should be happening?

Correct functionality as V5

### Your setup

```markdown
- OS: Windows 11 x64
- Prowlarr/Jackett: 
- Docker: 
- Misc:
```

### config.js (REDACT SENSITIVE INFORMATION)

Not applicable

### verbose logs

```
D:\cross-seed>npm install -g cross-seed@next

changed 105 packages in 5s

18 packages are looking for funding
  run `npm fund` for details

D:\cross-seed>cross-seed
file:///C:/Users/Audionut/AppData/Roaming/npm/node_modules/cross-seed/dist/configuration.js:29
        throw new CrossSeedError(`cross-seed does not have R/W permissions on your config directory.${dockerMessage}`);
              ^

[CrossSeedError: cross-seed does not have R/W permissions on your config directory.]

Node.js v20.16.0

D:\cross-seed>npm install -g cross-seed

added 6 packages, removed 4 packages, and changed 101 packages in 4s

18 packages are looking for funding
  run `npm fund` for details

D:\cross-seed>cross-seed gen-config
Configuration file created at C:\Users\Audionut\AppData\Local\cross-seed\config.js

D:\cross-seed>npm install -g cross-seed@next

added 4 packages, removed 6 packages, and changed 101 packages in 1s

18 packages are looking for funding
  run `npm fund` for details

D:\cross-seed>cross-seed

C:\Users\Audionut\AppData\Local\cross-seed\config.js:32
                        "http://127.0.0.1:9117/api/v2.0/indexers/xxxxxxx],
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Your config file is improperly formatted. The location of the error is above, but you may have to look backwards to see the root cause.
Make sure that
  - strings (words, URLs, etc) are wrapped in "quotation marks"
  - any arrays (lists of things, even one thing) are wrapped in [square brackets]
  - every entry has a comma after it, including inside arrays
  
D:\cross-seed>cross-seed daemon
2024-07-31 22:50:35 info: cross-seed v6.0.0-27
2024-07-31 22:50:35 info: Validating your configuration...
```

### Failure or refusal to provide the information we require to address bugs will result in potential closure of the issue.

- [X] I acknowledge that i have searched existing issues, read the documentation and faq, read the logs to see if the answer was provided in error messaging, and attempted to get support via the Discord prior to submitting this issue.

zakkarry:
in windows, permissions check using accessSync in appDir() fails when the directory does not exist, have no received reports of this on linux but potentially could apply to it as well.

for users installing v6 fresh, when there is no config path in the appDir accessSync will throw erroneously failing to launch and generate any config directory.

this PR addresses this by checking if a path exists before running accessSync() on it. enabling the config path to be made if it does not exist rather than throw and terminate

fixes #766

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -65,6 +65,9 @@ export function appDir(): string {
 	try {
 		accessSync(appDir, constants.R_OK | constants.W_OK);
 	} catch (e) {
+		if (e.code === "ENOENT") {
+			return appDir;
+		}
 		const dockerMessage =
 			process.env.DOCKER_ENV === "true"
 				? ` Use chown to set the owner to ${process.getuid!()}:${process.getgid!()}`
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -40,8 +40,8 @@ module.exports = {
 	 *
 	 * Example: sonarr: ["http://sonarr:8989/?apikey=12345"],
 	 *
-	 *			sonarr: ["http://sonarr:8989/?apikey=12345",
-	 *					"http://sonarr4k:8989/?apikey=12345"],
+	 *      sonarr: ["http://sonarr:8989/?apikey=12345",
+	 *               "http://sonarr4k:8989/?apikey=12345"],
 	 */
 	sonarr: undefined,
 
@@ -56,8 +56,8 @@ module.exports = {
 	 *
 	 * Example: radarr: ["http://radarr:7878/?apikey=12345"],
 	 *
-	 * 			radarr: ["http://radarr:7878/?apikey=12345",
-	 * 					"http://radarr4k:7878/?apikey=12345"],
+	 *       radarr: ["http://radarr:7878/?apikey=12345",
+	 *                "http://radarr4k:7878/?apikey=12345"],
 	 */
 	radarr: undefined,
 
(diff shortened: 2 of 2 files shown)


mmgoodnow (commented):
Can we check for ENOENT in the catch block instead?

zakkarry:
> Can we check for ENOENT in the catch block instead?

Yea that works, force pushed it.

mmgoodnow (approved):

src/configuration.ts
```diff
@@ -65,6 +65,9 @@ export function appDir(): string {
 	try {
 		accessSync(appDir, constants.R_OK | constants.W_OK);
```
mmgoodnow, line 66: why does this work on linux? should we move the check somewhere else? seems a little hard since appDir is needed so early in startup, maybe it's needed before we even get to something like `doStartupValidation`

zakkarry: I honestly have no idea...it would probably be traced back to the actual c code for the fs calls differing between platforms I suppose. I didn't go digging into fs - it seemed simple enough of a problem to fix.

zakkarry: > maybe it's needed before we even get to something like `doStartupValidation`

It would be needed even before validation, gen-config uses it and does zero validation other than this iirc

mmgoodnow: image

zakkarry: Then it also affects linux, I wasn't sure if it did, this was the only report of this we have had - probably because of the vast majority of our user-base on v6 being v5 users upgrading.

>have no received reports of this on linux but potentially could apply to it as well.

mmgoodnow: ok fair enough, either way we fixed it

PR #769, opened 2024-08-09
update(mediatype/video): support for identifying mediatype of video disc images

zakkarry:
this PR adds a new array for file extensions found on Blu-ray and DVD disc images (excluding single image files - iso/img/etc) and adds a case to the switch in getMediaType for unparsable titles/folders that contain these to be identified as the VIDEO mediaType

in addition, all video discs will be rechecked and paused due to inconsistencies possible. resolving these will require manual user intervention if a conflict exists - the invalid files will need to be deleted manually to break links - and then resumed. this will otherwise overwrite the files and break the searchee's torrent.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -98,7 +98,8 @@ export function getMediaType(searchee: Searchee): MediaType {
 			return unsupportedMediaType(searchee);
 	}
 }
-export function shouldRecheck(decision: Decision): boolean {
+export function shouldRecheck(searchee: Searchee, decision: Decision): boolean {
+	if (hasExt(searchee.files, VIDEO_DISC_EXTENSIONS)) return true;
 	switch (decision) {
 		case Decision.MATCH:
 		case Decision.MATCH_SIZE_ONLY:
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -10,6 +10,7 @@ import {
 	NON_UNICODE_ALPHANUM_REGEX,
 	SCENE_TITLE_REGEX,
 	SEASON_REGEX,
+	VIDEO_DISC_EXTENSIONS,
 	VIDEO_EXTENSIONS,
 	YEARS_REGEX,
 } from "./constants.js";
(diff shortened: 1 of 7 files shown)


mmgoodnow (approved):

src/utils.ts
```diff
@@ -86,6 +87,9 @@ export function getMediaType(searchee: Searchee): MediaType {
 			return MediaType.VIDEO;
 		case hasExt(searchee.files, [".rar"]):
 			if (MOVIE_REGEX.test(searchee.title)) return MediaType.MOVIE;
```
ShanaryS, line 89: This has a fall through which means that all .rar will be considered videos. It should last.

src/utils.ts
```diff
@@ -61,7 +62,7 @@ export function filesWithExt(files: File[], exts: string[]): File[] {
 	return files.filter((f) => exts.includes(path.extname(f.name)));
```
ShanaryS, line 62: This should also use lowerCase.

ShanaryS (dismissed):

ShanaryS:
Should `VIDEO_DISC_EXTENSIONS` also be added to `parseTitle` and `nonVideoSizeRatio`? Also do we still want to ignore the disc folders in dataFiles now that explicit support is added?

zakkarry:
> Should `VIDEO_DISC_EXTENSIONS` also be added to `parseTitle` and `nonVideoSizeRatio`? Also do we still want to ignore the disc folders in dataFiles now that explicit support is added?

As far as I'm aware, prefiltering will exclude all the folders that actually contain the files matching the extensions in VIDEO_DISC_EXTENSIONS anyway, It's only real utility is in checking the searchee.files for mediaType.

I'm not sure we should really add these to be considered video files, but I don't particularly see how it could hurt I suppose - I don't download full disc images like that so it doesn't really affect me and I'm not sure the interaction. Seems like it would make sense though given what nonvideo is used for in the majority of cases.

As far as parsing goes, none of those files will ever contain any useful information for parsing.

ShanaryS (dismissed):

ShanaryS (dismissed):

PR #771, opened 2024-08-20
fix(search): remove erroneous hyphens in torznab queries

zakkarry:
when certain release names, usually non-standard or unsupported media, end with a hyphen this can be appended to the search string erroneously. 

I've done the same for the beginning of the query as well for completeness.

example: `Gav Thorpe - Catechism Of Hate - [epub]`

[epub] is excluded in `cleanseSeparators` and thus a trailing - is appended to the string due to hyphens not being cleansed at all.

resulting query: `t=search&q=Gav+Thorpe+-+Catechism+Of+Hate+-`

this PR checks for preceding and trailing hyphens and removes them

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -128,9 +128,10 @@ export function getApikey(url: string) {
 }
 export function cleanseSeparators(str: string): string {
 	return str
-		.replace(/\[.*?\]|「.*?」|｢.*?｣|【.*?】/g, "")
-		.replace(/[._()[\]]/g, " ")
-		.replace(/\s+/g, " ")
+		.replace(/\[.*?\]|「.*?」|｢.*?｣|【.*?】/g, "") // bracketed text
+		.replace(/[._()[\]]/g, " ") // release delimiters (except '-')
+		.replace(/\s+/g, " ") // successive spaces
+		.replace(/^\s*-+|-+\s*$/g, "") // "trim()" hyphens
 		.trim();
 }
 


ShanaryS:
I actually just addressed this in the inject branch, though that is created by us stripping the release group. [`642d47c` (#670)](https://github.com/cross-seed/cross-seed/pull/670/commits/642d47cf52404ac1f6b3fef8f0bbfdee5223cd05#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R248)

src/utils.ts
```diff
@@ -128,9 +128,10 @@ export function getApikey(url: string) {
 }
 export function cleanseSeparators(str: string): string {
 	return str
-		.replace(/\[.*?\]|「.*?」|｢.*?｣|【.*?】/g, "")
-		.replace(/[._()[\]]/g, " ")
-		.replace(/\s+/g, " ")
+		.replace(/\[.*?\]|「.*?」|｢.*?｣|【.*?】/g, "") // bracketed text
+		.replace(/[._()[\]]/g, " ") // release delimiters (except '-')
+		.replace(/\s+/g, " ") // successive spaces
+		.replace(/^-+|-+$/g, "") // "trim()" hyphens
```
ShanaryS, line 134: Should we cover white spaces? It might appear since we are removing stuff so I think it's necessary. Otherwise we would need to call trim().

```suggestion
		.replace(/^\s*-+s\*|\s*-+\s*$/g, "") // "trim()" hyphens
```

zakkarry: I think assuming that this will always be surrounded by spaces is flawed. I think the following is potentially a better option that the end `trim()` addresses any space remnants that would remain?

`^\s*?-+|-+\s*?$`

ShanaryS: The * covers the zero case.

zakkarry: https://regex101.com/r/654OSp/1

Test case

zakkarry: Sorry, sick and just woke up. You're correct.

So `^\s*-+|-+\s*$`

https://regex101.com/r/654OSp/2

ShanaryS: Yeah that works, I forgot about trim() so both sides aren't necessary. I had a typo in my original so that could have been the issue.

zakkarry: I've updated that, if you can resolve this and approve if it looks good to you.

ShanaryS (approved):

PR #773, opened 2024-08-22
enhancement(blocklist): torrent categories, tags, labels, sizes, and tracker hosts

ShanaryS:
qBittorrent strips everything from `BT_backup/abc.torrent` except for the info dict. So we need to read the tracker urls from the `BT_backup/abc.fastresume`. This is trivial since it's in the same directory. It also has every bit of information about the torrent so we read categories and tags at the same time.

All other clients use API calls to get the labels. They have fastresume like structures but aren't in torrentDir (except for rTorrent) and so would require another config option. qBittorrent will also fallback to use API in the case of `cross-seed search --torrents` as they might not be from `BT_backup`.

Labels are considered tags, since a client may support multiple in the future (transmission does). Categories and tags must match exactly while tracker urls require the host `https://user:pass@tracker.example.com:8080/announce/key` -> `tracker.example.com:8080`.

This metadata is only added when creating searchees for `search`, `rss`, `announce`, `webhook`, and `inject` except for reading trackers from .torrent files.

Each blocklist item now include a prefix of it's type. Supported are: `name:` `folder:` `category:` `tag:` `tracker:` `hash:` `sizeBelow:` `sizeAbove:`
There is also two additional fields for regex: `nameRegex:` `folderRegex:`. This takes in a regex string and does `.test()` giving users more flexibility at their own risk. Verification is done whenever possible to fail at config time.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -2,13 +2,27 @@ import { readdirSync } from "fs";
 import ms from "ms";
 import { isAbsolute, relative, resolve } from "path";
 import { ErrorMapCtx, RefinementCtx, z, ZodIssueOptionalMessage } from "zod";
-import { Action, LinkType, MatchMode, NEWLINE_INDENT } from "./constants.js";
+import {
+	Action,
+	BlocklistType,
+	LinkType,
+	MatchMode,
+	NEWLINE_INDENT,
+	parseBlocklistEntry,
+} from "./constants.js";
 import { logger } from "./logger.js";
+import { formatAsList } from "./utils.js";
 
 /**
  * error messages and map returned upon Zod validation failure
  */
 const ZodErrorMessages = {
+	blocklistType: `Blocklist item does not start with a valid prefix. Must be of ${formatAsList(Object.values(BlocklistType), { sort: false, style: "narrow", type: "unit" })}`,
+	blocklistRegex: `Blocklist regex is not a valid regex.`,
+	blocklistFolder: `Blocklist folder must not contain path separators`,
+	blocklistTracker: `Blocklist tracker is not a valid URL host. If URL is https://user:pass@tracker.example.com:8080/announce/key, you must use "tracker:tracker.example.com:8080"`,
+	blocklistHash: `Blocklist hash must be 40 characters and alphanumeric`,
+	blocklistSize: `Blocklist size must be an integer for the number of bytes. You can only have one sizeBelow, one sizeAbove, and sizeBelow <= sizeAbove.`,
 	vercel: "format does not follow vercel's `ms` style ( https://github.com/vercel/ms#examples )",
 	emptyString:
 		"cannot have an empty string. If you want to unset it, use null or undefined.",
@@ -92,9 +106,10 @@ function addZodIssue(
 
 /**
  * helper function for ms time validation
+ * @param durationStr the duration string to validate
+ * @param ctx ZodError map
  * @return transformed duration (string -> milliseconds)
  */
-
 function transformDurationString(durationStr: string, ctx: RefinementCtx) {
 	const duration = ms(durationStr);
 	if (isNaN(duration)) {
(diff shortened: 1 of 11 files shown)


zakkarry:
Have not exhaustingly tested this, but feel like its been working well.

@ShanaryS did you end up pushing the halting of walking through a directory and its children even when the parent directory wasnt directly the folder?

As I had mentioned, you could employ similar means as the prefiltering did for directories on IGNORED

src/torrent.ts
```diff
@@ -48,8 +50,48 @@ export enum SnatchError {
...
+	return Metafile.decode(await readFile(filename));
+}
+
+export async function parseTorrentWithMetadata(
+	filename: string,
+	torrentInfos: TorrentMetadataInClient[],
+): Promise<Metafile> {
+	const meta = await parseTorrentFromFilename(filename);
+	const client = getClient();
+	if (client?.type === Label.QBITTORRENT) {
```
mmgoodnow, line 62: why is qBittorrent special cased? the idea of being able to rely more on public APIs and less on private APIs sounds great in theory. Is there a hard technical limitation?

src/preFilter.ts
```diff
@@ -139,12 +141,40 @@ export function findBlockedStringInReleaseMaybe(
...
+				);
+			case BlocklistType.HASH:
+				return blocklistValue === searchee.infoHash;
+			case BlocklistType.SIZE_BELOW:
+				return searchee.length < parseInt(blocklistValue);
+			case BlocklistType.SIZE_ABOVE:
+				return searchee.length > parseInt(blocklistValue);
+			default:
+				throw new Error(`Unknown blocklist type: ${blockedStr}`);
+		}
```
mmgoodnow, line 177: I think we should try to make the new colon-separated blocklist entry format in a purely additive way—if there's no colon or the stuff to the left of a colon isn't one of the supported things, we can preserve logic to fall back to the legacy path/name/infohash matching and warn the user that it's deprecated.

src/parseTorrent.ts
```diff
@@ -112,6 +186,21 @@ export class Metafile {
...
 		}
 		this.title = parseTitle(this.name, this.files) ?? this.name;
+		this.tags = [];
+		this.trackers =
+			raw["announce-list"]?.map((tier) =>
+				tier
+					.map((url) => sanitizeTrackerUrl(url.toString()))
+					.filter(isOk)
+					.map((r) => r.unwrap()),
+			) ??
```
mmgoodnow, line 196: there's probably a pretty easy way to share this code across `announce` and `announce-list` (and I just saw a third one in `updateMetafileMetadata`).

src/parseTorrent.ts
```diff
@@ -23,9 +25,78 @@ interface Torrent {
...
+ * qBittorrent: .fastresume, Transmission: .resume, rTorrent: .rtorrent
+ * Deluge doesn't store labels in fastresume, but in label.conf
+ */
+interface TorrentMetadata {
+	trackers?: Buffer[][];
+	"qBt-category"?: Buffer;
+	"qBt-tags"?: Buffer;
+	custom1?: Buffer;
+	label?: Buffer;
+	labels?: Buffer;
```
mmgoodnow, line 45: is Buffer the right type for these properties?

mmgoodnow (commented):
Looks pretty good, I'm glad we're mostly using public APIs as well. It seems like there's a logical continuation of this work where we remove or deprecate torrentDir in favor of assembling searchees entirely from torrent client APIs, which, other than solving the SQLite problem, also makes it very easy to do things like search in a frequent loop for recently completed downloads without users having to set up webhook calls.

src/configSchema.ts
```diff
@@ -107,9 +120,92 @@ function transformDurationString(durationStr: string, ctx: RefinementCtx) {
...
+ * helper function for verifying blocklist types
+ * @param blockList the blocklist to validate
+ * @param ctx ZodError map
+ * @return blocklist (string array)
+ */
+function transformBlocklist(blockList: string[], ctx: RefinementCtx) {
+	if (!Array.isArray(blockList)) {
+		return [];
+	}
+	const sizeTest = (value: string, existing: number) => {
```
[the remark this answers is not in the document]
ShanaryS: I think keeping bytes here is also preferable to keep it simple and aligned. I honestly don't see this getting much use over the other blocklist options.

src/parseTorrent.ts
ShanaryS: Yes, these should be all encoded similar to .torrent files. Though only qbit is using it currently.

src/torrent.ts
ShanaryS: qBit strips the trackers from the .torrent files in torrentDir. Getting it from client requires an individual request for each torrent, the normal get all torrents request only returns the active one (none if paused). Luckily, the trackers are in the .fastresume files that are also in torrentDir. It also contains the category and tag so we just read for there.

qBit does have a fallback if users are passing the files through the cross-seed arg.

hoonlight:
It works really well. Thanks for your contribution.

src/parseTorrent.ts
mmgoodnow: if only qBit is using it, the `custom1`/`label`/`labels` properties don't need to be there.

src/configSchema.ts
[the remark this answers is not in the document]
mmgoodnow: sounds good.

src/torrent.ts
mmgoodnow: I appreciate that we have a fallback that doesn't rely on private APIs. We can accept this tradeoff for now, but I think we should think about our path to supporting SQLite resume data. I'm really hoping that a lot of the performance issues with making API calls to qBittorrent go away but I haven't used it. 

In theory, making N+1 calls should be viable enough if we combine it with lazy loading of the relevant data and small searchLimits (100-200 api calls to localhost _should_ be basically nothing).

mmgoodnow:
Let's keep an eye on the performance—only one API call to the client shouldn't be too bad (knock on wood), but eagerly reading an extra file for every torrent file in `torrentDir` isn't exactly cheap, especially for users that aren't even using tracker blocking.

mmgoodnow (approved):

PR #776, opened 2024-08-24
fix(prefilter): add check for non-video ratio in arr dir check

zakkarry:
this PR validates that the folders we exclude for "arr movie/series folder" are video. 

this prevents poorly named non-video torrents/data from being included in this exclusion erroneously

- [x] consider using a constant percentage rather than fuzzysize?
   - using 0.02

### DIFF
--- a/src/preFilter.ts
+++ b/src/preFilter.ts
@@ -119,7 +119,8 @@ export function filterByContent(
 		ARR_DIR_REGEX.test(basename(searchee.path)) &&
 		!(
 			searchee.files.length > 1 &&
-			SONARR_SUBFOLDERS_REGEX.test(basename(searchee.path))
+			SONARR_SUBFOLDERS_REGEX.test(basename(searchee.path)) &&
+			nonVideoSizeRatio < 0.02
 		)
 	) {
 		logReason(
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -6,3 +6,5 @@
 /dataSources.local.xml
 # Editor-based HTTP Client requests
 /httpRequests/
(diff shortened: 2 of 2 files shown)


src/preFilter.ts
```diff
@@ -119,7 +119,8 @@ export function filterByContent(
 		ARR_DIR_REGEX.test(basename(searchee.path)) &&
 		!(
 			searchee.files.length > 1 &&
-			SONARR_SUBFOLDERS_REGEX.test(basename(searchee.path))
+			SONARR_SUBFOLDERS_REGEX.test(basename(searchee.path)) &&
+			nonVideoSizeRatio < fuzzySizeThreshold
```
ShanaryS, line 123: What non video can we reasonably expect to see in arr dirs? Just subtitles? If so we should hardcode to .02 though it doesn't matter.

If we can expect bigger files then we should hardcode it higher. The goal should be to prune what we can expect and leaving rare occurrences to be searched.

zakkarry: It would mostly be metadata and subs. Poorly named books, music, and other things are currently being detected.

I think .02 or .05 really doesn't make a big difference in accuracy, but I'd rather be lenient on this identification since its extremely lenient in what qualifies.

What do you think?

ShanaryS: Lenient in which direction? I think if 99% of media folders have less than 0.02, we should use that.

zakkarry: Lenient as in we shouldn't be excluding them aggressively. .02 seems fine. I'll update it.

ShanaryS (approved):

PR #778, opened 2024-08-26
fix(torznab/ebook): filter calibre library index from tail of data-based search titles

zakkarry:
[Calibre](https://calibre-ebook.com/) libraries (a popular ebooks manager) can append an index ID to the end of folder names, this PR removes that index if it is a databased match.

To accomplish this the query creation function now accepts a searchee rather than a stripped searchee title. This arg was changed elsewhere to accommodate this change for the ebook searchee check.

There is a debug function (unused) that currently will not work without some modifications, I've `ts-expect-error`'d it to prevent it without removing anything.

https://regex101.com/r/xUPDfI/1

### DIFF
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -9,6 +9,7 @@ import {
 	formatFoundIds,
 } from "./arr.js";
 import {
+	CALIBRE_INDEXNUM_REGEX,
 	EP_REGEX,
 	SEASON_REGEX,
 	UNKNOWN_TRACKER,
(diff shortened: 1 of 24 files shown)


src/torznab.ts
```diff
@@ -299,6 +300,15 @@ async function createTorznabSearchQueries(
 			t: "search",
 			q: animeQuery,
 		}));
+	} else if (mediaType === MediaType.BOOK) {
+		return [
+			{
+				t: "search",
+				q: searchee.path
+					? cleanTitle(stem).replace(/\s?\(\d+\)$/, "")
```
ShanaryS, line 308: I think this should be in constants so the name describes why it's needed.

ShanaryS (approved):

PR #779, opened 2024-08-28
fix(action/inject): prevent `cross-seed inject` if no client

ShanaryS:
Also updated `getClient()` to return null if no client is configured.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -203,12 +203,9 @@ export async function linkAllFilesInMetafile(
 	>
 > {
 	const { linkDir, flatLinking } = getRuntimeConfig();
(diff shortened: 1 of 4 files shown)


mmgoodnow (commented):
I don't really see how changing it to `TorrentClient | null` helps anything if we have to force unwrap it with `!` every time we use it.

ShanaryS:
It's necessary for the blocklist branch as the client get's used outside of injection, but I added it here so the diff is easier.

mmgoodnow:
If you want a simpler diff, would you mind making the case for that in a separate PR?

ShanaryS:
It's very straight forward so I don't think it's worth it. I might move all the new functions added to clients to it's out PR though. That would make the rest easier to review.

PR #781, opened 2024-08-29
deserialize indexer objects

mmgoodnow:
This PR consolidates all of the `JSON.parse` code we were doing for some indexer columns (cat_caps, movie_id_caps, tv_id_caps) into one place. From now on, all database calls for the `indexers` table should do so by calling functions in `indexers.ts`. We can add functions to `indexers.ts` as necessary to accomplish this.

- **define caps more rigorously**
- **use hydrated indexer objects everywhere**
- **consolidate field selection**
- **move last db call into indexers.ts**

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -120,7 +120,7 @@ export const VALIDATION_SCHEMA = z
 			.nonnegative({
 				message: ZodErrorMessages.delayNegative,
 			})
-			.gte(30, ZodErrorMessages.delayUnsupported)
+			.gte(process.env.DEV ? 0 : 30, ZodErrorMessages.delayUnsupported)
 			.lte(3600, ZodErrorMessages.delayUnsupported),
 		torznab: z.array(z.string().url()),
 		dataDirs: z.array(z.string()).nullish(),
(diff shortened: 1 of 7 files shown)


src/configSchema.ts
```diff
@@ -120,7 +120,7 @@ export const VALIDATION_SCHEMA = z
 			.nonnegative({
 				message: ZodErrorMessages.delayNegative,
 			})
-			.gte(30, ZodErrorMessages.delayUnsupported)
+			.gte(process.env.DEV ? 0 : 30, ZodErrorMessages.delayUnsupported)
```
zakkarry, line 123: lol

zakkarry (commented):
does deserializing them like this pose any risk of when fetching the caps failing (for instance) it throwing anywhere?

this was a problem we had initially that for some reason fetchCaps would fail and throw because the type was all messed up.

just want to be sure that this scenario (or any other with erroneous/failed/absent values in the type are not going to cause problems and handled appropriately

mmgoodnow:
I will test what happens when fetching the caps fails.

mmgoodnow:
It looks like it just skips over them. Seeing as there wasn't any error handling in the places I removed them from, I don't think it's a worry.

PR #783, opened 2024-08-29
enhancement(general): extract non feature functionality for upcoming features

ShanaryS:
This should make the feature PRs easier to review. Most of what was extracted won't have any immediate effect but will be necessary for the upcoming features.

Client type and `getClient()` being optional allows blocklisting using torrent data if available. Type allows handling qbit specifically since we read the data from fastresume usually.

`getAllTorrents()` is used to get the blocklisting information from the clients. Also updated `checkOriginalTorrent()` for rTorrent for this purpose.

`getAllDownloadDirs()` is to find the absolute file path of each episode torrent for ensemble. This prevents individual requests for each torrent.

`getLargestFile()` and `getNewestFileAge()` are for getting the episode file from a searchee and controlling when to search for ensemble.

Added binary for `humanReadableSize()` as the limit for resume is in binary since it more closely represents download data.

Others are code quality.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -203,9 +203,12 @@ export async function linkAllFilesInMetafile(
 	>
 > {
 	const { linkDir, flatLinking } = getRuntimeConfig();
-	const newMetaClientSavePathRes = await getClient().getDownloadDir(newMeta, {
-		onlyCompleted: false,
-	});
+	const newMetaClientSavePathRes = await getClient()!.getDownloadDir(
+		newMeta,
+		{
+			onlyCompleted: false,
+		},
+	);
 	const fullLinkDir = newMetaClientSavePathRes.isOk()
 		? newMetaClientSavePathRes.unwrap() // Use existing if ALREADY_EXISTS
 		: flatLinking
@@ -234,7 +237,7 @@ export async function linkAllFilesInMetafile(
 		}
 		sourceRoot = searchee.path;
 	} else {
-		const downloadDirResult = await getClient().getDownloadDir(
+		const downloadDirResult = await getClient()!.getDownloadDir(
 			searchee as SearcheeWithInfoHash,
 			{ onlyCompleted: options.onlyCompleted },
 		);
@@ -332,7 +335,7 @@ export async function performAction(
 		// should be a MATCH, as risky requires a linkDir to be set
 		destinationDir = dirname(searchee.path);
 	}
-	const result = await getClient().inject(
+	const result = await getClient()!.inject(
 		newMeta,
 		searchee,
 		decision,
--- a/src/action.ts
+++ b/src/action.ts
@@ -10,7 +10,7 @@ import {
 	symlinkSync,
 } from "fs";
 import { dirname, join, resolve } from "path";
-import { getClient } from "./clients/TorrentClient.js";
+import { getClient, shouldRecheck } from "./clients/TorrentClient.js";
 import {
 	Action,
 	ActionResult,
@@ -339,8 +339,8 @@ export async function performAction(
 
 	logActionResult(result, newMeta, searchee, tracker, decision);
 	if (result === InjectionResult.SUCCESS) {
-		// For an easy re-injection user workflow when multiple MATCH_PARTIAL
-		if (decision === Decision.MATCH_PARTIAL) {
+		// cross-seed may need to process these with the inject job
+		if (shouldRecheck(searchee, decision)) {
 			await saveTorrentFile(tracker, getMediaType(searchee), newMeta);
 		}
 	} else if (result !== InjectionResult.ALREADY_EXISTS) {
--- a/src/action.ts
+++ b/src/action.ts
@@ -203,17 +203,12 @@ export async function linkAllFilesInMetafile(
 	>
 > {
 	const { linkDir, flatLinking } = getRuntimeConfig();
-	const newMetaClientSavePathRes = await getClient()!.getDownloadDir(
-		newMeta,
-		{
-			onlyCompleted: false,
-		},
+	const clientSavePathRes = await getClient()!.getDownloadDir(newMeta, {
+		onlyCompleted: false,
+	});
+	const fullLinkDir = clientSavePathRes.orElse(
+		flatLinking ? linkDir : join(linkDir, tracker),
 	);
-	const fullLinkDir = newMetaClientSavePathRes.isOk()
-		? newMetaClientSavePathRes.unwrap() // Use existing if ALREADY_EXISTS
-		: flatLinking
-			? linkDir
-			: join(linkDir, tracker);
 
 	let sourceRoot: string;
 	if (searchee.path) {
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -144,34 +144,65 @@ export default class RTorrent implements TorrentClient {
 	}
 
 	private async checkOriginalTorrent(
-		data: SearcheeWithInfoHash | Metafile,
+		infoHash: string,
 		onlyCompleted: boolean,
 	): Promise<
 		Result<
-			{ directoryBase: string; isMultiFile: boolean },
+			{
+				name: string;
+				directoryBase: string;
+				bytesLeft: number;
+				hashing: number;
+				isMultiFile: boolean;
+				isActive: boolean;
+			},
 			"FAILURE" | "TORRENT_NOT_COMPLETE" | "NOT_FOUND"
 		>
 	> {
-		const infoHash = data.infoHash.toUpperCase();
+		const hash = infoHash.toUpperCase();
 		type ReturnType =
-			| [[string], ["0" | "1"], ["0" | "1"]]
-			| [Fault, Fault, Fault];
+			| [
+					[string],
+					[string],
+					[number],
+					[number],
+					["0" | "1"],
+					["0" | "1"],
+					["0" | "1"],
+			  ]
+			| Fault[];
 
 		let response: ReturnType;
 		try {
 			response = await this.methodCallP<ReturnType>("system.multicall", [
 				[
 					{
+						methodName: "d.name",
+						params: [hash],
+					},
+					{
 						methodName: "d.directory",
-						params: [infoHash],
+						params: [hash],
+					},
+					{
+						methodName: "d.left_bytes",
+						params: [hash],
+					},
+					{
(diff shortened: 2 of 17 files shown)


src/action.ts
```diff
@@ -234,7 +232,7 @@ export async function linkAllFilesInMetafile(
 		}
 		sourceRoot = searchee.path;
 	} else {
-		const downloadDirResult = await getClient().getDownloadDir(
+		const downloadDirResult = await getClient()!.getDownloadDir(
```
zakkarry, line 235: didn't @mmgoodnow say no? that this wasn't necessary?

https://github.com/cross-seed/cross-seed/pull/779#pullrequestreview-2267434290

https://github.com/cross-seed/cross-seed/pull/779/commits/77617fa1a6d3843d7fdfe035f12e5a3066fb567b

ShanaryS: It is necessary for the blocklist branch

zakkarry: Given that he discarded them and made that comment, I think a little more context and reasoning about why doing it this way would be welcome, perhaps.

zakkarry: Having them discarded, and then throwing them back in another PR without a note in the PR description seems a little....I don't know, devious? belligerent? weird?

ShanaryS: The reason it has never been needed is because getClient wasn't being used outside of an injection context, which requires the client to exist. With blocklisting, we need to use the client but the user could be on `action: save`.

ShanaryS: > Having them discarded, and then throwing them back in another PR without a note in the PR description seems a little....I don't know, devious? belligerent? weird?

It wouldn't be here if it wasn't necessary.

mmgoodnow:
Thanks for the separate commits, it makes it really easy to review. Can you enumerate all of the distinct changes here and what they're needed for?

ShanaryS:
Updated the description for the commits.

src/utils.ts
```diff
@@ -355,6 +350,17 @@ export function extractInt(str: string): number {
 	return parseInt(str.match(/\d+/)![0]);
 }
 
+export function getLargestFile(files: File[]): File {
+	return files.reduce((a, b) => (a.length > b.length ? a : b));
+}
+
+export function getNewestFileAge(files: File[]): number {
+	return files.reduce(
+		(acc, file) => Math.max(acc, statSync(file.path).mtimeMs),
```
mmgoodnow, line 359: In a tight loop like this, I prefer to do `fs` calls async:

```ts
import { stat } from 'fs/promises';

async function getNewestFileAge(files: File[]): Promise<number> {
	return Math.max(...await Promise.all(files.map(f => stat(f.path)));
}
```

also, these functions are a little bit more business logicky than most utils. It might make more sense for them to exist in `searchee.ts`, what do you think?

src/clients/Transmission.ts
```diff
@@ -177,6 +190,20 @@ export default class Transmission implements TorrentClient {
...
+	}): Promise<Map<string, string>> {
+		const res = await this.request<TorrentGetResponseArgs>("torrent-get", {
+			fields: ["hashString", "downloadDir", "percentDone"],
+		});
+		const downloadDirs = new Map<string, string>();
+		for (const { hashString, downloadDir, percentDone } of res.torrents) {
+			if (options.onlyCompleted && percentDone < 1) continue;
+			downloadDirs.set(hashString, downloadDir);
+		}
+		return downloadDirs;
```
mmgoodnow, line 204: very nit, but in idiomatic JS we would ordinarily do this with `.filter()` rather than `if () continue`. The for loop itself is fine

src/clients/RTorrent.ts
```diff
@@ -297,6 +397,60 @@ export default class RTorrent implements TorrentClient {
...
+					};
+				}),
+			]);
+		} catch (e) {
+			logger.error({
+				Label: Label.RTORRENT,
+				message: "Failed to get torrent info for all torrents",
+			});
+			logger.debug(e);
+			return [];
```
mmgoodnow, line 419: is this going to be a leaky blocklist?

src/clients/TorrentClient.ts
```diff
@@ -1,25 +1,51 @@
...
+
+export interface GenericTorrentInfo {
+	infoHash: string;
+	category: string;
+	tags: string[];
+	trackers?: string[][];
+}
 
 export interface TorrentClient {
+	type: TorrentClientType;
```
mmgoodnow, line 34: what is this for? can we keep it separate from logging labels?

src/clients/RTorrent.ts
```diff
@@ -297,6 +397,60 @@ export default class RTorrent implements TorrentClient {
...
+		try {
+			if (isFault(response)) {
+				logger.error({
+					Label: Label.RTORRENT,
+					message:
+						"Fault while getting torrent info for all torrents",
+				});
+				logger.debug(inspect(response));
+				return [];
+			}
```
mmgoodnow, line 435: looks like this can be moved above `try`

src/clients/RTorrent.ts
```diff
@@ -297,6 +397,60 @@ export default class RTorrent implements TorrentClient {
...
+
+			return hashes.map((hash, index) => ({
+				infoHash: hash.toLowerCase(),
+				category: "",
+				tags: response[index][0].length
```
mmgoodnow, line 440: can you add a comment showing what this looks like?

src/clients/RTorrent.ts
```diff
@@ -281,6 +326,61 @@ export default class RTorrent implements TorrentClient {
 			.mapErr((error) => (error === "FAILURE" ? "UNKNOWN_ERROR" : error));
 	}
 
+	async getAllDownloadDirs(options: {
+		metas: SearcheeWithInfoHash[] | Metafile[];
```
mmgoodnow, line 330: looks like this could just take `infoHashes` instead of `metas`?

src/clients/RTorrent.ts
```diff
@@ -281,6 +326,61 @@ export default class RTorrent implements TorrentClient {
...
+		onlyCompleted: boolean;
+	}): Promise<Map<string, string>> {
+		const hashes: string[] = options.metas.map((meta) => meta.infoHash);
```
mmgoodnow, line 333: nit: try to keep to `infoHash` naming convention, so `infoHashes`

src/clients/RTorrent.ts
```diff
@@ -297,6 +397,60 @@ export default class RTorrent implements TorrentClient {
 		}
 	}
 
+	async getAllTorrents(): Promise<GenericTorrentInfo[]> {
+		const hashes = await this.methodCallP<string[]>("download_list", []);
```
mmgoodnow, line 401:
```suggestion
		const infoHashes = await this.methodCallP<string[]>("download_list", []);
```

src/clients/TorrentClient.ts
```diff
@@ -1,25 +1,51 @@
...
-let activeClient: TorrentClient;
+let activeClient: TorrentClient | null = null;
+
+export type TorrentClientType =
+	| Label.QBITTORRENT
+	| Label.RTORRENT
+	| Label.TRANSMISSION
+	| Label.DELUGE;
+
+export interface GenericTorrentInfo {
```
mmgoodnow, line 26: I'm trying to think of something slightly more descriptive but nothing is great…`TorrentMetadata`? `TorrentMetadataInClient` `ClientTorrentMetadata` `TorrentClientMetadata` `TorrentAttributes` `TorrentParameters` `ClientTorrentParameters` etc. 

idk 🤷

src/clients/TorrentClient.ts
```diff
@@ -1,25 +1,51 @@
...
 	getDownloadDir: (
 		meta: SearcheeWithInfoHash | Metafile,
 		options: { onlyCompleted: boolean },
 	) => Promise<
 		Result<string, "NOT_FOUND" | "TORRENT_NOT_COMPLETE" | "UNKNOWN_ERROR">
 	>;
+	getAllDownloadDirs: (options: {
+		metas: SearcheeWithInfoHash[] | Metafile[];
+		onlyCompleted: boolean;
+	}) => Promise<Map<string, string>>;
```
mmgoodnow, line 48: I'm assuming this maps info hash to download dir?

mmgoodnow (commented):
Overall the changes are pretty innocuous. I think my primary concern is returning `[]` in error paths accidentally unblocklisting things.

src/clients/TorrentClient.ts
mmgoodnow: > Type allows handling qbit specifically since we read the data from fastresume usually.

I see. I don't really want to introduce a dependency on fastresume files which are a private API for qbittorrent (and being deprecated!).

 If we already support this via client calls what is the point of reading the fastresume files?

src/clients/RTorrent.ts
ShanaryS: I'm not sure what we can do about it. But if anything, I think that should be handled by the caller.

src/clients/TorrentClient.ts
ShanaryS: I used labels as it's one less thing to import and use. And it will always exist for any future clients.

We have to read the .fastresume files to get the tracker information, so we might as well read category and tags at the same time. The current api call only returns the active tracker (nothing if paused) when getting all torrents. To get all trackers regardless of state would require a separate api call for each torrent. Moving to the pure api method for qbit will need to cache that value between `cross-seed` restarts since they won't change (user can add tracker but we don't support modifying torrents anyways).

src/clients/RTorrent.ts
ShanaryS: This is an interface function and qbit needs `Searchee | Metafile` in order to handle subfolder layout.

src/clients/TorrentClient.ts
ShanaryS: Yes. It works similar to `getAllTorrents()` but that has the infoHash in the object.

src/clients/RTorrent.ts
mmgoodnow: I still think this should maybe throw or return an ErrResult, but probably not a huge deal. We can move forward.

mmgoodnow (approved):

mmgoodnow:
Approved but let's hold off until v6.0.0 has been released.

PR #785, opened 2024-09-01
enhancement(clear-cache): also clear search timestamps

ShanaryS:
`cross-seed clear-cache` will now also clear the search timestamps which will re-search the users library. To prevent all searches happening at once, they are each delayed randomly for up to `30 days`. `first_searched` is set to when the first new search will occur and `last_searched` starts at `nMsAgo(excludeRecentSearch ?? 0)`.

Users may need to use this command upon migrating to v6 or if they were on v6 pre-release due to incremental improvements. They should only need to do this once, unless we make significant improvements to `cross-seed` (unlikely).

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -301,7 +301,7 @@ program
 		"Clear the cache without causing torrents to be re-snatched and reset the timestamps for excludeOlder and excludeRecentSearch",
 	)
 	.action(async () => {
-		console.log("Clearing cache, this may take a few minutes...");
+		console.log("Clearing cache...");
 		await db("decision").whereNull("info_hash").del();
 		await db("timestamp").del();
 		await db.destroy();
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 import chalk from "chalk";
 import { Option, program } from "commander";
-import ms from "ms";
 import { inspect } from "util";
 import { getApiKey, resetApiKey } from "./auth.js";
 import { customizeErrorMessage, VALIDATION_SCHEMA } from "./configSchema.js";
@@ -26,11 +25,7 @@ import {
 	initializePushNotifier,
 	sendTestNotification,
 } from "./pushNotifier.js";
-import {
-	getRuntimeConfig,
-	RuntimeConfig,
-	setRuntimeConfig,
-} from "./runtimeConfig.js";
+import { RuntimeConfig, setRuntimeConfig } from "./runtimeConfig.js";
 import { createSearcheeFromMetafile } from "./searchee.js";
 import { serve } from "./server.js";
 import "./signalHandlers.js";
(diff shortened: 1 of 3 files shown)


src/cmd.ts
```diff
@@ -295,13 +300,31 @@ program
...
+	const rows = await db("timestamp").select("searchee_id").distinct();
+	for (const row of rows) {
+		const offset = Math.floor(Math.random() * limit);
+		await db("timestamp")
+			.where({ searchee_id: row.searchee_id })
+			.update({
+				first_searched: now + offset,
+				last_searched: start + offset,
+			});
+	}
```
mmgoodnow, line 324: I'm not really a fan of misrepresenting first and last searched in the database to achieve this goal of spreading when we have `searchLimit` that does the same thing without lying about the numbers. Is there any reason we can't rely on searchLimit here?

ShanaryS: We could clear the table and have the users rely on that. The timestamps would come out to same in the end but I imagine a large number of users won't do that. So I think it's better to force it since these have already been searched anyways.

mmgoodnow: Can we force searchLimit then?

ShanaryS: I'm not sure of a good way to do that. Maybe checking the searchee and timestamps table and enforcing it if `x` amount of first time searches? The thing is, this only a problem for launch. So what we can do instead is just hard code a search limit for the first few weeks. Regardless, this would affect new searchees as well and if a tracker was down, it will research the same searchees the next day on that tracker only.

The more I think about it, the more I find the timestamps solution to be better as it's a direct solution to the problem. There is nothing required from the user, no code change, and guarantees the desired behavior. We can have a max search limit of 500 as a fallback incase users start with a long exclude and lower it after clearing.

In any case, we will change it to clear the entire table after a few weeks, the concern is just tens of thousands of users doing this simultaneously at launch.

zakkarry: Has there been a concensus on this @mmgoodnow

mmgoodnow: I think we can make clear-cache wipe out the timestamps completely, and just not emphasize in the migration guide that people should do it. If they are satisfied with their cross-seed catalog they don't need to do anything special. If they want to start fresh and are have the initiative to search around for how to do it, they will find the clear-cache command and run it. I think the reduced volume of users performing this action, combined with the other protections we have in place around `excludeOlder`/`excludeRecentSearch` and `delay` are probably sufficient to mitigate any load issues.

src/cmd.ts
```diff
@@ -297,9 +297,13 @@ program
...
 program
 	.command("clear-cache")
-	.description("Clear the cache of downloaded-and-rejected torrents")
+	.description(
+		"Clear the cache without causing torrents to be re-snatched and reset the timestamps for excludeOlder and excludeRecentSearch",
+	)
 	.action(async () => {
+		console.log("Clearing cache, this may take a few minutes...");
 		await db("decision").whereNull("info_hash").del();
+		await db("timestamp").del();
```
mmgoodnow, line 306: I would not expect this to take a few minutes.

mmgoodnow (approved):

src/cmd.ts
ShanaryS: It previously did when it was updating everything individually. But yeah it takes a few seconds now.

mmgoodnow (approved):

PR #786, opened 2024-09-06
fix(utils/parsing): strip scene title before release group

ShanaryS:
`abc-...` the `...` would match as release group so we should remove the `abc-` scene prefix first.

Should we also add restrictions to release group?

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -245,11 +245,12 @@ export function getAnimeQueries(name: string): string[] {
 export function stripMetaFromName(name: string): string {
 	return sourceRegexRemove(
 		stripExtension(name)
-			.replace(RELEASE_GROUP_REGEX, "")
+			.match(SCENE_TITLE_REGEX)!
+			.groups!.title.replace(RELEASE_GROUP_REGEX, "")
 			.replace(/\s*-\s*$/, "")
 			.replace(RESOLUTION_REGEX, "")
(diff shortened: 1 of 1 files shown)


zakkarry:
No, in the scene this is sometimes abbreviated so it would be inconsistent across P2P renamed files/folders.

ShanaryS:
No to what exactly? The PR or release group restrictions?

zakkarry:
No to using the prefix as a release group, while it can sometimes be the group (all lowercase) it often times can also be abbreviations - usually for longer groupnames - so it isn't guaranteed to be a 1:1 comparison.

ShanaryS:
I think you're confused, I didn't say anything about that.

zakkarry:
>Should we also add restrictions to release group?

I was commenting on this, I assumed you meant comparing the release group the file's prefix (release group)?

ShanaryS:
No just adding max length or other restrictions.

zakkarry:
> No just adding max length or other restrictions.

In the other scene group regex I think I use a minimum, although arbitrary, I know some groups use their full group name. There isn't a measure we can add that won't have exceptions.

The only things that are usually consistent are group prefixes, hyphens, and dashes - no spaces. Most of the time video content may also contain the resolution and codec, but this is probably not enforceable.

PR #787, opened 2024-09-06
fix(inject): false positive blocklist message summary

ShanaryS:
Since all searchees are checked, a user will pretty much always have the blocklist message if no match and anything blocklisted.

### DIFF
--- a/src/inject.ts
+++ b/src/inject.ts
@@ -106,29 +106,28 @@ async function whichSearcheesMatchTorrent(
 	meta: Metafile,
 	searchees: SearcheeWithLabel[],
 ): Promise<{ matches: AllMatches; foundBlocked: boolean }> {
+	const isSimilar = (searchee: SearcheeWithLabel, meta: Metafile) =>
+		areMediaTitlesSimilar(searchee.title, meta.title) ||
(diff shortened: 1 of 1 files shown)


src/inject.ts
```diff
@@ -106,29 +106,28 @@ async function whichSearcheesMatchTorrent(
...
+				areMediaTitlesSimilar(searcheeFile.name, metaFile.name),
+			),
+		);
 	let foundBlocked = false;
 	const matches: AllMatches = [];
 	for (const searchee of searchees) {
 		const { decision } = await assessCandidate(meta, searchee, []);
 		if (decision === Decision.BLOCKED_RELEASE) {
-			foundBlocked = true;
+			if (isSimilar(searchee, meta)) foundBlocked = true;
```
mmgoodnow, line 124: I still think this is similar logic should be inside assessCandidate

mmgoodnow (approved):

src/inject.ts
ShanaryS: It would just be an unnecessary check for everything but inject. It would never catch anything that passes reverse lookup or the torznab query. It also would hurt risky matching as id search can correctly match torrents with completely different naming like the title in a different language.

Issue #362, opened 2023-01-21
Custom "cross-seed" Tag


CarlitoGil:
I'd like to use a custom tag instead of "cross-seed".
For example, I would use "(cross-seed)" with parentheses so it sorts before other tags.

PR #790, opened 2024-09-14
feat(clients/qbit): Add support for custom tagging of cross seed torrents

hexile00:
Closes https://github.com/cross-seed/cross-seed/issues/362

Adds a new `--cross-seed-tag` config for providing custom tags to cross seed torrents.

Still need to validate the changes, but don't have a setup going yet, these clients still need to be validated if someone already has a good test env:
- [ ] qbitorrent
- [ ] deluge
- [ ] transmission

This is my first PR to this project so please let me now if anything needs to be modified

### DIFF
--- a/src/clients/QBittorrent.ts
+++ b/src/clients/QBittorrent.ts
@@ -2,7 +2,6 @@ import { dirname, resolve } from "path";
 import {
 	DecisionAnyMatch,
 	InjectionResult,
-	TORRENT_TAG,
 	TORRENT_CATEGORY_SUFFIX,
 } from "../constants.js";
 import { CrossSeedError } from "../errors.js";
@@ -208,26 +207,27 @@ export default class QBittorrent implements TorrentClient {
 		searcheeInfo: TorrentInfo | undefined,
 		path: string | undefined,
 	): string {
-		const { duplicateCategories, linkCategory } = getRuntimeConfig();
+		const { duplicateCategories, linkCategory, crossSeedTag } = getRuntimeConfig();
 
 		if (!duplicateCategories || !searcheeInfo || !path) {
-			return TORRENT_TAG; // Require path to duplicate category using tags
+			return crossSeedTag; // Require path to duplicate category using tags
 		}
 		const searcheeCategory = searcheeInfo.category;
 		if (!searcheeCategory.length || searcheeCategory === linkCategory) {
-			return TORRENT_TAG;
+			return crossSeedTag;
 		}
 
 		if (searcheeCategory.endsWith(TORRENT_CATEGORY_SUFFIX)) {
-			return `${TORRENT_TAG},${searcheeCategory}`;
+			return `${crossSeedTag},${searcheeCategory}`;
 		}
-		return `${TORRENT_TAG},${searcheeCategory}${TORRENT_CATEGORY_SUFFIX}`;
+		return `${crossSeedTag},${searcheeCategory}${TORRENT_CATEGORY_SUFFIX}`;
 	}
 
 	async createTag(): Promise<void> {
+		const { crossSeedTag } = getRuntimeConfig();
 		await this.request(
 			"/torrents/createTags",
-			`tags=${TORRENT_TAG}`,
+			`tags=${crossSeedTag}`,
 			X_WWW_FORM_URLENCODED,
 		);
 	}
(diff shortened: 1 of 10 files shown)


zakkarry:
We are generally pretty hesitant to add additional config options, particularly since this issue was brought up specifically for qBittorrent and the other clients do not have tags.

If you could outline in your PR description exactly how and when this is going to be applied to the clients without tags, and in what scenarios this would be used and desired for the clients other than qbittorrent, I think that would help describe your implementation without having to actually read the code and decipher what it is doing and where, as well as allow us to review better - since we know your intention - in the case of any potential bugs.

zakkarry:
Also, rTorrent utilizes the TORRENT_TAG as well, this wasn't mentioned?

zakkarry:
Were you going to edit your PR description to explain what you did and stuff?

src/clients/QBittorrent.ts
```diff
@@ -425,7 +425,7 @@ export default class QBittorrent implements TorrentClient {
 						category: searcheeInfo!.category,
 					};
 			if (!isComplete) return InjectionResult.TORRENT_NOT_COMPLETE;
-			const filename = `${newTorrent.getFileSystemSafeName()}.${TORRENT_TAG}.torrent`;
+			const filename = `${newTorrent.getFileSystemSafeName()}.${crossSeedTag}.torrent`;
```
zakkarry, line 428: This replacement, while somewhat consistent, is not really required. I can't see where it would make a difference unless you save the torrent file added to qbittorrent on add/completion in the client - which it would then be reflected in the file saved to your specified directory this way. 

I think leaving this as TORRENT_TAG would make a clear distinction that would not generally affect the workflow. 

Just my opinion.

src/config.template.cjs
```diff
@@ -157,6 +157,13 @@ module.exports = {
 	 */
 	linkCategory: "cross-seed-link",
 
+	/**
+	 * Defines what tag to set on linked torrents
+	 *
+	 * Default is "cross-seed".
+	 */
+	crossSeedTag: "cross-seed",
```
zakkarry, line 165: I think given that we fallback with commander and allow nullish in Zod that this would be better set to null to imply that the default is set and no changes are being made.

The description is unclear, and given the differences in the client implementations, I think clarity on this and its behavior and effects would be beneficial.

src/configSchema.ts
```diff
@@ -126,6 +127,7 @@ export const VALIDATION_SCHEMA = z
 		dataDirs: z.array(z.string()).nullish(),
 		matchMode: z.nativeEnum(MatchMode),
 		linkCategory: z.string().nullish(),
+		crossSeedTag: z.string().nullish(),
```
zakkarry, line 130: As I said, I am not overly excited about this option name given the only client actually have a "Tags" feature is qbit, and the rest are called labels.

src/constants.ts
```diff
@@ -5,7 +5,6 @@ const packageDotJson = require("../package.json");
 export const PROGRAM_NAME = packageDotJson.name;
 export const PROGRAM_VERSION = packageDotJson.version;
 export const USER_AGENT = `CrossSeed/${PROGRAM_VERSION}`;
-export const TORRENT_TAG = "cross-seed";
```
zakkarry, line 8: I think leaving this intact and falling back to TORRENT_TAG in commander would be a better approach, as it keeps our constants in the same place, rather than having the default inline.

zakkarry (changes requested):
Still looking for discussion due to additional option being added, but a few notes that are minor and easy to apply your thoughts based on my feedback. Just suggestions after all..

<hr>
The problem I see is the behavior change when it comes to non-tag-supporting clients (all except qbit). Which was what had me asking about behavior in the first place. I'm not sure if you didn't understand I was looking for a dialogue, but you haven't responded.

All injected torrents that are linked will already go to linkCategory in Deluge, there is no reason that you can't make your label anything you want with linkCategory to begin with if you have linking enabled. As a result, this would only apply to the duplicateCategories labels. If I made my `crossSeedTag` into "seeds" it would result in `label.seeds` instead only if I wasn't linking and duplicateCategories was true.

This behavior should be noted in the description, as it could be construed as some sort of override even for linking given the ambiguity when tags themselves are not native. So this would potentially be confusing for all clients other than qBittorrent.

PR #809, opened 2024-10-08
update(logging/webhook): warn on no indexers and verbose log the request

zakkarry:
When a search is performed via the `/api/webhook` endpoint with no configured indexers, this throws an error. Since this workflow is potentially useful for autobrr-dependent instances and the error is somewhat vague, after discussions we decided to demote this to a simple warning.

In the case of no indexers with a webhook search, it will now warn, log verbosely the request, and return a 503 (service unavailable).

This leaves the opportunity for misconfigurations (albeit rare with Zod) or potentially adding this return to "all failing indexers", etc, to be warned of the issue more concisely so it can be handled, while also not leading users to think that there is a larger problem plaguing their instance.

Thoughts on messaging and status code returned by the API are needed. 

500, 501, and 503 all seem somewhat applicable. 

Notes:

Currently should be set for all failed indexers when a webhook is performed, but handling this condition in a searchCadence job is still untouched.

Zod could be used for when searchCadence is defined without indexers configured, however this would still need to be handled potentially in pipeline for instances where _ALL_ indexers are in a failure state.

- [x] Check if any torznab indexers are configured when receiving search (`/api/webhook`) call and early return
- [x] Check for all failing indexers (handled for `/api/webhook`)
- [ ] Check during search job

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -125,7 +125,6 @@ module.exports = {
 	 * for you and friendlier for trackers.
 	 * Minimum value of 30.
 	 */
-
 	delay: 30,
 
 	/**
@@ -144,7 +143,6 @@ module.exports = {
 	 * or for Windows users
 	 *     dataDirs: ["C:\\My Data\\Downloads\\Movies"],
 	 */
-
 	dataDirs: [],
 
 	/**
--- a/src/server.ts
+++ b/src/server.ts
@@ -12,6 +12,7 @@ import {
 	InjectionResult,
 	SaveResult,
 } from "./constants.js";
+import { getEnabledIndexers } from "./indexers.js";
 import { Label, logger } from "./logger.js";
 import {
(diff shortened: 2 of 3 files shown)


mmgoodnow:
what if we changed this to log a warning and `return []`?

https://github.com/cross-seed/cross-seed/blob/89564daa7aaa7a905578018975f0c0a7ef0cf7a5/src/torznab.ts#L432-L434

zakkarry:
> what if we changed this to log a warning and `return []`?
> 
> https://github.com/cross-seed/cross-seed/blob/89564daa7aaa7a905578018975f0c0a7ef0cf7a5/src/torznab.ts#L432-L434

I think then we'd still have to make adjustments and contend with the "No results found" output potentially?

I haven't traced that flow around, but I would assume it would still output something right now that isn't desirable for the situation.

PR #812, opened 2024-10-14
enhancement(inject): exponential backoff to check if injection is complete

ShanaryS:
It may take a couple seconds for the client to report that a torrent is complete. This often means that the files don't get cleaned up until the next inject cycle. This causes the creation of all searchees even though the torrent files only need to be deleted. For torrents that are being rechecked, they are still expected to be primarily handled in the next cycle especially with the upcoming resume feature.

### DIFF
--- a/src/inject.ts
+++ b/src/inject.ts
@@ -3,7 +3,10 @@ import { stat, unlink } from "fs/promises";
 import ms from "ms";
 import { basename, dirname } from "path";
 import { linkAllFilesInMetafile, performAction } from "./action.js";
-import { getClient } from "./clients/TorrentClient.js";
+import {
+	getClient,
+	waitForTorrentToComplete,
+} from "./clients/TorrentClient.js";
 import {
 	Decision,
 	DecisionAnyMatch,
@@ -106,11 +109,7 @@ async function deleteTorrentFileIfComplete(
 	torrentFilePath: string,
 	infoHash: string,
 ): Promise<void> {
-	if (
-		(await getClient().isTorrentComplete(infoHash, { retries: 6 })).orElse(
-			false,
-		)
-	) {
+	if (await waitForTorrentToComplete(infoHash)) {
 		await deleteTorrentFileIfSafe(torrentFilePath);
 	} else {
 		logger.info({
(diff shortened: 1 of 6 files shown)


mmgoodnow:
Putting the exponential backoff inside the `isTorrentComplete` implementations forces us to duplicate the retry logic. It also feels a bit weird because it gives special treatment to `true` (as opposed to `false`)—doing retries when you expect it to be false is slow because now you essentially have to wait for it to time out. 

I think `waitForTorrentToComplete` can be its own function in inject.ts that calls `isTorrentComplete`

ShanaryS:
I moved it to TorrentClient so that it can be used else where if necessary.

src/inject.ts
```diff
@@ -363,6 +380,8 @@ async function injectionAlreadyExists({
 	summary.INCOMPLETE_CANDIDATES += isComplete ? 0 : 1;
 	if (isComplete) {
 		await deleteTorrentFileIfSafe(torrentFilePath);
+	} else {
+		deleteTorrentFileIfComplete(torrentFilePath, meta.infoHash);
```
mmgoodnow, line 384: This one looks redundant?

mmgoodnow (approved):
Good idea to put it in TorrentClient

src/inject.ts
ShanaryS: `isComplete` is manually set to true in a scenario above to prevent infinite rechecks.

mmgoodnow: Would it be possible to just delete directly there rather than setting iscomplete?

ShanaryS: We make a call to client for completion at the start which most likely will be true. It gets overriden if it was false but all files are present (corrupted torrent). Doing it this way prevents an unnecessary call most times this function is ran.

mmgoodnow: Which call is unnecessary?

ShanaryS: We need to check if a torrent is complete for already exists logic. If it is, we can use that value to delete instead of the unnecessary request the wait function will make. That with the fact that it can be overriden, means it's best handle the checks like this.

PR #817, opened 2024-10-27
enhancement(announce): specify missing parameters

ShanaryS:
Example:

```
2024-10-27 16:12:24 error: [announce] Missing required params: {guid, link} in {
  name: 'ABC',
  gui: 'https://XYZ.com/torrent/download/3112917.[REDACTED]',
  lin: 'https://XYZ.com/torrent/download/3112917.[REDACTED]',
  tracker: 'XYZ'
}
```

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -30,7 +30,9 @@ const ANNOUNCE_SCHEMA = z
 		link: z.string().url(),
 		tracker: z.string().refine((tracker) => tracker.trim().length > 0),
 	})
-	.strict();
+	.strict()
+	.required()
+	.refine((data) => data.guid === data.link);
 
 const WEBHOOK_SCHEMA = z
 	.object({
@@ -231,7 +233,7 @@ async function announce(
 		const message = `Missing required params (https://www.cross-seed.org/docs/v6-migration#autobrr-update): {${formatAsList(
 			errors.map(({ path }) => path.join(".")),
 			{ sort: true, type: "unit" },
-		)}} in ${inspect(data)}`;
+		)}} in ${inspect(data)}\n${inspect(errors)}`;
 		logger.error({ label: Label.ANNOUNCE, message });
 		res.writeHead(400);
 		res.end(message);
--- a/src/server.ts
+++ b/src/server.ts
@@ -4,6 +4,7 @@ import http, { IncomingMessage, ServerResponse } from "http";
 import { pick } from "lodash-es";
 import { parse as qsParse } from "querystring";
 import { inspect } from "util";
+import { z } from "zod";
 import { checkApiKey } from "./auth.js";
 import {
 	ActionResult,
@@ -20,7 +21,14 @@ import {
 } from "./pipeline.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
-import { formatAsList, isTruthy, sanitizeInfoHash } from "./utils.js";
+import { formatAsList, sanitizeInfoHash } from "./utils.js";
+
+const ANNOUNCE_SCHEMA = z.object({
+	guid: z.string(),
+	name: z.string(),
+	link: z.string(),
+	tracker: z.string(),
+});
 
 function getData(req: IncomingMessage): Promise<string> {
 	return new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


src/server.ts
```diff
@@ -217,15 +217,14 @@ async function announce(
...
-	) {
-		const message = "Missing params: {guid, name, link, tracker} required";
+	const missingParams = [
+		"guid" in data ? "" : "guid",
+		"name" in data ? "" : "name",
+		"link" in data ? "" : "link",
+		"tracker" in data ? "" : "tracker",
+	].filter(isTruthy);
+	if (missingParams.length > 0) {
+		const message = `Missing required params: {${formatAsList(missingParams, { sort: true, type: "unit" })}} in ${inspect(data)}`;
```
mmgoodnow, line 227: can we just use zod for this validation?

src/server.ts
```diff
@@ -20,7 +21,14 @@ import {
 } from "./pipeline.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
 import { indexNewTorrents, TorrentLocator } from "./torrent.js";
-import { sanitizeInfoHash } from "./utils.js";
+import { formatAsList, sanitizeInfoHash } from "./utils.js";
+
+const ANNOUNCE_SCHEMA = z.object({
+	guid: z.string(),
+	name: z.string(),
+	link: z.string(),
```
zakkarry, line 29:
```suggestion
	link: z.string().url(),

**Review by zakkarry — CHANGES_REQUESTED** [review]
since we snatch here, this needs to be a URL

File: src/server.ts [hunk 3]
```diff
@@ -20,7 +21,27 @@ import {
...
+	.strict();
+
+const WEBHOOK_SCHEMA = z
+	.object({
+		infoHash: z.string().length(40).nullish(),
+		path: z
+			.string()
+			.nullish()
+			.refine((path) => !path || existsSync(path)),
+	})
```
zakkarry, line 42: `nullish()` is not the right primitive to use, as we shouldn't be, and don't currently, accept `null` as a value.

usage of  [`.optional()`](https://zod.dev/?id=optionals) might be better while also checking that there is at least one value after in a refine?

zakkarry (changes requested):

src/server.ts
ShanaryS: I think `verifyPath` won't be necessary as the exist check should be enough. `verifyPath` also currently requires a directory and permission checks.

zakkarry: Not sure if it makes it less readable as to what's going on, but also worth mentioning since they are all optional is [`partial()`](https://zod.dev/?id=partial)

zakkarry (approved):

src/server.ts
```diff
@@ -20,7 +21,25 @@ import {
...
+
+const ANNOUNCE_SCHEMA = z
+	.object({
+		name: z.string().refine((name) => name.trim().length > 0),
+		guid: z.string().url(),
+		link: z.string().url(),
+		tracker: z.string().refine((tracker) => tracker.trim().length > 0),
+	})
+	.strict();
+
```
zakkarry, line 34:
```suggestion
	.strict()
        .required();

**Review by zakkarry — CHANGES_REQUESTED** [review]

File: src/server.ts [hunk 5]
```diff
@@ -20,7 +21,27 @@ import {
...
+	.refine((data) => data.guid === data.link);
+
+const WEBHOOK_SCHEMA = z
+	.object({
+		infoHash: z.string().length(40),
+		path: z.string().refine((path) => !path || existsSync(path)),
+	})
+	.strict()
+	.partial()
+	.refine((data) => data.infoHash || data.path);
```
zakkarry, line 44:
```suggestion
	.refine((data) => data.infoHash || data.path);
	.refine((data) => Object.keys(data).length === 1);

**Review by zakkarry — DISMISSED** [review]

File: src/server.ts [hunk 6]
```diff
@@ -98,23 +119,12 @@ async function search(
...
-			message: `Received search request: ${inspect(criteria)}`,
-		});
+	let criteria: TorrentLocator = pick(data, ["infoHash", "path"]);
 
-		const message =
-			"A valid infoHash or accessible path must be provided (infoHash is preferred).";
+	try {
+		criteria = WEBHOOK_SCHEMA.parse(criteria) as TorrentLocator;
+	} catch {
+		const message = `A valid infoHash or accessible path must be provided (infoHash is preferred, https://www.cross-seed.org/docs/basics/daemon#set-up-automatic-searches-for-finished-downloads): ${inspect(criteria)}`;
```
bakerboy448, line 127:
```suggestion
		const message = `A valid infoHash or an accessible path must be provided (infoHash is recommended see https://www.cross-seed.org/docs/basics/daemon#set-up-automatic-searches-for-finished-downloads): ${inspect(criteria)}`;
```

zakkarry:
```suggestion
		const message = `A valid infoHash or an accessible path must be provided (infoHash is recommended: see https://www.cross-seed.org/docs/basics/daemon#set-up-automatic-searches-for-finished-downloads): ${inspect(criteria)}`;
```
probably needs a colon there too.

PR #818, opened 2024-10-28
fix(decide/cache): optimize query for using cached torrents

ShanaryS:
The previous method for getting the infohash was extremely slow. It took `500ms` to retrieve per candidate for which there could be 2000. Meaning daemon mode is overloaded for 16 minutes.

Now the decision table is only queried once and stored in a hashmap for instant lookups. These same 2000 candidates are now processed in 15s instead of the 16 minutes currently. The `fuzzyGuidLookup` would take 1s if it was relevant so that is also using the hashmap for instant lookups as well.

After making an assessment, we will add the `guid: infoHash` to the map. This is to prevent duplicate snatches for the next candidate check if `fuzzyGuidLookup` is needed, and for the next searchee to use in the case of rss/announce.

Now checking the link instead of just the guid. Announce only uses the link (no access to guid), so so we may end up snatching the same torrent twice as on some trackers they are different.

Only caching decisions with snatches. The others were used only to "speed up" processing, but essentially just bloats the database.

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -292,7 +292,7 @@ function sourceDoesMatch(searcheeTitle: string, candidateName: string) {
 export async function assessCandidate(
 	metaOrCandidate: Metafile | Candidate,
 	searchee: SearcheeWithLabel,
-	hashesToExclude: string[],
+	infoHashesToExclude: Set<string>,
 ): Promise<ResultAssessment> {
 	const { blockList, includeSingleEpisodes, matchMode } = getRuntimeConfig();
 
@@ -355,7 +355,7 @@ export async function assessCandidate(
 		return { decision: Decision.SAME_INFO_HASH, metafile };
 	}
 
-	if (hashesToExclude.includes(metafile.infoHash)) {
+	if (infoHashesToExclude.has(metafile.infoHash)) {
 		return { decision: Decision.INFO_HASH_ALREADY_EXISTS, metafile };
 	}
 
@@ -444,7 +444,7 @@ async function assessAndSaveResults(
 	metaOrCandidate: Metafile | Candidate,
 	searchee: SearcheeWithLabel,
 	guid: string,
-	infoHashesToExclude: string[],
+	infoHashesToExclude: Set<string>,
 	firstSeen: number,
 	guidInfoHashMap: Map<string, string>,
 ) {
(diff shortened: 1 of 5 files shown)


src/decide.ts
```diff
@@ -475,28 +480,43 @@ async function assessAndSaveResults(
...
  * @returns The info hash of the torrent if found
  */
-async function fuzzyGuidLookup(guid: string): Promise<string | undefined> {
+async function fuzzyGuidLookup(
+	guid: string,
+	guidInfoHashMap: Map<string, string>,
+): Promise<string | undefined> {
 	if (!guid.includes(".tv/torrent/")) return;
-	const torrentIdStr = guid.match(/\.tv\/torrent\/(\d+)\/group/)?.[1];
+	const torrentIdRegex = /\.tv\/torrent\/(\d+)\/group/;
```
mmgoodnow: where's the alt-title-unique part? after `/group/`?

src/torznab.ts
mmgoodnow, line 1: are these changes accidental?

mmgoodnow:
I'm a little lukewarm on the idea of adding a bunch of performance optimizations because one site has weird guids. I'm also pretty lukewarm on the special casing for that site, but it looks like we're already doing that and I must have forgotten when it happened. The other changes look good though. Do they minimize the problem enough that we don't need the special casing any more?

src/decide.ts
ShanaryS: `.tv/torrent/132469/group?nh=8CA2A342113D4GFJOWKP911A4044F`

It seems to be a hash of something, .torrent files are the same.

src/torznab.ts
ShanaryS: I update candidate to make size and pubDate optional. They don't exist from announce in the decide stage.

ShanaryS:
> I'm a little lukewarm on the idea of adding a bunch of performance optimizations because one site has weird guids.

The optimizations is for all queries, they were all extremely slow.

> I'm also pretty lukewarm on the special casing for that site, but it looks like we're already doing that and I must have forgotten when it happened.

The special handling for that site is pretty much necessary. It would regularly snatch the same torrent 3+ times depending on alt title.

> Do they minimize the problem enough that we don't need the special casing any more?

The other way to solve the inefficient db queries is to use an index which does solve it in the main case. However we need the hashmap for the fuzzy lookup case. So I opted to use the hashmap for both.

mmgoodnow:
Fair enough. Can you add a comment to the special casing code to that effect?

mmgoodnow (dismissed):

mmgoodnow (approved):

PR #821, opened 2024-11-01
enhancement(searchLimit): per unique queries on each indexer

ShanaryS:
This addresses two issues:
1. `cross-seed search` will now always make progress on other indexers even if some indexers were down or a new one was added.
2. Cached searches aren't counted against this limit


mmgoodnow (approved):
I remember rejecting a similar PR in the past but this seems to solve for some specific problematic indexers with very low rate limits so I think I'm in favor now.

PR #823, opened 2024-11-08
fix(rss): always check candidates from first page

ShanaryS:
Indexer timestamps cannot be trusted. They could always report less than `pageBackUntil` which means no candidates gets checked. So now we always check candidates from the first page.

Also will now end paging early if a pubDate on the current page is already below pageBackUntil. This should prevent unnecessary queries when only a single page is needed.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -510,6 +510,6 @@ export async function scanRssFeeds() {
 
 	logger.info({
 		label: Label.RSS,
-		message: `RSS scan complete - checked ${i} new candidates since ${humanReadableDate(lastRun)}`,
+		message: `RSS scan complete: checked ${i} candidates - previous run was at ${humanReadableDate(lastRun)}`,
 	});
 }
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -44,6 +44,7 @@ import {
 	getApikey,
 	getLogString,
 	getMediaType,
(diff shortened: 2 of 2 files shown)


zakkarry:
If we're going to change the way RSS jobs are processed, I think it might be time we revisit the possibility that we just track the timestamp of the last entry and always process everything until it surpasses that point?

ShanaryS:
I was considering that and it would be the best solution. We can store a global hashmap by indexer id and last seen time. If a user were to restart cross-seed, this info would be lost, but it would solve the duplicate processing.

To handle it between restarts, we would need a new column in the indexer table, probably can keep it json so it's easy to expand in the future. But I'm not sure this is valuable since most indexers don't support paging.

zakkarry:
> I was considering that and it would be the best solution. We can store a global hashmap by indexer id and last seen time. If a user were to restart cross-seed, this info would be lost, but it would solve the duplicate processing.
> 
> To handle it between restarts, we would need a new column in the indexer table, probably can keep it json so it's easy to expand in the future. But I'm not sure this is valuable since most indexers don't support paging.

I think storing this on a run-by-run basis would be a fine solution, and then upon restart - when RSS is ran - we just reparse the first page and start the timestamp over in the instance.

I don't think its necessary that we store it persistently beyond what we are already doing, it's relatively low cost to process the first page once.

PR #826, opened 2024-11-16
enhancement(rss): better pagination using last seen candidates

ShanaryS:
Stores the last seen guid per indexer in a new `rss` table. This should handle all previous edge cases of rss processing.

### DIFF
--- a/src/migrations/08-rss.ts
+++ b/src/migrations/08-rss.ts
@@ -7,7 +7,7 @@ async function up(knex: Knex.Knex): Promise<void> {
 			.references("id")
 			.inTable("indexer")
 			.primary();
-		table.string("last_seen");
+		table.string("last_seen_guid");
 	});
 }
 
--- a/src/migrations/08-rss.ts
+++ b/src/migrations/08-rss.ts
@@ -1,0 +1,18 @@
+import Knex from "knex";
+
+async function up(knex: Knex.Knex): Promise<void> {
+	await knex.schema.createTable("rss", (table) => {
+		table
+			.integer("indexer_id")
+			.references("id")
+			.inTable("indexer")
+			.primary();
+		table.string("last_seen");
+	});
+}
+
+async function down(knex: Knex.Knex): Promise<void> {
+	return knex.schema.dropTable("rss");
+}
+
+export default { name: "08-rss", up, down };
--- a/src/torznab.ts
+++ b/src/torznab.ts
@@ -348,6 +348,7 @@ export function indexerDoesSupportMediaType(
 
 export async function* rssPager(
 	indexer: Indexer,
+	timeSinceLastRun: number,
 ): AsyncGenerator<Candidate, void, undefined> {
 	const limit = indexer.limits.max;
 	const lastSeenGuid: string | undefined = (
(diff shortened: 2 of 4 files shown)


zakkarry:
I'm curious, if a torrent is deleted after being newly uploaded that is the "last seen" - will this page back to its max?

ShanaryS:
Yes, the only way to avoid that is to use the time, which should be fine as it would be internal per tracker. Potential issues if the users change timezone though.

I would prefer to do it that way as the last_seen torrent is more likely to be deleted than the user changing timezones, but it's rare enough to not be a big deal.

src/migrations/08-rss.ts
```diff
@@ -0,0 +1,18 @@
+import Knex from "knex";
+
+async function up(knex: Knex.Knex): Promise<void> {
+	await knex.schema.createTable("rss", (table) => {
+		table
+			.integer("indexer_id")
+			.references("id")
+			.inTable("indexer")
+			.primary();
+		table.string("last_seen");
```
mmgoodnow, line 10: would there a way to use both guid _and_ datetime? I saw some chatter in #development about releases getting deleted which seems like a valid concern.

mmgoodnow: if it ends up that datetime _is_ just better, sorry for the thrash 😬

zakkarry: > would there a way to use both guid _and_ datetime? I saw some chatter in #development about releases getting deleted which seems like a valid concern.

Yes, my concern was that the release we are triggering completion off of could be removed fornsomething earoy in its life, for instance beong banned group, mislabeled, or invalid in some other way.

The result would be full paging back.

I think we could set a time max of like an hour before the previously seen release.

mmgoodnow: It does seem kind of pathological that we could accidentally page back all 10 pages looking for a guid that doesn't exist.

zakkarry: Exactly

ShanaryS: We could check for guid always then fallback to time if missing. Storing the last seen time is still susceptible to timezones changes, though unlikely to happen simultaneously.

Since we would only need to cover the deleted torrent case, we could take the time since last run and subtract that from the current latest candidate as the limit. If the tracker uploads frequently then this time will be perfect. If they update rarely then the first page likely covers the time range.

mmgoodnow: I like that idea.

zakkarry: sounds fine

mmgoodnow: Can we call this `last_seen_guid`?

src/torznab.ts
```diff
@@ -348,80 +348,101 @@ export function indexerDoesSupportMediaType(
...
-				query: { t: "search", q: "", limit, offset: i * limit },
-			});
+			currentPageCandidates = (
+				await makeRequest({
+					indexerId: indexer.id,
+					baseUrl: indexer.url,
+					apikey: indexer.apikey,
+					query: { t: "search", q: "", limit, offset: i * limit },
+				})
+			).sort((a, b) => b.pubDate - a.pubDate);
```
mmgoodnow, line 375: could be a good application of `comparing`

mmgoodnow (dismissed):

mmgoodnow:
LGTM but I do still think `last_seen_guid` is probably a better column name.

src/torznab.ts
ShanaryS: I think it's mainly beneficial when doing multiple sorts, I find this more straightfoward.

mmgoodnow (approved):

PR #827, opened 2024-11-24
Always restrict fuzzySizeThreshold

mmgoodnow:
Not sure why we would want to have different rules for search vs daemon here. @ShanaryS was there a reason behind this? Otherwise I'll just make them consistent with this PR.


ShanaryS:
If users are using `cross-seed inject`, we don't need to restrict fuzzy to `0.1` since there is no snatching happening. This would be if a user manually downloads a torrent (such as for ensemble matching) or is using `cross-seed inject` as a standalone tool.

ShanaryS:
I think your change here is good if we can have the ‘cross-seed inject` command ignore it. So the use case still exists but prevents search/rss/announce/webhook from ever having it.

mmgoodnow:
Gonna close this for now then.

PR #828, opened 2024-11-24
add(api/status): introduce status endpoint for health checks

zakkarry:
There is no endpoint for checking health via the API.

This PR introduces a `/api/status` endpoint that returns "200 OK" and does not log to the terminal.

Health checks that are performed frequently would not need to be logged, as they are always going to succeed assuming cross-seed is running anyway, and would just be "spam".

Currently requires API auth same as any other API call.

Applications that could utilize this are things like uptimekuma and notifarr.

This could also be expanded in the future to return any current jobs down the road if that was something we desired to do (such as "performing search" type of status). Currently it just acts as a ping endpoint to check if cross-seed is operational.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -276,25 +276,33 @@ async function announce(
 		res.end(e.message);
 	}
 }
-
+async function statusCheck(
+	req: IncomingMessage,
+	res: ServerResponse,
+): Promise<void> {
+	res.writeHead(200);
+	res.end("OK");
+}
 async function handleRequest(
 	req: IncomingMessage,
 	res: ServerResponse,
 ): Promise<void> {
 	if (!(await authorize(req, res))) return;
+	const endpoint = req.url!.split("?")[0];
 
-	if (req.method !== "POST") {
+	if (req.method !== "POST" && endpoint !== "/api/status") {
 		res.writeHead(405);
 		res.end("Methods allowed: POST");
 		return;
 	}
 
-	const endpoint = req.url!.split("?")[0];
 	switch (endpoint) {
 		case "/api/webhook":
 			return search(req, res);
 		case "/api/announce":
 			return announce(req, res);
+		case "/api/status":
+			return statusCheck(req, res);
 		default: {
 			const message = `Unknown endpoint: ${endpoint}`;
 			logger.error({ label: Label.SERVER, message });
--- a/src/server.ts
+++ b/src/server.ts
@@ -290,7 +290,10 @@ async function handleRequest(
 	if (!(await authorize(req, res))) return;
 	const endpoint = req.url!.split("?")[0];
 
-	if (req.method !== "POST" && endpoint !== "/api/status") {
+	if (
+		(req.method === "POST" && endpoint === "/api/status") ||
+		(req.method === "GET" && endpoint !== "/api/status")
+	) {
 		res.writeHead(405);
 		res.end("Methods allowed: POST");
 		return;


ShanaryS (approved):

src/server.ts
```diff
@@ -276,25 +276,33 @@ async function announce(
...
+	res: ServerResponse,
+): Promise<void> {
+	res.writeHead(200);
+	res.end("OK");
+}
 async function handleRequest(
 	req: IncomingMessage,
 	res: ServerResponse,
 ): Promise<void> {
 	if (!(await authorize(req, res))) return;
```
mmgoodnow, line 290: I could see an argument for not requiring auth but we can start like this and see what people say.

src/server.ts
```diff
@@ -276,25 +276,33 @@ async function announce(
...
+	const endpoint = req.url!.split("?")[0];
 
-	if (req.method !== "POST") {
+	if (req.method !== "POST" && endpoint !== "/api/status") {
```
mmgoodnow, line 293: this leaves the door open for /api/status to be called with either POST or GET. we should pick one and enforce it

src/server.ts
zakkarry: There should be no situation when someone who _couldn't_ provide the api-key would want to check the status of cross-seed, and in the future if we add information on jobs running/scheduling etc, then we should absolutely require auth.

It can easily be passed as a param in the URL, so there's no circumstance where a service like kuma or notifarr cannot provide a ?apikey= and it be perfectly fine.

I don't want to start without an apikey now, and then add it later if we expand the return.

src/server.ts
zakkarry: I was aware of this when writing it, but it would obviously be a GET in this case, and I assumed no one would be POST'ing to it, and even if so it would still return 200 OK.

I'll add GET exclusivity.

mmgoodnow (approved):

buroa:
Can we remove auth for this call? cc @onedr0p

frantathefranta:
Just like @buroa, I run cross-seed in a k8s cluster. Having a native healthcheck for a pod is pretty nice and to me it seems they are rarely protected by an API key, even if the rest of the API endpoints are. One example is [autobrr](https://autobrr.com/api#health-check-endpoints).

zakkarry:
> Just like @buroa, I run cross-seed in a k8s cluster. Having a native healthcheck for a pod is pretty nice and to me it seems they are rarely protected by an API key, even if the rest of the API endpoints are. One example is [autobrr](https://autobrr.com/api#health-check-endpoints) (the examples show the use of an `X-API-Token` header but they work without it).

That's cool. Honestly, this is rather frustrating to have to say _HERE_ again, given that I just suggested you guys handle this appropriately for the use-case in a FR, but let me really go into detail since this seems to be ambiguous or unclear still.

If you'd like to save some time reading this longer response, I made a response to the _EXACT SAME_ thing just mentioned and gave my suggestion if you wanted a _BASIC_ health check endpoint. If you're unfamiliar with how feature requests would work on OSS and GitHub, you would make an Issue for a feature request for something like an `/api/ping` endpoint. My previous response and multiple other places in this PR explain why that is not what I planned for this endpoint from the start.

<hr>

I have a clear direction I am planning/going with this endpoint, this is the initial PR to get the barebones active, and I outline exactly what sort of use I envision in this PR (linked comments in a previous reply), and despite "health check" being present in the title of this PR (which I guess the lot of you see and cannot resist posting before reading any/all of the details) there is significantly more about what my plans and goals are for this, explaining exactly why this is not intended to be _JUST_ a basic health check, for instance. **This is why I didn't choose `/api/ping` which would align with _THAT_ purpose and other notable services' endpoint naming.**

Honestly, despite the almost constant push-back from the LITERAL handful of k8's users we have in regards to auth since its implementation, we have tried to accommodate where we can  - and compromise when needed. In this PR, however, it seems like you guys are disregarding my multiple comments relevant to this decision that thoroughly explain the use case I intend and outline what would be appropriate if you had just _SUDDENLY_ had the epiphany that health checks could exist in cross-seed, despite no one even mentioning it prior...ever.

<hr>
**tl;dr-ish**

Once again, from my perspective, it looks like you saw my PR, read the words health check, and didn't bother to read details beyond "auth" before doing the k8's ideology dance. You, thus, incorrectly assumed that it is intended to be a simple health check and nothing more and think that I arbitrarily put it behind auth.

That was, is, and won't be the plan I have for this specific endpoint. I created this with the purpose I have detailed in multiple comments, that is why it is `/api/status` and not `/api/ping`. I made that clear when I said exactly that, outlined why I was putting it behind auth, and detailed the functionality I wanted when it was completed.

If you have issues or want to make a feature request, then there is a section of the repo specifically for that. Commenting your discontentedness with auth as a philosophy and concept in a PR that has been open for days and already merged is not the process to accomplish what you seem to want.

zakkarry:
> Just like @buroa, I run cross-seed in a k8s cluster. Having a native healthcheck for a pod is pretty nice and to me it seems they are rarely protected by an API key, even if the rest of the API endpoints are. One example is [autobrr](https://autobrr.com/api#health-check-endpoints) (the examples show the use of an `X-API-Token` header but they work without it).

Please see my previous post in response to @buroa for context as to the intended use case of this endpoint. 

As I suggested to @buroa - he has made #828 to meet his (and your) use case. 

Please use that FR/Issue for your communication in regards to specifics you may want to have added/included and further discussions regarding an authless basic/simple health check endpoint.

PR #833, opened 2024-11-27
feat: add simple /api/ping endpoint

buroa:
Per me and @zakkarry comments in #828

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -82,6 +82,10 @@ async function authorize(
 	req: IncomingMessage,
 	res: ServerResponse,
 ): Promise<boolean> {
+	/**
+	 * checks all http API requests for authorized apiKey
+	 * uses param `?apikey=` or as header `x-api-key`
+	 */
 	const url = new URL(req.url!, `http://${req.headers.host}`);
 	const apiKey =
 		(req.headers["x-api-key"] as string) ?? url.searchParams.get("apikey");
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -276,18 +276,25 @@ async function announce(
...
 	res.writeHead(200);
 	res.end("OK");
 }
+
+async function ping(req: IncomingMessage, res: ServerResponse): Promise<void> {
+	res.writeHead(200);
+	res.end("OK");
+}
+
 async function handleRequest(
```
ShanaryS, line 293: Since an endpoint without auth is being added, this function should refactored:
```ts
async function handleRequest(
	req: IncomingMessage,
	res: ServerResponse,
): Promise<void> {
	const allowedMethods = ["POST", "GET"];
	if (!allowedMethods.includes(req.method!)) {
		res.writeHead(405);
		res.end(
			`Methods allowed: ${formatAsList(allowedMethods, { sort: true })}`,
		);
		return;
	}

	const endpoint = req.url!.split("?")[0];
	switch (endpoint) {
		case "/api/announce":
			if (!(await authorize(req, res))) return;
			return announce(req, res);
		case "/api/webhook":
			if (!(await authorize(req, res))) return;
			return search(req, res);
		case "/api/ping":
			return ping(req, res);
		case "/api/status":
			if (!(await authorize(req, res))) return;
			return status(req, res);
		default: {
			const message = `Unknown endpoint: ${endpoint}`;
			logger.error({ label: Label.SERVER, message });
			res.writeHead(404);
			res.end(message);
			return;
		}
	}
}
```

ShanaryS (approved):

mmgoodnow (approved):
it's probably time to take a dep on `fastify`.

PR #836, opened 2024-12-07
update(config-template): clarify usage across data and torrent for option descriptions

zakkarry:
updates language used in the config template descriptions to more accurately explain when the option is relevant, specifically when using the term "torrent" to mean any searches.

also updated the tagging behavior when using linking in the duplicateCategories option, as this has come up previously too as ambiguous if linking is disabled.

fixes @bakerboy448 

oh wait cant fix bakerboy :P

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -237,7 +237,7 @@ module.exports = {
 	outputDir: ".",
 
 	/**
-	 * Whether to include single episode torrents in a search (not those from
+	 * Whether to include single episodes during a search (not those from
 	 * season packs).
 	 *
 	 * This setting does not affect matching episodes from rss and
@@ -246,7 +246,7 @@ module.exports = {
 	includeSingleEpisodes: false,
 
 	/**
-	 * Include torrents which are comprised of non-video files.
+	 * Include torrents/data comprised of non-video files.
 	 *
 	 * If this option is set to false, any folders or torrents whose
 	 * totalNonVideoFilesSize / totalSize > fuzzySizeThreshold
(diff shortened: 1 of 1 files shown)


src/config.template.cjs
```diff
@@ -321,7 +322,8 @@ module.exports = {
 	 * Whether to inject using the same labels/categories as the original
 	 * torrent.
 	 *
-	 * qBittorrent: This will apply the category's original category as a tag.
+	 * qBittorrent: When using linking, this will apply the category's original
+	 * category as a tag.
```
bakerboy448, line 326:
```suggestion
	 * category as a tag when duplicate categories is true.
```
?

bakerboy448 (approved):
1 tweak else LGTM

src/config.template.cjs
zakkarry: IIRC the only difference is that with true, .cross-seed is appended to the tag, doesn't false still tag the original category without .cross-seed appended?

Frankly, @ShanaryS should be doubled checking the qBit documentation - as he was the last one to fix up the bullshit tag autotmm category non-sense you guys had requested lol

Either way, I'll fix it to what I think it is, but @ShanaryS will have to give the OK on this, you know since you don't read the docs @bakerboy448

zakkarry: Pushed a more concise and thorough description of all the behavior, but needs validation.

ShanaryS: `duplicateCategories: false` will not apply any extra tags. `duplicateCategories: true` will apply `category.cross-seed`.

zakkarry: Ok, well can you push the corrections as you had last had them?

PR #841, opened 2024-12-17
enhancement(torrentDir): verify contents against client

ShanaryS:
This is better than simply checking for the presence of a single .torrent file. Users have frequently set `torrentDir` to the directory their client copies .torrent files to, instead of the one that mirrors the client state.

I also added the check for searchee blocklist before testing the save path. It's still only being logged until the new blocklist features are included.

### DIFF
--- a/src/pipeline.ts
+++ b/src/pipeline.ts
@@ -464,7 +464,6 @@ async function findSearchableTorrents(): Promise<{
 		}
 		grouping.get(key)!.push(searchee);
 	}
-	const keysToDelete: string[] = [];
 	for (const [key, groupedSearchees] of grouping) {
 		// If one searchee needs to be searched, use the candidates for all
 		const filteredSearchees = filterDupesFromSimilar(groupedSearchees);
@@ -472,7 +471,7 @@ async function findSearchableTorrents(): Promise<{
 			filteredSearchees.map(filterTimestamps),
 		);
 		if (!results.some(isTruthy)) {
-			keysToDelete.push(key);
+			grouping.delete(key);
 			continue;
 		}
 		filteredSearchees.sort(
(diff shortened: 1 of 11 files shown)


mmgoodnow:
after this is there still an advantage to using torrentDir rather than just directly polling the client for searchees (which I would support, fwiw, because it means less configuration)?

src/pipeline.ts
```diff
@@ -464,15 +464,14 @@ async function findSearchableTorrents(): Promise<{
...
-	const keysToDelete: string[] = [];
 	for (const [key, groupedSearchees] of grouping) {
 		// If one searchee needs to be searched, use the candidates for all
 		const filteredSearchees = filterDupesFromSimilar(groupedSearchees);
 		const results = await Promise.all(
 			filteredSearchees.map(filterTimestamps),
 		);
 		if (!results.some(isTruthy)) {
-			keysToDelete.push(key);
+			grouping.delete(key);
```
mmgoodnow, line 474: nice cleanup

mmgoodnow (approved):
I think I initially misunderstood—seems like this is more of a follow-up to #835

zakkarry:
> I think I initially misunderstood—seems like this is more of a follow-up to #835

It basically prevents the users who do not use the torrentDir as their actual client dir and use something like this instead.

image

from encountering torrent files in a search that are not present in the client. and TORRENT_NOT_FOUND errors during injection.

kilinrax:
This change seems to break my setup; using a NFS mounted torrentDir under k8s.

Ideally this feature should be disable-able via config.js, otherwise I'm stuck on version 6.2.2.

ShanaryS:
@kilinrax What do you mean "break" your setup? It sounds like it was never valid in the first place.

PR #842, opened 2024-12-20
fix(zod/linkdir): do not require linkdir for risky or partial when action is save

zakkarry:
using save should allow for `linkDir` to not be set, as it doesn't link.

this also prevents users who use things like autotorrent, or manually add the torrents and links (if needed) from using risky or partial.

there is no reason to require a setting that is not used for `matchMode`.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -315,7 +315,10 @@ export const VALIDATION_SCHEMA = z
 		ZodErrorMessages.needsTorrentDir,
 	)
 	.refine(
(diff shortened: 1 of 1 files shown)


mmgoodnow (approved):

zakkarry:
> There's an argument that there's no other software that can link and inject risky/partial/ensemble matches but I agree that we can still leave the door open for it.

It doesn't stop me from using `ln` and making these links myself, or renaming the torrent after adding paused.

PR #846, opened 2024-12-21
update(cli/ensemble): add cli arg --no-season-from-episodes

zakkarry:
since `seasonFromEpisodes` defaults to 1 - there is not any way to use cli to disable it.

this adds `--no-season-from-episodes`

documentation has been updated for this

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -344,7 +344,10 @@ export async function performAction(
 ): Promise<{ actionResult: ActionResult; linkedNewFiles: boolean }> {
 	const { action, linkDir } = getRuntimeConfig();
 
-	if (action === Action.SAVE) {
+	if (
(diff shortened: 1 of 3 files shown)


mmgoodnow:
We accidentally made a breaking change with 6.5 - upgrading from 6.0.0 to 6.5.0 should not make anybody's cross-seed instance fail to start. This added parameter is good, but we should also make this change backwards compatible.

zakkarry:
> We accidentally made a breaking change with 6.5 - upgrading from 6.0.0 to 6.5.0 should not make anybody's cross-seed instance fail to start. This added flag is good, but we should also make this change backwards compatible.

I'm not sure what specifically you mean? Sorry...

mmgoodnow (approved):

PR #847, opened 2024-12-21
update(config-template): Fix comment for disabling seasonFromEpisodes

Shraymonks:
Fix comment to clarify that only `null` disables `seasonFromEpisodes`. Since it is set using [`fallback`](https://github.com/cross-seed/cross-seed/blob/19054f3b9424e6850c4ddc06c4fbeab091b3f488/src/cmd.ts#L195), `undefined` actually sets it to `1`.


zakkarry:
Thanks, i did the website but forgot about the template.

PR #849, opened 2024-12-21
update(blocklist/deluge): change labels to be category rather than tag

zakkarry:
Labels were previously associated with the `tag:` prefix in blocklist, the rationale was that there might be a time when labels are able to be applied in more than one and the type was a string.

This will not be the case, Deluge uses labels in the same way that qBit does, for save path, bandwidth and queue, etc. Multiple labels will not ever be a concern.

This changes labels to category. Tags are assigned empty array, since no tags exist in Deluge.

- [x] determine if No Label (and Uncategorized in qBit) need a constant to distinguish them, both return `""` for this assignment (or rather lack of)

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -154,6 +154,8 @@ function transformBlocklist(blockList: string[], ctx: RefinementCtx) {
 		}
 		const { blocklistType, blocklistValue } = parseBlocklistEntry(blockRaw);
 		switch (blocklistType) {
+			case BlocklistType.NAME:
+				break;
 			case BlocklistType.FOLDER:
 				if (/[/\\]/.test(blocklistValue)) {
 					addZodIssue(
@@ -169,6 +171,20 @@ function transformBlocklist(blockList: string[], ctx: RefinementCtx) {
 					new RegExp(blocklistValue);
 				} catch (e) {
 					addZodIssue(blockRaw, ZodErrorMessages.blocklistRegex, ctx);
+				}
+				break;
+			case BlocklistType.CATEGORY:
+				if (blocklistValue.length === 0) {
+					logger.info(
+						`Blocklisting all torrents without a category due to empty ${blockRaw}`,
+					);
+				}
+				break;
+			case BlocklistType.TAG:
+				if (blocklistValue.length === 0) {
+					logger.info(
+						`Blocklisting all torrents without a tag due to empty ${blockRaw}`,
+					);
 				}
 				break;
 			case BlocklistType.TRACKER:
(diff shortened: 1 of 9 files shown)


mmgoodnow:
rtorrent should be tag, flood supports multiple (they may be comma separated in custom1)

src/configSchema.ts
```diff
@@ -171,6 +173,20 @@ function transformBlocklist(blockList: string[], ctx: RefinementCtx) {
...
+					logger.info(
+						`Blocklisting all torrents without a category due to empty ${blockRaw}`,
+					);
+				}
+				break;
+			case BlocklistType.TAG:
+				if (blocklistValue.length === 0) {
+					logger.info(
+						`Blocklisting all torrents without a tag due to empty ${blockRaw}`,
+					);
```
mmgoodnow, line 187: kinda feel like empty array should not block anything, both intuitively and also because you can't combine it with other stuff - like if you want to block no category and also block books, this doesn't let you do that

 what about something like `[null]` or `[""]`?

src/configSchema.ts
```diff
@@ -433,6 +445,34 @@ export const VALIDATION_SCHEMA = z
...
+				`${BlocklistType.TAG}: blocklisting is deprecated for Deluge, please use ${BlocklistType.CATEGORY}: instead (all ${BlocklistType.TAG}: blocklist items has being automatically converted to ${BlocklistType.CATEGORY}:).`,
+			);
+			config.blockList = config.blockList.map((b) => {
+				if (b.startsWith(`${BlocklistType.TAG}:`)) {
+					return b.replace(BlocklistType.TAG, BlocklistType.CATEGORY);
+				}
+				return b;
+			});
+		}
+		if (
```
mmgoodnow, line 463: I wouldn't worry about backwards compatibility as much in this situation because it's very new. there's an argument it's just a bug fix

mmgoodnow: Fine either way though

src/configSchema.ts
ShanaryS: This is not when `blockList: []`, but for `blockList: ["category:", "tag:", "category:books"]`. The first two blocks torrents without categories or tags.

If a user tried the first two before this PR, config validation would have failed (or said they are using legacy linking with raw blocklist of `category:` or `tag:`)

mmgoodnow: ohhhh I see yeah that makes sense then

src/configSchema.ts
ShanaryS: This is no cost to do since it's handling during validation and can be removed in a few versions.

mmgoodnow (approved):

PR #851, opened 2024-12-22
fix(config/seasonFromEpisodes): better validation

ShanaryS:
In case of `seasonFromEpisodes: undefined` (or missing from config), it will now default to `null`, preventing validation errors. Even though a default of `1` is fine for most users, the error has confused others.

Will now warn if using `action: save` due to the complexity of these matches.

Ensemble requires clients to be connectable if torrentDir is enabled, this will now be checked during validation.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -192,7 +192,7 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			"--season-from-episodes <decimal>",
 			"Match season packs from episode torrents",
 			parseFloat,
-			fallback(fileConfig.seasonFromEpisodes, 1),
+			fallback(fileConfig.seasonFromEpisodes, null),
 		)
 		.option(
 			"--no-season-from-episodes",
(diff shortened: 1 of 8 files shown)


src/cmd.ts
```diff
@@ -192,11 +192,7 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			"--season-from-episodes <decimal>",
 			"Match season packs from episode torrents",
 			parseFloat,
-			fallback(fileConfig.seasonFromEpisodes, 1),
-		)
-		.option(
-			"--no-season-from-episodes",
-			"Don't match season packs from episode torrents",
+			fallback(fileConfig.seasonFromEpisodes, null),
 		)
```
mmgoodnow, line 196: we should keep this for overriding the file config

ShanaryS: Yep, I just realized this myself.

mmgoodnow (approved):

PR #852, opened 2024-12-23
enhancement(config/blockList): prevent useless blockList entries

ShanaryS:
This prevents users from using types that their config does not support.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -513,30 +513,45 @@ export const VALIDATION_SCHEMA = z
 		}
 		return true;
 	})
-	.refine(
-		(config) =>
+	.refine((config) => {
+		if (
 			!config.blockList.some(
 				(b) =>
 					b.startsWith(`${BlocklistType.CATEGORY}:`) ||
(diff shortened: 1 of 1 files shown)


mmgoodnow (approved):

src/configSchema.ts
```diff
@@ -464,18 +501,42 @@ export const VALIDATION_SCHEMA = z
...
+		ZodErrorMessages.blocklistNeedsClient,
+	)
+	.refine(
+		(config) =>
+			!config.blockList.some(
+				(b) =>
+					b.startsWith(`${BlocklistType.FOLDER}:`) ||
+					b.startsWith(`${BlocklistType.FOLDER_REGEX}:`),
+			) || config.dataDirs?.length,
+		ZodErrorMessages.blocklistNeedsDataDirs,
```
mmgoodnow, line 538: can we add these as warning and promote to error in v7? right now they're not really harming anything and accidentally cross-seeding a couple extra things you didn't want is probably preferable to crash looping

PR #853, opened 2024-12-23
enhancement(inject): skip recheck for complete data/ensemble

ShanaryS:
This is primarily for racing ensemble from announce.

With the new `searchee.mtimeMs` for data/ensemble searchees, we should be able to tell if complete in most scenarios. The only exception is for stalled torrents, which should be rare. The inject job already handles and sometimes inject stalled torrents anyways.

Will always save data/ensemble .torrent files anyways in case there is an issue, the inject job will automatically resolve it.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -402,7 +402,7 @@ export async function performAction(
 	logActionResult(result, newMeta, searchee, tracker, decision);
 	if (result === InjectionResult.SUCCESS) {
 		// cross-seed may need to process these with the inject job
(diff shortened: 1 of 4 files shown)


zakkarry:
Are we not assuming that data that combines with torrent could be incomplete anymore? I guess it will recheck and be left paused, but what will it do - just recheck when inject runs and no files were missing?

mmgoodnow (approved):

PR #856, opened 2024-12-25
ehancement(pushNotifier): additional info in response

ShanaryS:
This estimates if the torrent will be paused when sending the push notification. Since we aren't relying on the result of a recheck (which could be a while), this can be off by a few pieces.

New fields added to existing `extra` field:
`paused` - boolean for if the torrent will be paused and require user intervention to resume
`decisions` - array of decisions that corresponds to the injections
`searchee` - an object with information about the source match
- `category` - the category of the source torrent for the match, or `null` if it wasn't parsed
- `tags` - an array of strings for containing the tags from the source torrent, or `null` if it wasn't parsed
- `trackers` - a nested array `[[]]` of trackers from the source torrent, or `null` if it wasn't parsed
- `length` - the size of the source torrent in bytes
- `infoHash` - the infoHash of the source if torrent based or `null` otherwise
- `path` - The path of the source if data based or `null` otherwise
- `source` - `torrentDir` `dataDirs` `virtual` (virtual is seasonFromEpisodes)


mmgoodnow (approved):

src/pushNotifier.ts
```diff
@@ -72,58 +72,94 @@ export function sendResultsNotification(
...
 				result: failures[0][2],
+				paused: false,
+				decisions,
+				searcheeCategory,
+				searcheeTags,
+				searcheeTrackers,
+				searcheeLength,
+				searcheeInfoHash,
+				searcheePath,
+				searcheeSource,
```
mmgoodnow, line 162: would it make sense to group these together into a searchee object?

PR #857, opened 2024-12-28
enhancement(dataDirs): support rss/announce matching

ShanaryS:
This uses `fs.watcher`, which recently supported recursive watch all on platforms in Node v19. This detects new/removed files and modified files, which is all we need. We do not run it all the time, as soon as a change is detected we cancel it for that dataDir. Then on next rss/announce we will reindex only the dataDirs with a change and restart the watcher.

The indexing is stored in memDB, essentially the same as the torrents table for lookup. But we parse the title instead of name so that media season folders and others are meaningful. As with ensemble, memDB makes the lookups much quicker and is only a couple megabytes (when on disk) for 20k entries. Also adds a few seconds for indexing at startup.

Recommended `maxDataDepth` is 3 if using on media libraries. This is so that `includeSingleEpisodes` and `seasonFromEpisodes` can reach individual episodes. The season pack episode filter (all tv library episodes) does not apply to rss/announce for data, only the normal `includeSingleEpisodes` does. `seasonFromEpisodes` is technically does not need `maxDataDepth` 3 as it's covered by the season searchee, however it can potentially source from torrents not currently imported.

Since we will be indexing dataDirs a lot, I tried to minimize the load on the disk, but of course it's far more intensive that torrentDir. I added memoization to the `File[]` creation as there is a lot of repeated work when creating searchees of parent folders. Full searchees are only created once the reverse lookup is successful.

### DIFF
--- a/src/dataFiles.ts
+++ b/src/dataFiles.ts
@@ -1,8 +1,155 @@
-import { readdirSync, statSync } from "fs";
-import { basename, extname, join } from "path";
+import { existsSync, FSWatcher, readdirSync, statSync, watch } from "fs";
+import Fuse from "fuse.js";
+import { basename, dirname, extname, join } from "path";
 import { IGNORED_FOLDERS_SUBSTRINGS, VIDEO_EXTENSIONS } from "./constants.js";
+import { memDB } from "./db.js";
 import { logger } from "./logger.js";
 import { getRuntimeConfig } from "./runtimeConfig.js";
+import {
+	createSearcheeFromPath,
+	File,
+	getFilesFromDataRoot,
+	parseTitle,
+	SearcheeWithoutInfoHash,
+} from "./searchee.js";
+import { createEnsemblePieces, EnsembleEntry } from "./torrent.js";
+import { createKeyTitle } from "./utils.js";
+import { isOk } from "./Result.js";
+
+interface DataEntry {
+	title: string;
+	path: string;
+}
+
+const watchers: Map<string, FSWatcher | null> = new Map();
+
+function startWatcher(dataDir: string): FSWatcher {
+	return watch(dataDir, { recursive: true, persistent: false }, () => {
+		watchers.get(dataDir)!.close();
+		watchers.set(dataDir, null);
+	});
+}
+
+export async function initializeDataDirs(): Promise<void> {
+	const { dataDirs } = getRuntimeConfig();
+	if (await memDB.schema.hasTable("data")) return;
+	await memDB.schema.createTable("data", (table) => {
+		table.string("path").primary();
+		table.string("title");
+	});
+	if (!dataDirs?.length) return;
+	for (const dataDir of dataDirs) {
+		watchers.set(dataDir, null);
+	}
+	logger.info("Indexing dataDirs for reverse lookup...");
+	await indexDataDirs();
+}
+
+async function indexDataDir(
+	dataDir: string,
+	prevWatcher: FSWatcher | null,
+): Promise<void> {
+	const { maxDataDepth, seasonFromEpisodes } = getRuntimeConfig();
+	if (prevWatcher) return;
+	logger.verbose(`Indexing ${dataDir} due to recent changes...`);
+	watchers.set(dataDir, startWatcher(dataDir));
+	for (const dirent of readdirSync(dataDir)) {
+		const entry = join(dataDir, dirent);
+		const paths = findPotentialNestedRoots(entry, maxDataDepth);
+		const dataRows: DataEntry[] = [];
+		const ensembleRows: EnsembleEntry[] = [];
+		for (const path of paths) {
+			const files = getFilesFromDataRoot(path);
+			const title = parseTitle(basename(path), files, path);
+			if (!title) continue;
+			dataRows.push({ title, path });
+			if (seasonFromEpisodes) {
+				const ensembleEntry = await indexEnsembleDataEntry(
+					title,
+					path,
+					files,
+				);
+				if (ensembleEntry) ensembleRows.push(ensembleEntry);
+			}
+		}
+		const batchSize = 100;
+		for (let i = 0; i < dataRows.length; i += batchSize) {
+			const batch = dataRows.slice(i, i + batchSize);
+			if (!batch.length) break;
+			await memDB("data").insert(batch).onConflict("path").ignore();
+		}
+		for (let i = 0; i < ensembleRows.length; i += batchSize) {
+			const batch = ensembleRows.slice(i, i + batchSize);
+			if (!batch.length) break;
+			await memDB("ensemble").insert(batch).onConflict("path").ignore();
+		}
+	}
+}
+
+export async function indexDataDirs(): Promise<void> {
+	await Promise.all(
+		Array.from(watchers.entries()).map(([dataDir, prevWatcher]) =>
+			indexDataDir(dataDir, prevWatcher),
+		),
+	);
+}
+
+async function indexEnsembleDataEntry(
+	title: string,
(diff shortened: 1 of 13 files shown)


mmgoodnow (commented):
not really understanding how the watching part works. to me it looks like the watchers just get created and then we don't do anything with them, i.e. we don't subscribe to any of their events.

mmgoodnow:
ah, I think I understand what it's for now after reading the description. 

> We do not run it all the time, as soon as a change is detected we cancel it for that dataDir. Then on next rss/announce we will reindex only the dataDirs with a change and restart the watcher.

I think I am less concerned about creating searchees of parent folders than the repeated full indexing of a dataDir. It doesn't really seem like the fs watching will give us much benefit with a granularity of "something in this dataDir changed". That said, we should probably deploy it without any further optimizations and see what people say

src/dataFiles.ts
```diff
@@ -1,8 +1,161 @@
-import { readdirSync, statSync } from "fs";
-import { basename, extname, join } from "path";
+import { existsSync, FSWatcher, readdirSync, statSync, watch } from "fs";
```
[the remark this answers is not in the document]
ShanaryS: I haven't. What benefit would it provide? All the dataDirs are Promise.all on their async functions.

src/dataFiles.ts
```diff
@@ -1,8 +1,161 @@
...
+	});
+}
+
+export async function initializeDataDirs(): Promise<void> {
+	const { dataDirs } = getRuntimeConfig();
+	if (await memDB.schema.hasTable("data")) return;
+	await memDB.schema.createTable("data", (table) => {
+		table.string("path").primary();
+		table.string("title");
+	});
```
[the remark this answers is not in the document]
ShanaryS: I'm not sure I understand.

src/dataFiles.ts
```diff
@@ -1,8 +1,161 @@
...
+	if (!dataDirs?.length) return;
+	for (const dataDir of dataDirs) {
+		watchers.set(dataDir, null);
+	}
```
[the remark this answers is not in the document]
ShanaryS: I don't think so. I did a lot of checks to see if setting the watcher to null gets rid of it. It doesn't raise anymore events and the garbage collector should clean it up.

This function here is only called at startup. The watcher callback is what sets itself to null.

src/dataFiles.ts
```diff
@@ -1,8 +1,161 @@
...
+			if (!batch.length) break;
+			await memDB("ensemble").insert(batch).onConflict("path").ignore();
+		}
+	}
+}
+
+export async function indexDataDirs(): Promise<void> {
+	await Promise.all(
+		Array.from(watchers.entries()).map(([dataDir, prevWatcher]) =>
+			indexDataDir(dataDir, prevWatcher),
```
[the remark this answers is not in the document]
ShanaryS: That was the watcher for each dataDir. It can either be null, meaning it found something and we should reindex, or FSWatcher which means we can skip indexing.

ShanaryS:
We would never need to create the searchee at indexing. We only need the path and parsed title.

I considered running watcher all the time and doing partial updates. But then we have to worry about catching problems. But I think so of the latest changes I made makes this easier so I'll look into again.

src/dataFiles.ts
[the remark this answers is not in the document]
ShanaryS: This returns an async iterator which seems harder to deal with than a callback.

ShanaryS:
Now only reindexing the affected paths. It's not enough just to use the path returned by the event, as parent searchees could also need updating.

src/dataFiles.ts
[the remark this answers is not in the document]
mmgoodnow: performing 

```ts
await memDB.schema.createTable("data", (table) => {
		table.string("path").primary();
		table.string("title");
	});
```
on memDB inline with its creation inside db.ts means we don't have to worry about the code running multiple times, and there's one place to see all of the DB schema code

src/dataFiles.ts
```diff
@@ -39,65 +36,104 @@ export async function initializeDataDirs(): Promise<void> {
 	});
 	if (!dataDirs?.length) return;
 	for (const dataDir of dataDirs) {
-		watchers.set(dataDir, null);
+		modifiedPaths.set(dataDir, new Set());
```
mmgoodnow, line 34: would it be easier for modifiedPaths to just be a Set instead of Map<string, Set>?

src/dataFiles.ts
```diff
@@ -39,65 +36,104 @@ export async function initializeDataDirs(): Promise<void> {
...
+	const batchSize = 100;
+	for (let i = 0; i < dataRows.length; i += batchSize) {
+		const batch = dataRows.slice(i, i + batchSize);
+		if (!batch.length) break;
+		await memDB("data").insert(batch).onConflict("path").merge();
+	}
+	for (let i = 0; i < ensembleRows.length; i += batchSize) {
+		const batch = ensembleRows.slice(i, i + batchSize);
+		if (!batch.length) break;
+		await memDB("ensemble").insert(batch).onConflict("path").merge();
```
mmgoodnow, line 130: knex has a batch insert util that could help here

src/dataFiles.ts
```diff
@@ -39,65 +36,104 @@ export async function initializeDataDirs(): Promise<void> {
...
+					affectedPaths.push(parentPath);
+					parentPath = dirname(parentPath);
+				}
+				for (const affectedPath of affectedPaths.slice(-maxDataDepth)) {
+					if (seenPaths.has(affectedPath)) continue;
+					seenPaths.add(affectedPath);
+					acc.push(affectedPath);
+				}
+				return acc;
+			}, []);
```
mmgoodnow, line 87: seems like we could do this processing in multiple stages (flatmap all eventPaths into their ancestry chains, dedupe by throwing everything into a Set, then do the maxdatadepth check) and maybe have to juggle fewer variables at once.

src/dataFiles.ts
```diff
@@ -39,65 +36,104 @@ export async function initializeDataDirs(): Promise<void> {
...
+	await Promise.all(
+		Array.from(watchers.keys()).map((dataDir) => {
+			const modified = modifiedPaths.get(dataDir)!;
+			const eventPaths: string[] = [];
+			while (modified.size) {
+				const path: string | undefined = modified.values().next().value;
+				if (!path) continue;
+				if (!modified.delete(path)) continue;
+				eventPaths.push(path);
+			}
```
mmgoodnow, line 63: is this equivalent to `const eventPaths = Array.from(modified).filter(isTruthy)`?

ShanaryS: We need to remove the item from the set as we will be reindexing it. I'm using the iterator as JS does not have `.pop()` on sets. The goal here is to go through all the modified paths in a concurrent safe way. As multiple indexes could be triggered at the same time from announces or rss.

src/dataFiles.ts
ShanaryS: I'm not seeing the structure you're proposing. I want to handle the deepest paths first, as that makes the most of the memoization. Though it's not as important for reindexing since not much will change between rssCadence or announces. `eventPaths` won't have duplicates, but getting the parent paths will create duplicates. Having it grouped like this makes it easy to still obey `maxDataDepth`.

src/dataFiles.ts
ShanaryS: `memDB.batchInsert("data", dataRows, batchSize)` seems like it would work but it doesn't support `.onConflict()`.

mmgoodnow (approved):

PR #858, opened 2024-12-29
use node:sqlite instead of better-sqlite3

mmgoodnow:
Late versions of Node 22 have the `sqlite` module available without a flag. When we release our next major version, we can snip cross-seed's dependency on `better-sqlite3`, which is our only native code dependency and frequently causes installation/upgrade errors, as well as being particular about the versions of Node it runs on. 

All dependencies being JS also makes it possible to bundle cross-seed into a single JS file, which, using the [SEA](https://nodejs.org/api/single-executable-applications.html)) API, will allow us to more easily provide single-executable binaries for download.


zakkarry:
I'm going to assume this would fix the node-gyp errors I've been experiencing on a certain seedbox provider when building with outdated glibc.

Interested to test this out once it passes checks.

PR #860, opened 2024-12-30
fix(zod/webhook): corrects bad logic on path validation in webhook api calls

zakkarry:
This corrects bad logic on a throw which allows for falsey path (empty string) to not return 400.

Throws now on:

path string is falsy (e.g., '', null, or undefined)
file or directory specified does not exist

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -37,7 +37,7 @@ const ANNOUNCE_SCHEMA = z
 const WEBHOOK_SCHEMA = z
 	.object({
 		infoHash: z.string().length(40),
-		path: z.string().refine((path) => !path || existsSync(path)),
+		path: z.string().refine((path) => path && existsSync(path)),
 	})
 	.strict()
 	.partial()


zakkarry:
> This was done to not fail if path was omitted. But partial may fix it or it wasn't needed in the first place.

I had mentioned this fact in #dev but wasn't 100% that it was done with reason, 

I've tested this though, both infohash and path work fine, and path behaves correctly now.

mmgoodnow (approved):

mmgoodnow:
Hmm actually, this is a breaking change for anyone with dangling `path`s in their query params, like `?infoHash=abcd&path=`. Is there a problem that prompted this?

zakkarry:
> Hmm actually, this is a breaking change for anyone with dangling `path`s in their query params, like `?infoHash=abcd&path=`, so we should be careful. Is there a problem that prompted this?

In what scenario would that param combination be sent though? There hasn't ever been a scenario where we have advocated to use both of the params at the same time as far as I'm aware, and the alternative is that we just discard it - and then if the length of the object is 0 I guess throw the error?

There was a situation where sonarr/radarr was setting env vars incorrectly during the wrong import event that ended up empty, somewhat edge-case but given that this was the original intention that was later solved when I suggested usage of .partial to Q, i'm not sure what route you want to take.

mmgoodnow:
The problem doesn't really have anything to do with what we've recommended - it's that people have working setups that they don't expect a minor or patch update to break, because that's the contract semver establishes. 

Anyway, I retract my concern - I just saw we have a `.refine` check for `Object.keys(data).length === 1` which means this isn't a breaking change.

src/server.ts
```diff
@@ -37,7 +37,7 @@ const ANNOUNCE_SCHEMA = z
 const WEBHOOK_SCHEMA = z
 	.object({
 		infoHash: z.string().length(40),
-		path: z.string().refine((path) => !path || existsSync(path)),
+		path: z.string().refine((path) => path && existsSync(path)),
 	})
 	.strict()
 	.partial()
```
mmgoodnow, line 43: I think a union would be a truer representation of the actual data shape here and what we do with it. That would prevent the need to use Object.keys

https://zod.dev/?id=unions

PR #861, opened 2024-12-31
enhancement(action/linking): support multiple linkDirs

### LINKED ISSUE #625: feature request(hardlinking): Automagically hardlink to differing partition/disk/mount
Hardlinking seems like the most preferred linking solution, and indeed, https://github.com/cross-seed/cross-seed/pull/619 seeks to set hardlinking as the default.

Currently only one `linkDir:` is allowed to be set, resulting in what would appear to be non-optimal outcomes for those who have media files spread across different partition/disk/mount.

The most ideal solution, for us dumb users, would seem to allow multiple entries within `linkDir:`, where the user can set the directories for each partition/disk/mount, and xseed automagically uses the appropriate directory when linking.

ShanaryS:
fixes #625 

`linkDir` is converted to `linkDirs: [linkDir]`.

We choose a linkDir by testing if stat.dev if between source path and any linkDir is the same. Hardlink requires this to have a match but symlink will only display and warning and use `linkDirs[0]`.

The only possible issue is with `seasonFromEpisodes`, these could be sourced from multiple drives. The ones that are will be logged as failures before attempting to link.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,6 @@
 import chalk, { ChalkInstance } from "chalk";
 import { distance } from "fastest-levenshtein";
-import { readdirSync, statSync } from "fs";
+import { readdirSync } from "fs";
 import path from "path";
 import {
 	ALL_EXTENSIONS,
@@ -10,7 +10,6 @@ import {
 	EP_REGEX,
 	JSON_VALUES_REGEX,
 	LEVENSHTEIN_DIVISOR,
-	LinkType,
 	MediaType,
 	MOVIE_REGEX,
 	NON_UNICODE_ALPHANUM_REGEX,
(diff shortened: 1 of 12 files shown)


zakkarry:
The option name change alone is a breaking change....

ShanaryS:
> The option name change alone is a breaking change....

> `linkDir` is converted to `linkDirs: [linkDir]`.

zakkarry:
Yes, so what - you're just going to accept either type or option now? I can provide a single string or an array in linkDirs or an array in linkDir or any variation of this?

I don't have any of that nuance from "It's converted"

src/utils.ts
```diff
@@ -38,6 +40,38 @@ export function isTruthy<T>(value: T): value is Truthy<T> {
 	return Boolean(value);
 }
 
+export function getLinkDir(path: string): string | null {
```
mmgoodnow, line 43: I think this might fit better in action.ts

mmgoodnow (approved):
seems like it was pretty easy.

ShanaryS:
> Yes, so what - you're just going to accept either type or option now? I can provide a single string or an array in linkDirs or an array in linkDir or any variation of this?

It won't allow having both defined. linkDirs must be an array.

zakkarry:
Documentaiton needs updating.

Issue #223, opened 2022-04-11
Support multiple clients (local and remote)


ludviglundgren:
Hi,

I'm starting to plan and work on the autobrr cross-seed integration and some things came to mind. Some users (myself included) use multiple instances of say qBittorrent for various reasons and not always on the same system.

Being remote clients they wouldn't have access to the torrent state directory and only be able to use the data from the torrent list.

Is this something you could consider adding support for, or should users have a cross-seed app per server? The latter could probably make it harder with ratelimits against Prowlarr, while the first option to allow more clients would probably need quite a bit of changes to take into account which instance holds what.

mmgoodnow:
I would consider it in the long term; though I'd expect a feature like this to be 2+ years away. 

I think there would be significant complexity in the configuration for this. I urge you to propose what the configuration would look like for this feature. 

> only be able to use the data from the torrent list

what do you mean by this?

> The latter could probably make it harder with ratelimits against Prowlarr

It depends on how you group the torrents into clients. If they're grouped by tracker, then you can just set up one tracker for each instance of cross-seed.

Issue #48, opened 2021-03-05
Node out of memory


Drazzilb08:
https://paste.ubuntu.com/p/NrTC7rrJ8g/

So I was checking on things and noticed this in the logs as well as the container not running when it should have been.

If you require a larger paste let me know I'll snag that for you.

Note: running `:latest`

mmgoodnow:
When you run `cross-seed search` it will load all of the torrent files into memory. 

Since you're running it on ~7000 torrents the process is hitting its memory limit. If you assume an average torrent size of 100 kB, that means that all of the torrents loaded into memory will take roughly 700 MB, which, coincidentally, is the default max heap size of a Node process. 

I can make some optimizations to have the process use less memory but it might take a couple weeks. In the meantime, I think you can set the `NODE_OPTIONS` environment variable to something like `--max-old-space-size=2048` which will allocate up to 2GB of memory towards `cross-seed`, but never more than it needs. If you're in a memory-constrained environment you may not want to do this but it's an easy temporary fix.

Let's talk about your use case for a bit. I saw your `cross-seed-guide` repo and it looks like you're running `cross-seed search` with cron. That's fine but it was not really designed to be used like this - it loads all torrents into memory, and it hammers all of your trackers with search requests every run. If you run it daily it will search all trackers for all torrents every day. With 7000 torrents and 20 trackers that's 140,000 searches a day. (though it's not quite that bad if you use the --exclude-recent-search option)

What is your ideal solution here? Clearly you are a power user and I'd like to help out. I'd like to make daemon mode solve for your use-case.

Visorask:
I am one of the dev's for https://github.com/StuffAnThings/qbit_manage. 

We are basically using your outputs to inject into qbit. But at the moment your daemon mode only supports rtorrent. 

So basically the goal is to search any torrents as they come in and are completed downloading (daemon mode would be nice) and inject them into qbittorrent with the proper category, tag. Honestly a daemon mode that can detect when a torrent is done in qbit, search it and put it into a folder as is now would make the memory issue go away and cause less searches for trackers at once. 

My script could do the rest of the work just fine it's just the initial search being able to tell when a torrent is in qbit and finished downloading.

mmgoodnow:
What does your script do @Visorask?

mmgoodnow:
I'm definitely willing to add qbit integration if you can help me out a bit

Drazzilb08:
[...]

PR #864, opened 2025-01-02
feat(clients): option to use APIs instead of torrentDir

ShanaryS:
Fixes: #365 

Added `useClientTorrents` which replaces `torrentDir`. They cannot both be enabled.

There is now a new type of Searchee, `SearcheeClient`. This is like `SearcheeWithInfoHash`, but has a new field `savePath` which makes linking much easier.

This change means that we now support renames in client or any modification a user can do. Right before linking, we refresh the searchee to catch any renames as the db only updates on added/deleted. All searchees are also refreshed at the start of each search cadence.

Indexing torrents and data now happens asynchronously. There is also now prevention of multiple concurrent indexes to not overload the client/disk. If an index is in progress, announce/rss/webhoook will simply wait on its completion without triggering a new one.

I think we should store the client and data indexes in the physical db to prevent any chance of memory issues. However I think that should be saved for the PR supporting multiple clients, as that implementation will affect the schema. Considering that this feature is opt-in for existing users, memDB should be fine until then. On a similar note, the indexing functions can be refactored but should be saved for then.

### DIFF
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -196,39 +196,43 @@ export default class RTorrent implements TorrentClient {
 			| Fault[];
 
 		let response: ReturnType;
+		const args = [
+			[
+				{
+					methodName: "d.name",
+					params: [hash],
+				},
+				{
+					methodName: "d.directory",
+					params: [hash],
+				},
+				{
+					methodName: "d.left_bytes",
+					params: [hash],
+				},
+				{
+					methodName: "d.hashing",
+					params: [hash],
+				},
+				{
+					methodName: "d.complete",
+					params: [hash],
+				},
+				{
+					methodName: "d.is_multi_file",
+					params: [hash],
+				},
+				{
+					methodName: "d.is_active",
+					params: [hash],
+				},
+			],
+		];
 		try {
-			response = await this.methodCallP<ReturnType>("system.multicall", [
-				[
-					{
-						methodName: "d.name",
-						params: [hash],
-					},
-					{
-						methodName: "d.directory",
-						params: [hash],
-					},
-					{
-						methodName: "d.left_bytes",
-						params: [hash],
-					},
-					{
-						methodName: "d.hashing",
-						params: [hash],
-					},
-					{
-						methodName: "d.complete",
-						params: [hash],
-					},
-					{
-						methodName: "d.is_multi_file",
-						params: [hash],
-					},
-					{
-						methodName: "d.is_active",
-						params: [hash],
-					},
-				],
-			]);
+			response = await this.methodCallP<ReturnType>(
+				"system.multicall",
+				args,
+			);
 		} catch (e) {
 			logger.debug(e);
 			return resultOfErr("FAILURE");
@@ -354,25 +358,29 @@ export default class RTorrent implements TorrentClient {
 		);
 		type ReturnType = string[][] | Fault[];
 		let response: ReturnType;
+		const args = [
+			infoHashes
+				.map((hash) => [
+					{
+						methodName: "d.directory",
+						params: [hash],
+					},
+					{
+						methodName: "d.is_multi_file",
+						params: [hash],
+					},
+					{
+						methodName: "d.complete",
+						params: [hash],
+					},
+				])
+				.flat(),
+		];
 		try {
-			response = await this.methodCallP<ReturnType>("system.multicall", [
-				infoHashes
-					.map((hash) => [
-						{
-							methodName: "d.directory",
-							params: [hash],
-						},
-						{
-							methodName: "d.is_multi_file",
-							params: [hash],
-						},
-						{
-							methodName: "d.complete",
-							params: [hash],
-						},
-					])
-					.flat(),
-			]);
+			response = await this.methodCallP<ReturnType>(
+				"system.multicall",
+				args,
+			);
 		} catch (e) {
 			logger.error({
 				Label: Label.RTORRENT,
@@ -442,15 +450,19 @@ export default class RTorrent implements TorrentClient {
 		);
 		type ReturnType = string[][] | Fault[];
 		let response: ReturnType;
+		const args = [
+			infoHashes.map((hash) => {
+				return {
+					methodName: "d.custom1",
+					params: [hash],
+				};
+			}),
+		];
 		try {
-			response = await this.methodCallP<ReturnType>("system.multicall", [
-				infoHashes.map((hash) => {
-					return {
-						methodName: "d.custom1",
-						params: [hash],
-					};
-				}),
-			]);
+			response = await this.methodCallP<ReturnType>(
+				"system.multicall",
+				args,
+			);
 		} catch (e) {
 			logger.error({
 				Label: Label.RTORRENT,
@@ -511,53 +523,44 @@ export default class RTorrent implements TorrentClient {
 				);
 				type ReturnType = any[][] | Fault[]; // eslint-disable-line @typescript-eslint/no-explicit-any
 				let response: ReturnType;
+				const args = [
+					hashes
+						.map((hash) => [
+							{
+								methodName: "d.name",
+								params: [hash],
+							},
+							{
+								methodName: "d.size_bytes",
+								params: [hash],
+							},
+							{
+								methodName: "d.directory",
+								params: [hash],
+							},
+							{
+								methodName: "d.is_multi_file",
+								params: [hash],
+							},
+							{
+								methodName: "d.custom1",
+								params: [hash],
+							},
+							{
+								methodName: "f.multicall",
+								params: [hash, "", "f.path=", "f.size_bytes="],
+							},
+							{
+								methodName: "t.multicall",
+								params: [hash, "", "t.url=", "t.group="],
+							},
+						])
+						.flat(),
+				];
 				try {
 					response = await this.methodCallP<ReturnType>(
 						"system.multicall",
-						[
-							hashes
-								.map((hash) => [
-									{
-										methodName: "d.name",
-										params: [hash],
-									},
-									{
-										methodName: "d.size_bytes",
-										params: [hash],
-									},
-									{
-										methodName: "d.directory",
-										params: [hash],
-									},
-									{
-										methodName: "d.is_multi_file",
-										params: [hash],
-									},
-									{
-										methodName: "d.custom1",
-										params: [hash],
-									},
-									{
-										methodName: "f.multicall",
-										params: [
-											hash,
-											"",
-											"f.path=",
-											"f.size_bytes=",
-										],
-									},
-									{
-										methodName: "t.multicall",
-										params: [
-											hash,
-											"",
-											"t.url=",
-											"t.group=",
-										],
-									},
-								])
-								.flat(),
-						],
+						args,
 					);
 				} catch (e) {
 					logger.error({
--- a/src/clients/RTorrent.ts
+++ b/src/clients/RTorrent.ts
@@ -27,10 +27,8 @@ import {
 	extractCredentialsFromUrl,
 	humanReadableSize,
 	isTruthy,
-	Mutex,
 	sanitizeInfoHash,
 	wait,
-	withMutex,
 } from "../utils.js";
 import {
 	ClientSearcheeResult,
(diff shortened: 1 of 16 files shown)


zakkarry:
Had we decided whether or not we were going to completely rely on API for clients? For deluge this could be problematic, and I think we'd need to look at some code I've had sitting locally for a while in regards to the desyncs that can potentially occur.

ShanaryS:
> Removing the potential for me to opt to use torrentDir if I prefer to do so seems like a flawed decision

We are relying on each torrent clients implementation of how they store data which no application should do. We are deeply coupled to how their internals operate and whatever changes they may do (and have already done!). Unless we have a clear reason not to (such as the deluge issues if it cannot be addressed), we should use the documented, maintained, and the intended method for querying the client state.

torrentDir adds an entirely new class of issues that cross-seed has to deal with. It offers no inherent benefits over the APIs, aside from indexing possibly being slower. Supporting both adds a great deal of unnecessary complexity everywhere, from validation to linking, requiring a significant amount of workarounds. It is responsible for a signification portion of support tickets and first time user confusion. I cannot imagine if the arrs, autobrr, or any other application requiring access to another's internal directory, instead of solely relying on their APIs.

But perhaps most importantly, it's an untenable solution long term for addressing #365, #223, and more. Using the APIs are necessary now (qBittorrent sqlite) and for future development of cross-seed (supporting multiple clients and remote clients). It directly supersedes torrentDir and so it should be the only option long term (again, assuming the deluge API issues can be addressed).

zakkarry:
No, **_your_** code is deeply coupled. 

Deluge for instance only reads the plain torrent files. The spec is documented and consistent with the original .torrent, and there are no issues with relying on it. You have all these fastresume files, inserted keys that are out of spec, SQLite support to mask bad performance, and altered/mutated .torrent files that you must worry about. No one else has to worry about any of that nonsense, and you're making your client's problem's solution a burden on everyone else by not giving them the option to keep doing what has worked for years.

There is nothing that prevents the implementation that we've had for years from continuing to function perfectly for clients other than qBit. I keep telling you that, and I suggested that you can default to using the API as a compromise, but if I want to set your little `useClientTorrents` to false and use a torrentDir - you have provided me no valid reason other than "cuz I want it my way cuz that's how i have to do it" as justification. You're not the only user and you're use case and client is not the only ones either. Claiming that there's some major flaw that doesn't present any problems for years for other clients outside of Qbit with your new features is just bullshit.

If I query the API to index my 25k torrents, it's noticeably slower than indexing them off my SSD and file system. You're forcing me to use API and not letting me - who has a large amount of torrent while you sit with substantially less pretend you know what I'm talking about - you can't even test what you're proposing. 

You claimed you tested Deluge, but you didn't even try to test Deluge properly. Adding torrents in a paused state, not downloading anything on them, and querying an empty set of data that hasn't even generated any state or fastresume data is not a valid way to see if it's "the same". You're proposing changes, and having used the Deluge API extensively myself, I know that it is slower to pull what we would need for searchee generation... It's a single-threaded HTTP 1 connection. It's parsing a pickled .state file, formatting it, and "streaming" it asynchronously over a single socket. How can you possibly think that's the same as reading it from the file system as we do now?

I don't want this change because I don't need it and it's slower, that should be enough. So let me opt-in to use torrentDir if I want. What do you care if I want to use torrentDir? Have it default to use the API over it, but if I want to do what works best for me who are you to dictate? You haven't even tested anything - 1K torrents is not a benchmark to even consider, because that's not even a level of torrents where limitations or pitfalls would arise and you know it.

There is no "deep coupling" for anyone else. You're projecting, and we've done it this way for years, long before you had probably even heard of cross-seed, probably before you even started torrenting most certainly before you were using PTs. You're coming in and shouting "the world is on fire" over what YOU'VE been developing for the last six months being a problem now for YOU, and then claiming that it somehow is problematic for the other clients and needs intervention to the point of removing my ability to have the same performance, the performance I've come to expect.

The entire premise is ridiculous, and I told you I didn't want to keep talking about this with you yesterday, that I was not enjoying your obstinance and belligerence around this proposal, and here you are the next day because you think this exchange you've been dragging me back into multiple times is "good" (your words) despite you adamantly refusing to compromise or bend on any aspect of this PR - just covering your ears and shouting "lalalala" every time I tell you why my stance is what it is. The entire time I've asked you for your rationale you provide me nothing but abstract "change for the better" cliche when in the same breath you tell me "better" is "the same".

**I don't want to discuss this with you any further until you can learn to listen, work together, and potentially make a compromise that benefits (or at least works for) everyone and not just you and your arbitrary desire to control people.**

mmgoodnow:
I like the direction this PR heads.

`torrentDir` has been around since pretty much the beginning. At the time, I decided to use it because I was racing to an MVP, and it was one less API client to write—the MVP already _had_ to deal with snatched torrent files, so why not use them to compare against too? That would make the comparison much easier too—just compare two torrent files to each other, and I could write the rTorrent code later. It was also nice because it was client-agnostic—just give `cross-seed` a bunch of torrent files, everyone has those. This expanded my total addressable market to all PT users (rather than just rTorrent users), and the value proposition of `cross-seed` at the time didn't really have anything to do with your client—`autotorrent` already did that part.

> Aside: at the time there was no `Searchee` abstraction - you would load the contents of every torrent file into memory, including all the piece hashes, and cross-seed would just sit on all of those `Metafile`s until it exited, which could take several days. Predictably, https://github.com/cross-seed/cross-seed/issues/48 came shortly after.

`torrentDir` isn't the problem. The problem is using client session directories as torrentDir, while we simultaneously increase our level of client integration. Client session directories are **private, internal APIs** and depending on them carries a maintenance cost. We saw this with Jackett - the MVP of cross-seed used Jackett's private API, and then when Prowlarr rolled around the block, we weren't ready for it and had to make a breaking change. We've seen this with qBittorrent—users have been forced to hold back from the much more performant SQLite backend for 2 years because of cross-seed's reliance on their session dir. glassez confirmed that BT_backup is not for external use in [this comment](https://github.com/qbittorrent/qBittorrent/issues/1606#issuecomment-2395173676). We have gotten away with it on the other clients because they are not very actively developed, but I wouldn't be too surprised to see _all_ the relevant clients using a SQLite storage layer ten years from now. 

So, what do we do?

We should switch to pulling searchees from client APIs as the default behavior. It's less configuration and easier for new users. Anyone pointing torrentDir to their session directory and also using injection or category filtering should switch to using their client's API. We should make people try it out—I don't think we'll have problems with performance if we persist the cache (disclaimer: I haven't read the PR yet). For nearly all users (everyone with injection) this change should be invisible. If it's not, we should work to make it invisible.

I don't think we should remove torrentDir - its benefit is that it's dependency free. We should still let people use the matching algorithm, either for an unsupported torrent client or just hacking stuff together for fun. We could restrict it in a few ways and hopefully still delete most of the complexity - 
1. don't support injection
2. don't support linking
3. don't support advanced match modes
4. don't support advanced prefiltering
 
Of these, 1 and 4 seem to make sense to me, **eventually**, when it's time to make another major version release. 

As for Deluge, if its API breaks catastrophically once a day, it's not on cross-seed to fix a pretty egregious bug. We can and should look into workarounds, and should make sure we're not breaking Deluge users' setups, but ultimately this kind of showstopping bug just harms Deluge's credibility as a torrent client.

zakkarry:
Also just as a side note, I do not have to rely on state for Deluge.  I can use an EXTERNAL and not private directory, simply by enabling a checkbox.

image

I also worked with glassez on GHI to push for the same feature to be added to qBit. This makes the entire point moot, because its external and meant for third parties to use - which for Deluge it always was an invalid argument that I just didn't pursue because every other argument I made prior to it was ignored, dismissed, and antagonized for.

zakkarry:
> As for Deluge, if its API breaks catastrophically once a day, it's not on cross-seed to fix a pretty egregious bug. We can and should look into workarounds, and should make sure we're not breaking Deluge users' setups, but ultimately this kind of showstopping bug just harms Deluge's credibility as a torrent client.

The bug is only present in docker implementations because of the way every single maintainer of docker images does this, I tried to get them to fix it and no one was interested in the manner I suggested.

It has nothing to do with Deluge itself.

I've discussed this with you previously, and there's an issue on this very repository outlining some of what I did and found somewhere.

You and Q can talk as much shit as you want about Deluge, but neither of you know what you're talking about.

mmgoodnow:
The diff is pretty intimidating, so I'm going to be aggressive about asking you to extract stuff into separate PRs that we can merge first.

src/db.ts
```diff
@@ -22,12 +22,62 @@ export const memDB = Knex.knex({
...
+	table: string,
+	column: string,
+	values: any[], // eslint-disable-line @typescript-eslint/no-explicit-any
+): Promise<void> {
+	for (let i = 0; i < values.length; i += BATCH_SIZE_DB) {
+		const batch = values.slice(i, i + BATCH_SIZE_DB);
+		if (!batch.length) break;
+		await database(table).whereIn(column, batch).del();
+	}
+}
```
mmgoodnow, line 83: I think these helpers would benefit from taking a callback:

```ts
function inBatches<T>({ rows: T[], batchSize = 1000 }, cb: (batch: T[]) => Promise<void>) {
	for (let i = 0; i < rows.length; i += batchSize) {
 		const batch = rows.slice(i, i + batchSize);
 		if (!batch.length) break;
 		await cb(batch);
 	}
}

inBatches({ rows }, (batch) => {
	await memDB(table).insert(batch).onConflict('foo').ignore();
});
```

Returning the control to the consumer this way means we don't have to take a bunch of passthrough arguments like `database`, `table`, or `conflictKey`, and we don't need to write separate functions for `.ignore(), .merge()`, or deleting instead of inserting.

This can even be generalized to non-database operations - like doing torznab searches in batches or something. The only thing we'd have to change is the name of the `rows` argument to something more generic. 

I think this is a good candidate for extracting into a separate PR upstream of this one, along with whatever new consumptions can be added to existing code.

src/cmd.ts
mmgoodnow, line 1: The `useClientTorrents` option looks good to me. Let's extract the boring boilerplate parts of scaffolding the new option into an upstream PR.

src/startup.ts
mmgoodnow, line 1: please extract this as well (if you care about it). it doesn't seem super consequential to me, but happy to approve it separately

src/server.ts
mmgoodnow, line 1: and these changes

src/searchee.ts
```diff
@@ -79,6 +81,10 @@ export interface Searchee {
 
 export type SearcheeWithInfoHash = WithRequired<Searchee, "infoHash">;
 export type SearcheeWithoutInfoHash = WithUndefined<Searchee, "infoHash">;
+export type SearcheeClient = WithRequired<
+	Searchee,
+	"infoHash" | "savePath" | "category" | "tags" | "trackers"
+>;
 export type SearcheeVirtual = WithUndefined<Searchee, "infoHash" | "path">;
 export type SearcheeWithLabel = WithRequired<Searchee, "label">;
```
mmgoodnow, line 89: we have a lot of these. should we make it a union type?

src/clients/TorrentClient.ts
```diff
@@ -31,19 +31,38 @@ export type TorrentClientType =
 	| Label.TRANSMISSION
 	| Label.DELUGE;
 
+export type Tracker =
+	| { url: string; announce?: string; tier: number }
+	| { url?: string; announce: string; tier: number };
```
mmgoodnow, line 36: are `announce` and `url` not the same?

src/clients/TorrentClient.ts
```diff
@@ -31,19 +31,38 @@ export type TorrentClientType =
...
 	getAllTorrents: () => Promise<TorrentMetadataInClient[]>;
+	/**
+	 * Get all searchees from the client and update the db
+	 * @param options.newSearcheesOnly only return searchees that are not in the db
+	 * @param options.refresh true to refresh all searchees, or an array of infoHashes to refresh
+	 * @return an object containing all searchees and new searchees (refreshed searchees are considered new)
+	 */
+	getClientSearchees: (options?: {
+		newSearcheesOnly?: boolean;
+		refresh?: boolean | string[];
```
mmgoodnow, line 64: this `boolean | string[]` is kind of weird. can you find a cleaner API?

src/clients/TorrentClient.ts
ShanaryS: Transmission returned `announce` in their API but now I'll just convert it to `url`.

src/searchee.ts
ShanaryS: I'm not sure how that works. Is this to combine them all into one type or change the method we use to differentiate.

src/clients/TorrentClient.ts
ShanaryS: Now changed to `@param options.refresh undefined uses the cache, [] refreshes all searchees, or a list of infoHashes to refresh`

`undefined` - most uses of this function
`[]` - used at the start of a search to update any changes since this can happen at most of 1d. Users are still expected to run `cross-seed --clear-client-cache` if they rename or modify trackers in client (which will use the arg `[]`). Doing this at the start of a search is just a precaution against users who don't (should be rarely needed)
`[infoHash]` - used during action right before linking the source searchee to catch any renames. The refresh before search is mainly for updating tracker blocklists or maybe getting a better title to parse in case of a name change.

src/searchee.ts
mmgoodnow: The benefit of the union type is that TS can enforce that any searchee has at least one of `infoHash`/`path` specified, and it automatically narrows types for you. 

```
if ("infoHash" in searchee) {

} else {
// here, TS knows that `searchee.path` exists
}
```

Just a minor quality of life thing.

src/clients/RTorrent.ts
```diff
@@ -478,6 +496,165 @@ export default class RTorrent implements TorrentClient {
 		}
 	}
 
+	async getClientSearchees(options?: {
+		newSearcheesOnly?: boolean;
+		refresh?: string[];
+	}): Promise<ClientSearcheeResult> {
+		return withMutex(
+			Mutex.QUERY_CLIENT,
```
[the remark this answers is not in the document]
ShanaryS: The only time this could happen is if search/webhook is triggered at the same time as announce/rss/webhook. We may as well do it here since the function already exists.

src/clients/RTorrent.ts
```diff
@@ -478,6 +496,165 @@ export default class RTorrent implements TorrentClient {
...
+			async () => {
+				const searchees: SearcheeClient[] = [];
+				const newSearchees: SearcheeClient[] = [];
+				const hashes = await this.methodCallP<string[]>(
+					"download_list",
+					[],
+				);
+				type ReturnType = any[][] | Fault[]; // eslint-disable-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
ShanaryS: I trying using unknown but I had to explicitly `as type` each item that it didn't seem any safer and made the logic more complicated.

src/clients/RTorrent.ts
```diff
@@ -478,6 +496,165 @@ export default class RTorrent implements TorrentClient {
...
+				let response: ReturnType;
+				try {
+					response = await this.methodCallP<ReturnType>(
+						"system.multicall",
+						[
+							hashes
+								.map((hash) => [
+									{
```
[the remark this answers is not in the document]
ShanaryS: I'm not too sure for rTorrent but others clients should not be an issue. I tried many many different ways to query rTorrent and this method was the only one that worked.

Doing it in batches could be fine if we have the results ordered by added time. Then we can only check the recent additions and only delete from db on search or reverse lookup success.

src/clients/RTorrent.ts
```diff
@@ -478,6 +496,165 @@ export default class RTorrent implements TorrentClient {
...
+							? false
+							: options.refresh.length === 0
+								? true
+								: options.refresh.includes(infoHash);
+					if (dbTorrent && !refresh) {
+						if (!options?.newSearcheesOnly) {
+							searchees.push(createSearcheeFromDB(dbTorrent));
+						}
+						continue;
+					}
```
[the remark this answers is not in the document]
ShanaryS: The loop over the torrents can't be meaningfully abstracted since each client API results are different. In the case of qBittorrent, we need separate API calls for the files and trackers so we can't get all the data to pass onto a sperate function.

src/searchee.ts
ShanaryS: I just tested it and it seems like a nice change, but will require a large refactor to take advantage of. I'll save it for a later PR.

mmgoodnow:
Oops, I didn't realize those PRs were branched off this one.

src/clients/RTorrent.ts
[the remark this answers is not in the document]
mmgoodnow: I'm not sure what you mean by 

> so we can't get all the data first to then pass onto a sperate function.

src/clients/RTorrent.ts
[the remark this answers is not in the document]
mmgoodnow: I'm just not really sure I see the benefit since they're concurrency safe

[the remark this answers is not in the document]
ShanaryS: It was to prevent the client having to process the requests at the same time. But they might even have some caching internally. I've removed it.

mmgoodnow (dismissed):
We have a couple unresolved threads here, but I don't feel super strongly about any of them and the solutions to them all are  user-invisible. I think we can merge to keep things moving, nothing's set in stone. @ShanaryS I'll leave the merge to you, in case you want to make a couple more commits.

src/clients/RTorrent.ts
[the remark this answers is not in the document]
ShanaryS: With qbit we can't get all the data then send it to a boilerplate function that creates the searchees. Because we need to make additional api calls if the torrent is not it the db or based on the function arguments. Even outside of that, the parts that can be abstracted are about 7 lines. Almost all of the code is converting the client API results into a searchee.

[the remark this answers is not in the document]
mmgoodnow: Maybe a modification to the API would help then—taking `knownSearchees` as a parameter, or `knownInfoHashes`.

zakkarry:
Set it to null.

ShanaryS:
Set `torrentDir: null`

zakkarry:
@ShanaryS maybe push to the docs and add this to the config template, although users modifying existing configs will still not see it without reading.

mmgoodnow:
"Fixed" in https://github.com/cross-seed/cross-seed/commit/9b3d405, released in v6.9.1

d4rkfella:
Thanks for the quick update :)

PR #867, opened 2025-01-03
fix(db): prevent cross-seed hanging on exit

ShanaryS:
Knex requires `.destroy()` to be called if the database was used. The change discussed [here](https://github.com/cross-seed/cross-seed/pull/857#discussion_r1899281648) means memDB always has the schema created on startup, and thus needs to be destroyed for all `cross-seed cmd`.

`destroy()` seems to have no effects so calling both everywhere is best.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -325,7 +325,11 @@ program
 		"-d, --docker",
 		"Generate the docker config instead of the normal one",
 	)
-	.action(() => void generateConfig());
+	.action(async () => {
+		void generateConfig();
+		await db.destroy();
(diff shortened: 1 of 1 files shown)


zakkarry:
lol seems a bit insane calling it for absolutely every exit - not sure it's needed _EVERYWHERE_, right?

but i'd rather be safe than sorry. can't hurt.

zakkarry (approved):

ShanaryS:
`memDB.destroy()` is necessary everywhere, `db.destroy()` is not but it causes no issues.

PR #872, opened 2025-01-06
Build platform-native binaries

mmgoodnow:
https://notes.billmill.org/programming/javascript/Making_a_single-file_executable_with_node_and_esbuild.html

Required changes:
- [x] stop using better-sqlite3 (#858)
- [x] stop using top level await
- [x] conditionally retrieve the config template with `sea.getAsset()` API
- [x] import `config.js` using createRequire

the rough build process is:
- [x] bundle everything together into a single CommonJS file with esbuild or something else
- [x] run `node --experimental-sea-config sea-config.json` to create an injectable blob
- [x] use `postject` to inject the blob (and the config template as a separate asset) into a copy of the `node` binary
- [ ] code sign the new binary


zakkarry:
is the end goal to be providing a single js file, or an actual binary (i.e. .exe on windows) that we can distribute re: single executable

mmgoodnow:
The latter

PR #875, opened 2025-01-06
fix(decide/cache): don't update db if snatch already exists

ShanaryS:
The goal is to prevent matching guid to a cached torrent from another tracker if the infoHash is the same. This can cause a torrent to be injected with the wrong save path with `flatLinking: false` and incorrect tracker logging. The torrents from these two trackers would not be able to coexist in client regardless so first come first served is fine.

This prevents caching the guid in the db if we try to snatch but the infoHash already exists in the cache. It already won't overwrite the existing file and only update the last access time. Now it also won't cache this result in the database. This should only happen in these scenarios:
- This bug
- If the user has deleted the config.db without deleting the torrent_cache
- Any more trackers that benefit from fuzzy lookups that we haven't covered

This does not affect the guid (or fuzzy guid) lookups that we already cover. It's only when that check fails and we try to snatch the candidate.

The 2nd commit is to clean up old cached entries after a year past the last access time which was previously rejected when the caching was implemented. However it would now benefit the first two use cases above aside from just cleaning up space. It also serves to correct this issue for current users over time.

Since the access time is updated each time we check if it exists (so using the cache or trying to snatch), the timing of cleanup is a year after the torrent is unregistered or a year after `excludeOlder`, whichever comes first. It's currently being done before caching a torrent file as timing could matter.

We could also include the guid along with the infoHash when saving. But it doesn't really seem worth changing it since only one can exist in client and having to deal with existing user migrations.

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -1,5 +1,6 @@
 import {
 	existsSync,
+	readdirSync,
 	statSync,
 	unlinkSync,
 	utimesSync,
@@ -42,6 +43,7 @@ import {
 	getLogString,
 	getMediaType,
 	getMinSizeRatio,
+	inBatches,
 	sanitizeInfoHash,
 	stripExtension,
 	wait,
@@ -360,7 +362,7 @@ export async function assessCandidate(
 			}
 		}
 		metafile = res.unwrap();
-		metaCached = cacheTorrentFile(metafile);
+		metaCached = await cacheTorrentFile(metafile);
 		metaOrCandidate.size = metafile.length; // Trackers can be wrong
 	} else {
 		metafile = metaOrCandidate;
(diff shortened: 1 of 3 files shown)


zakkarry:
Just to get an idea of what you're thinking, are we always assuming the original is still in the client and seeding (do we check that or just the existence in the cache) and that the original is still relevant in some capacity?

I'm curious if it would be more advantageous to invalidate the previous after checking its existence in the client.

ShanaryS:
I don't think there is a need to checking if it's in client. Even if it wasn't, the cache might still be valuable in the future.

This new tracker will still get injected (if not already exists), just not cached for future use. I see the value in removing the previous tracker and caching this one if it's not in client but then it would just rotate which torrent is actually cached.

zakkarry:
> I don't think there is a need to checking if it's in client. Even if it wasn't, the cache might still be valuable in the future.
> 
> This new tracker will still get injected (if not already exists), just not cached for future use. I see the value in removing the previous tracker and caching this one if it's not in client but then it would just rotate which torrent is actually cached.

Ya, I think the scenario I was bringing this up for was basically that the torrent got removed - think episodes for a season pack, for example - and that the previously cached one would never be useful again.

I could see this being somewhat difficult to determine, and the resulting and subsequent snatches when necessary due to the lack of caching on one of them not necessarily being that critical anyway.

Just wanted to bring up the premise.

src/decide.ts
```diff
@@ -446,14 +460,30 @@ async function getCachedTorrentFile(
...
+	const now = Date.now();
+	const entriesToDelete: string[] = [];
+	for (const file of files) {
+		const filePath = path.join(torrentCacheDir, file);
+		if (now - statSync(filePath).atimeMs > ms("1 year")) {
+			logger.verbose(`Deleting ${file}`);
+			entriesToDelete.push(file.split(".")[0]);
+			unlinkSync(filePath);
+		}
+	}
```
mmgoodnow, line 475: seems a little odd to me to do this here, do you prefer doing it here over a job?

ShanaryS: This allows the latest check right before trying to save a potential conflict. But considering the timescales a job may be fine.

ShanaryS: From a performance perspective a job is probably better.

mmgoodnow (approved):

PR #878, opened 2025-01-11
fix(action/linking): trim tracker to match qbit behavior

ShanaryS:
If a user has trailing white space in the Prowlarr indexer names, `flatLinking: false` will preserve this when linking but qBittorrent trims the savePath when being injected.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -25,10 +25,16 @@ import { getRuntimeConfig } from "./runtimeConfig.js";
 
 const ANNOUNCE_SCHEMA = z
 	.object({
-		name: z.string().refine((name) => name.trim().length > 0),
+		name: z
+			.string()
+			.transform((name) => name.trim())
+			.refine((name) => name.length > 0),
 		guid: z.string().url(),
 		link: z.string().url(),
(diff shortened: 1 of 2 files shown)


zakkarry:
I think the real issue stems from the fact that you can't have trailing whitespace at the end of a directory name on windows for example, not that we should do this to match qbit behavior...

ShanaryS:
I'm not sure how Windows not allowing that factors in here. The trim has no downsides either way.

zakkarry:
> I'm not sure how Windows not allowing that factors in here. The trim has no downsides either way.

I mean, that was the reason I thought we were going to do this...to maintain the behavior across OS's, I guess it's just semantics, but the limiting factor being the OS seems to me to supersede the claim that this is qBittorrent behavior.

mmgoodnow:
could this break something?

ShanaryS:
The tracker names are only used for logging and linking.

If a user has whitespace only in prowlarr it may fail to link but that's extreme. Users with other clients might have multiple tracker directories if they had a trailing space.

mmgoodnow (approved):

PR #882, opened 2025-01-17
enhancement(decide): check release version

ShanaryS:
I've readded the repack check which was removed in #707 due to decision cache. Since we are no longer caching decisions without snatches, this no longer applies.

Previously, this would only fail if the user was in safe mode, it can now fail in all modes. It's also no longer checking the `ARR_PROPER_REGEX` as this is a subset of the main regex. The searchee and candidate must now share the same version or both be missing.

This will increase the false negatives as some places don't correctly tag repack status, but a few repacks have been seen with the same size. Since repacks are rare, this should be a fine tradeoff for safety.


zakkarry:
I had thought about proposing this to you during the triage we were doing, but I figured you'd add it back in one of the PRs you made if it was good to go.

mmgoodnow (approved):

PR #886, opened 2025-01-19
enhancement(webhook): ability to override certain config options

### LINKED ISSUE #869: Add webhook bypass for includeSingleEpisodes
### Does your feature request solve a problem you're facing?

Currently it is recommended to set `includeSingleEpisodes` to false.  However, this then means that trying to search single episodes [as recommended for Usenet downloads](https://www.cross-seed.org/docs/tutorials/triggering-searches); they are never searched.

```
2024-12-06 21:38:25 info: [webhook] Received search request: {
  path: '/mnt/data/media/tv/Fire Country (2022) {imdb-tt16098700}/Season 03/Fire Country (2022) - S03E07 - False Alarm [HDTV-720p][AC3 5.1][x264]-SYNCOPY.mkv'
}
2024-12-06 21:38:25 verbose: [prefilter] /mnt/data/media/tv/Fire Country (2022) {imdb-tt16098700}/Season 03/Fire Country (2022) - S03E07 - False Alarm [HDTV-720p][AC3 5.1][x264]-SYNCOPY.mkv | MediaType: EPISODE - was not selected for searching because it is a season pack episode
2024-12-06 21:38:25 info: [webhook] Did not search for {
  path: '/mnt/data/media/tv/Fire Country (2022) {imdb-tt16098700}/Season 03/Fire Country (2022) - S03E07 - False Alarm [HDTV-720p][AC3 5.1][x264]-SYNCOPY.mkv'
} (check verbose logs for preFilter reason)
```
Makes the xseed data import script useless

### Does an existing issue/feature request already exist?

### What would you like for cross-seed to do?

Bypass or allow a flag/attribute for includeSingleEpisodes to be bypassed webhook data searches.

this doesn't seem to be a sane reason why this should apply to webhook (I.e. ad-hoc) searches

### What are you currently doing, if anything, to solve or work around the issue?

Enabling `includeSingleEpisodes` accepting the annoying warning and accepting possibly unecessary seeding and extra torrents.

### Additional comments

https://discord.com/channels/880949701845872672/1314800254926852097

ShanaryS:
Fixes #869

This `getRuntimeConfig()` now takes `configOverride: Partial<RuntimeConfig> = {}` to override config values for that call. This is used by these webhook params.

The primary use for `webhook` is to immediately trigger a search on download completion. Any other automation should be handled by `searchCadence`. As such, we should default to config options but allow overwriting a few notable ones.

The docs should also be updated with overriding `includeSingleEpisodes` being a valid alternative for download completion as they are unlikely to have been trumped by a season pack.

`curl -XPOST <BASE_URL>/api/webhook?apikey=<API_KEY> -d "infoHash=%I" -d "includeSingleEpisodes=true"`

Additionally, I think `webhook` should ignore torrents with the `cross-seed` tag by default, in service of the primary use. Currently, lots of wasted searches are triggered from partial matches completing which we now have the opportunity to rectify.

Respecting excludes and ignoring the `cross-seed` tag is slightly breaking but doesn't affect the primary use. The main concern is any current automation being used as a replacement for `searchCadence` which of course is not recommended anyways. Users running the command manually in one off situations should receive instant feedback with the change.

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -29,7 +29,7 @@ import {
 	isSingleEpisode,
 } from "./preFilter.js";
 import { Result, resultOf, resultOfErr } from "./Result.js";
-import { getRuntimeConfig } from "./runtimeConfig.js";
+import { getRuntimeConfig, RuntimeConfig } from "./runtimeConfig.js";
 import {
 	File,
 	getReleaseGroup,
@@ -60,8 +60,9 @@ function logDecision(
 	decision: Decision,
 	metafile: Metafile | undefined,
 	tracker: string,
+	options?: { configOverride: Partial<RuntimeConfig> },
 ): void {
-	const { blockList, matchMode } = getRuntimeConfig();
+	const { blockList, matchMode } = getRuntimeConfig(options?.configOverride);
 
 	let reason: string;
 	switch (decision) {
@@ -315,8 +316,11 @@ export async function assessCandidate(
 	metaOrCandidate: Metafile | Candidate,
 	searchee: SearcheeWithLabel,
 	infoHashesToExclude: Set<string>,
+	options?: { configOverride: Partial<RuntimeConfig> },
 ): Promise<ResultAssessment> {
-	const { blockList, includeSingleEpisodes, matchMode } = getRuntimeConfig();
+	const { blockList, includeSingleEpisodes, matchMode } = getRuntimeConfig(
+		options?.configOverride,
+	);
 
 	// When metaOrCandidate is a Metafile, skip straight to the
 	// main matching algorithms as we don't need pre-download filtering.
@@ -506,11 +510,13 @@ async function assessAndSaveResults(
 	infoHashesToExclude: Set<string>,
 	firstSeen: number,
 	guidInfoHashMap: Map<string, string>,
+	options?: { configOverride: Partial<RuntimeConfig> },
 ) {
 	const assessment = await assessCandidate(
 		metaOrCandidate,
 		searchee,
 		infoHashesToExclude,
+		options,
 	);
 
 	if (assessment.metaCached && assessment.metafile) {
@@ -581,6 +587,7 @@ export async function assessCandidateCaching(
 	searchee: SearcheeWithLabel,
 	infoHashesToExclude: Set<string>,
 	guidInfoHashMap: Map<string, string>,
+	options?: { configOverride: Partial<RuntimeConfig> },
 ): Promise<ResultAssessment> {
 	const { name, guid, link, tracker } = candidate;
 
@@ -629,6 +636,7 @@ export async function assessCandidateCaching(
 			infoHashesToExclude,
 			cacheEntry?.firstSeen ?? Date.now(),
 			guidInfoHashMap,
+			options,
 		);
 	}
 
@@ -638,6 +646,7 @@ export async function assessCandidateCaching(
 		assessment.decision,
 		assessment.metafile,
 		tracker,
+		options,
 	);
 
 	return assessment;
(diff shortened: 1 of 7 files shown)


src/decide.ts
```diff
@@ -60,8 +60,9 @@ function logDecision(
 	decision: Decision,
 	metafile: Metafile | undefined,
 	tracker: string,
+	options?: { configOverride: Partial<RuntimeConfig> },
```
mmgoodnow, line 63: if we're going to add `options` to all the arguments through the tree, I think it'd be better to not call `getRuntimeConfig` at all - it'll pave the way for restartless config updates. It'd be nice to get to a point where we just call `getRuntimeConfig` once on each entry point (a job starting, an api call). This might be a convenient halfway point though.

src/runtimeConfig.ts
```diff
@@ -49,6 +49,13 @@ export function setRuntimeConfig(configObj: RuntimeConfig): void {
...
 
-export function getRuntimeConfig(): RuntimeConfig {
-	return runtimeConfig;
+export function getRuntimeConfig(
+	configOverride: Partial<RuntimeConfig> = {},
+): RuntimeConfig {
+	return {
+		...runtimeConfig,
+		...Object.fromEntries(
+			Object.entries(configOverride).filter(([, v]) => v !== undefined),
```
mmgoodnow, line 58: would we expect `configOverride` to have explicitly set `undefined` values?

src/pipeline.ts
```diff
@@ -285,7 +294,13 @@ export async function searchForLocalTorrentByCriteria(
 	for (const [i, searchee] of searchees.entries()) {
 		const progress = chalk.blue(`(${i + 1}/${searchees.length}) `);
 		try {
-			if (!filterByContent(searchee, includeEpisodes)) {
+			if (
+				!filterByContent(searchee, {
+					configOverride: options.configOverride,
+					includeEpisodes,
```
mmgoodnow, line 300: why is includeEpisodes passed separately?

src/server.ts
```diff
@@ -133,45 +150,48 @@ async function search(
...
 	});
 
+	const configOverride: Partial<RuntimeConfig> = {
+		includeSingleEpisodes: data.includeSingleEpisodes,
+		includeNonVideos: data.includeNonVideos,
+		excludeRecentSearch: data.ignoreExcludeRecentSearch ? 1 : undefined,
+		excludeOlder: data.ignoreExcludeOlder
+			? Number.MAX_SAFE_INTEGER
+			: undefined,
+		blockList: data.ignoreBlockList ? [] : undefined,
```
mmgoodnow, line 182: never mind, I see the undefined values here

mmgoodnow (approved):

src/decide.ts
ShanaryS: The issue would be validation and what we do if a user breaks something. With a UI we can prevent saving it to config and so can validate it there with immediate feedback. But I think it would create too many problems if the users is editing the file directly.

src/pipeline.ts
ShanaryS: This is not a config option, but a check to allow a data based search for a single episode on sonarr import. Otherwise it would show `is a season pack episode`.

src/runtimeConfig.ts
ShanaryS: Yes it's used as a non-override/use existing config.

src/decide.ts
mmgoodnow: > But I think it would create too many problems if the users is editing the file directly.

Yeah, I agree. By restartless config updates I mean UI-assisted updates to the `settings` table of the database (which currently only has `apiKey` in it, but is named generically on purpose)

mmgoodnow: I'm 1 hour into a 9 hour train ride, so I'm going to spike on storing more config in the database.

src/pipeline.ts
ShanaryS: Changed to `allowSeasonPackEpisodes`

PR #889, opened 2025-01-19
enhancement(pushNotifier): support multiple urls

### LINKED ISSUE #884: Add option of creating multiple  notificationwebhookURLs
I would like to have the option to have multiple  notificationwebhookURLs - for example I would use one for notifiiarr and one for apprise.

Related request - direct support within cross-seed for ntfy or shoutbrr (which I find is easier to use than apprise - and doesn't need apprise to be hosted locally)

ShanaryS:
Fixes #884

Had to use `createCommandWithSharedOptions` for `cross-seed test-notification` since we are doing config validation now.

### DIFF
--- a/src/pushNotifier.ts
+++ b/src/pushNotifier.ts
@@ -14,6 +14,7 @@ import { formatAsList } from "./utils.js";
 export let pushNotifier: PushNotifier;
 
 enum Event {
+	TEST = "TEST",
 	RESULTS = "RESULTS",
 }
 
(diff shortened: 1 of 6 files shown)


src/pushNotifier.ts
```diff
@@ -26,45 +27,46 @@ interface PushNotification {
...
 
 export class PushNotifier {
-	url?: string;
+	urls: string[];
 
-	constructor(url?: string) {
-		this.url = url;
+	constructor(urls: string[]) {
+		this.urls = urls;
 	}
```
mmgoodnow, line 34: TBH I have no idea why I wrote this as a class—it doesn't have any state, lol.

mmgoodnow (approved):

PR #893, opened 2025-01-23
enhancement(decide): treat repack and proper the same

ShanaryS:
We can't rely on REPACK/PROPER consistency between trackers and with data. The check is reduced to if it's a re-release rather than a specific re-release.

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -301,15 +301,10 @@ function sourceDoesMatch(searcheeTitle: string, candidateName: string) {
 }
 
 function releaseVersionDoesMatch(searcheeName: string, candidateName: string) {
-	const searcheeVersionType = stripExtension(searcheeName)
-		.match(REPACK_PROPER_REGEX)
-		?.groups?.type?.trim()
-		?.toLowerCase();
-	const candidateVersionType = stripExtension(candidateName)
-		.match(REPACK_PROPER_REGEX)
-		?.groups?.type?.trim()
-		?.toLowerCase();
-	return searcheeVersionType === candidateVersionType;
+	return (
+		REPACK_PROPER_REGEX.test(searcheeName) ===
+		REPACK_PROPER_REGEX.test(candidateName)
+	);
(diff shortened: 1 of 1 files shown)


zakkarry:
https://github.com/cross-seed/cross-seed/blob/bcc6f462bdd50229864559e9d80ba57a33c0e1be/src/decide.ts#L220-L245

This was accounted for specifically when I implemented it prior, was there a problem with it that it wasn't just readded with the same (or similar) logic?

I can see ways to improve upon this code a bit now, but this implementation never saw any issues I was made aware of, did you find a problem with it?

ShanaryS:
The previous was removed to bring parity to these checks regardless of matchMode when decision caching was added, so users can switch match modes without any cache invalidation. This aspect no longer matters since we are only caching snatched decisions.

The previous check would also always return true for risky/partial, but now we want it to affect those as well. This was brought on by a couple users reporting REPACKs that matched with the original as they came out to be the same size. #882 overcorrected for this by requiring the release type PROPER/REPACK/REPACK1 etc to match. This main downside of this is the arrs always using Proper, but I've also seen incorrect release type between trackers.

I think the best comprise is to not match the original releases with any REPACK/PROPER, but to not require any consistency between two releases of PROPER/REPACK/REPACK1 etc. Essentially betting that multiple REPACKs are rare and they won't come to the same size.

zakkarry:
I agree that that may be the simplest method. If you look at the logic in the original implementation - it necessitated the Arr Proper (which was making assumptions on the case to identify it) would still be vulnerable to mismatches of the same size and inconsistent naming on trackers which do not honor the original naming.

I'm not sure we can do anything within reason to address that besides advising users about when to resume manually.

It should be a rather edge-case for those things to happen (same size and mismatch naming), and I think the only reason it seems to be presenting itself more now Is because of the increase in users we have had since v6 went stable.

mmgoodnow (approved):

PR #894, opened 2025-01-24
feat(cmd): add cmd to update torrent_cache trackers

ShanaryS:
If a user changes their announce key or a tracker url changes, the `torrent_cache` needs to be updated. This can be done using tools like `sed`, but this is not cross-platform and a bit more complicated for users.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,11 +1,7 @@
-import bencode from "bencode";
 import { readFile, rename } from "fs/promises";
 import path from "path";
 import { appDir, createAppDir } from "./configuration.js";
-import { Decision, TORRENT_CACHE_FOLDER } from "./constants.js";
-import { findAllTorrentFilesInDir } from "./torrent.js";
-import { writeFileSync } from "fs";
-import { Torrent } from "./parseTorrent.js";
+import { Decision } from "./constants.js";
 
 createAppDir();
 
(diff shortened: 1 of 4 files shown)


src/cache.ts
```diff
@@ -43,3 +47,55 @@ export async function renameCacheFile(): Promise<void> {
...
+			);
+			if (
+				announce?.includes(oldUrl) ||
+				announceList?.some((t) => t.some((url) => url.includes(oldUrl)))
+			) {
+				count++;
+				console.log(`#${count}: ${filePath}`);
+				let updated = false;
+				if (announce) {
+					const newAnnounce = announce.replace(oldUrl, newUrl);
```
mmgoodnow, line 77: can we call these oldStr and newStr rather than oldUrl and newUrl, given that we support substring replacement?

mmgoodnow (dismissed):

zakkarry:
Is there an reason why we might consider providing this functionality for the client as well?

For instance, at the same time the user is trying to do this, they would potentially also be looking to update the new URL in their client.

The API can be used quite easily in all clients to accomplish this, potentially with an argument to signify "yea update them in deluge too please" - I wouldn't think we would want to default to doing both behaviors though in one command.

This would solve the issue many users have when passkeys or announce domains are changed, although not persistent in most cases as libtorrent generally only saves these updates to fastresume, it is better than leaving the user to "figure it out for themselves" and potentially breaking their .torrent files by ending up with mismatched lengths and breaking things with sed etc.

Thoughts?

ShanaryS:
I think this works but as a separate command. And if we clearly state that the command won't add/remove trackers, then the tiers shouldn't be an issue. I was able to maintain the tracker tiers from each client in a previous implementation of the API searchees.

zakkarry:
> I think this works but as a separate command. And if we clearly state that the command won't add/remove trackers, then the tiers shouldn't be an issue. I was able to maintain the tracker tiers from each client in a previous implementation of the API searchees.

That's what I was saying, either with a special flag in a general "update_tracker" command for -client or -cache, or separate commands entirely.

But seems like related functionality the user is likely to be seeking when using the other.

mmgoodnow:
I think we're pretty borderline on out of scope for `cross-seed` already with this PR to be honest. My preference was to offer a command that just deletes torrents with a given announce host from the cache. It is just a cache after all, and trackers should expect extra snatches when they rotate all users' passkeys. But the implementation was offered so I figured why not. 

A command that updates announce urls in-client feels solidly out of scope to me for the following reasons:
- fairly unrelated to cross-seeding
- more complicated to do via API because we have to write 4 implementations
- already offered by some clients in a GUI
- a lot more dangerous due to mass-editing data that's not backed up or easily undone

I think there's a market for it, but I don't really think it fits into cross-seed. I'm open to changing my mind if either of you feel strongly about it.

zakkarry:
> I think we're pretty borderline on out of scope for `cross-seed` already with this PR to be honest. My preference was to offer a command that just deletes torrents with a given announce host from the cache. It is just a cache after all, and trackers should expect extra snatches when they rotate all users' passkeys. But the implementation was offered so I figured why not.
> 
> A command that updates announce urls in-client feels solidly out of scope to me for the following reasons:
> 
>     * fairly unrelated to cross-seeding
> 
>     * more complicated to do via API because we have to write 4 implementations
> 
>     * already offered by some clients in a GUI
> 
>     * a lot more dangerous due to mass-editing data that's not backed up or easily undone
> 
> 
> I think there's a market for it, but I don't really think it fits into cross-seed. I'm open to changing my mind if either of you feel strongly about it.

Understood, just was broaching the idea since we hadn't gone with simply nuking old matching cached torrents, no harm no foul.

PR #899, opened 2025-01-29
fix(import): remove signalHandlers.js

ShanaryS:
#896 refactored the startup but didn't remove this import


mmgoodnow:
My bad, I had it locally but uncommitted during testing. thanks for the fix!

PR #902, opened 2025-02-03
enhancement(searchee): use meta info in parsed titles

ShanaryS:
This will allow more pre-snatch rejections during decide for parsed titles (e.g. sonarr season folders). Resolution, source, and group needs to be unanimous. Release version only requires a single file.

Since the release version only requires a single file, there could be false positives on sonarr libraries where the episode title contains `Repack|Proper`. This also happens with titles for a movie/show, but is more likely here due to there being more episodes.

### DIFF
--- a/src/searchee.ts
+++ b/src/searchee.ts
@@ -12,6 +12,7 @@ import {
 	MOVIE_REGEX,
 	parseSource,
 	RELEASE_GROUP_REGEX,
+	REPACK_PROPER_REGEX,
 	RES_STRICT_REGEX,
 	SEASON_REGEX,
 	SONARR_SUBFOLDERS_REGEX,
@@ -33,6 +34,7 @@ import {
 	humanReadableSize,
 	inBatches,
 	isBadTitle,
+	isTruthy,
 	stripExtension,
 	WithRequired,
 	WithUndefined,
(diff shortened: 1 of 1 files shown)


zakkarry:
All Sonarr episodes that are repack or proper container "Proper" or the 'v' anime terminaology.

There is no 'Repack'

ShanaryS:
Doesn't the regex cover that?

src/searchee.ts
```diff
@@ -195,6 +197,49 @@ export function getFilesFromDataRoot(
...
+ * @returns Info to add to the title if all files match
+ */
+function parseMetaInfo(videoFileNames: string[]): string {
+	let metaInfo = "";
+	const videoStems = videoFileNames.map((name) => stripExtension(name));
+	const types = videoStems
+		.map((stem) => stem.match(REPACK_PROPER_REGEX)?.groups?.type)
+		.filter(isTruthy);
+	if (types.length) {
+		metaInfo += ` REPACK`;
```
mmgoodnow, line 212: @zakkarry do you mean we should change this to PROPER?

mmgoodnow (approved):

src/searchee.ts
zakkarry: I wasn't necessarily stating that things either needed to be changed or were incorrect as much as I was documenting a nuance that I felt was slightly ambugious in the PR description for both posterity on GH and in case @ShanaryS was unaware.
<hr>

>could be false positives on sonarr libraries where the episode title contains Repack|Proper

The sonarr file names will always be either `v\d` or `Proper` (case specific, btw) - and the only time `Repack` (case insensitive) would be applicable in a Sonarr (or Radarr) library is when either a Custom Format is used for renaming this aspect or original file names are kept and renaming during import is disabled entirely.

If there are any questions about this that I can help answer let me know.

>do you mean we should change this to PROPER?

I haven't had a chance to review the code yet.

PR #906, opened 2025-02-04
fix(rtorrent): batch requests for getClientSearchees

ShanaryS:
I believe this was discussed in the review and it ended being necessary.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -486,6 +486,18 @@ export async function inBatches<T>(
 	}
 }
 
+export async function fromBatches<T, R>(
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -486,6 +486,18 @@ export async function inBatches<T>(
 	}
 }
 
+export async function fromBatches<T, R>(
```
mmgoodnow, line 489: there's probably a way to merge this with inBatches. But fine for now.

mmgoodnow (approved):
batching is complicated :(

PR #907, opened 2025-02-05
fix(cmd): only add options to relevant commands

ShanaryS:
Options such as `searchCadence` should only defined in daemon mode, so that `cross-seed search --no-exclude...` doesn't require setting `searchCadence: null` in the config before hand.

I also updated the `cross-seed diff` command to only show the files. Since this command is to see why two torrents are not a match, `name` and `length` are distracting. If the user wants to see the these, they can use `cross-seed tree` separately (which also displays the parsed title). The files are also sorted in different ways to make the diff easier to read.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -127,7 +127,7 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			"--max-data-depth <depth>",
 			"Max depth to look for searchees in dataDirs",
 			(n) => parseInt(n),
-			fallback(fileConfig.maxDataDepth, 3),
+			fallback(fileConfig.maxDataDepth, 2),
 		)
 		.option(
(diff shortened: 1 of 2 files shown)


mmgoodnow:
Can we just set defaults for searchCadence and rssCadence? 

I did this on purpose to cut down the amount of customization we have to do on individual commands. I thought I made sure everything already had defaults but clearly not

ShanaryS:
> Can we just set defaults for searchCadence and rssCadence?

Since the user config will always have these defined for daemon mode, we can essentially never set it undefined from `cmd`. There might be a way to override it just for search and rss, but it doesn't see like that would be better that this. `--inject-dir` plays a similar role where it affects what gets checked during validation.

zakkarry (approved):

PR #911, opened 2025-02-10
update(matchMode): use "strict" and "flexible" instead of "safe" and "risky"

ShanaryS:
This updates the names for `matchMode` to be less misleading. The new ones are `strict` `flexible` `partial`. This better conveys that these options are equally safe as many users took `safe` and `risky` too literally.

I also think we should update the config template to `matchMode: "partial"` and `seasonFromEpisodes: 0.8`. Users have long asked for `optimal` settings and this should be close to it for most. If we are assuming that users will be setting up linking, similar to how the current options assume a client will be configured, we should start with these options. If a user cannot use linking or they do not want to, the options become opt-out for new users.

I think we have enough evidence that these options bring the user more matches with no real increases in risk. Though I understand if we still want to be conservative with this.

### DIFF
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -301,7 +301,7 @@ module.exports = {
 	 * 1 - must have all episodes (values below 1 requires matchMode: "partial")
 	 * 0.8 - must have at least 80% of the episodes
 	 */
-	seasonFromEpisodes: 0.8,
+	seasonFromEpisodes: 1,
 
 	/**
 	 * You should NOT modify this unless you have good reason.
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -42,8 +42,10 @@ module.exports = {
 	 *
 	 *      sonarr: ["http://sonarr:8989/?apikey=12345",
 	 *               "http://sonarr4k:8989/?apikey=12345"],
+	 *
+	 * For benefits, please read: https://www.cross-seed.org/docs/tutorials/id-searching
 	 */
-	sonarr: undefined,
+	sonarr: [],
 
 	/**
 	 * URL(s) to your Radarr instance(s), included in the same way as torznab
@@ -58,11 +60,13 @@ module.exports = {
 	 *
 	 *       radarr: ["http://radarr:7878/?apikey=12345",
 	 *                "http://radarr4k:7878/?apikey=12345"],
+	 *
+	 * For benefits, please read: https://www.cross-seed.org/docs/tutorials/id-searching
 	 */
-	radarr: undefined,
+	radarr: [],
 
 	/**
-	 * Bind to a specific host address.
+	 * Bind to a specific host address. If you are using docker, keep this as undefined
 	 * Example: "127.0.0.1"
 	 * Default is "0.0.0.0"
 	 */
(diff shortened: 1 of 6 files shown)


bakerboy448:
if you set `0.8` then i suggest adding huge warnings that one will be unable to have seasonpacks auto-upgraded by sonarr if using autobrr as cross-seed would be the only one getting the push.

src/config.template.cjs
```diff
@@ -58,11 +60,13 @@ module.exports = {
...
+	 * For benefits, please read: https://www.cross-seed.org/docs/tutorials/id-searching
 	 */
-	radarr: undefined,
+	radarr: [],
 
 	/**
-	 * Bind to a specific host address.
+	 * Bind to a specific host address. If you are using docker, keep this as undefined
 	 * Example: "127.0.0.1"
 	 * Default is "0.0.0.0"
```
bakerboy448, line 71: `keep this as undefined` and `Default is "0.0.0.0"` are contradictory statements.

zakkarry: Defaults are set when the variable is undefined.

ShanaryS:
> if you set `0.8` then i suggest adding huge warnings that one will be unable to have seasonpacks auto-upgraded by sonarr if using autobrr as cross-seed would be the only one getting the push.

I think it's worth setting `1` for that

zakkarry:
I don't agree with the changes to the config template defaults in the first place, particularly defaulting to partial and less than 1.0 for ensemble.

zakkarry:
> > if you set `0.8` then i suggest adding huge warnings that one will be unable to have seasonpacks auto-upgraded by sonarr if using autobrr as cross-seed would be the only one getting the push.
> 
> I think it's worth setting `1` for that

Can link to the docs, I can write up a description explaining that if you change to <1 than you simply need to move to the import category prior to resuming. It's not really a huge problem considering it has to be manually resumed anyway.

src/config.template.cjs
```diff
@@ -191,23 +195,22 @@ module.exports = {
...
-	 * "partial" is like risky but allows matches if they are missing small
+	 * "partial" is like "flexible" but allows matches even if they are missing small
 	 * files like .nfo/.srt.
 	 *
 	 * We recommend reading the following entries:
 	 * https://www.cross-seed.org/docs/basics/options#matchmode
 	 * https://www.cross-seed.org/docs/basics/faq-troubleshooting#my-partial-matches-from-related-searches-are-missing-the-same-data-how-can-i-only-download-it-once
 	 */
-	matchMode: "safe",
+	matchMode: "partial",
```
zakkarry, line 213:
```suggestion
	matchMode: "flexible",

zakkarry (changes requested):
We should not default to having anything downloaded, this should be an extension people have to opt into, regardless of whether some see it as optimal or not.

ShanaryS:
> I don't agree with the changes to the config template defaults in the first place, particularly defaulting to partial and less than 1.0 for ensemble.

I think that's fair and I don't feel strongly about it. But since we recommend linking anyways, this encourages the users to opt-out rather than opt-in.

zakkarry (dismissed):

src/config.template.cjs
mmgoodnow: why specifically keep it as undefined? setting 0.0.0.0 should also be fine

mmgoodnow:
I agree that we should continue to be conservative with the default settings. I'd rather people configure partial matching with intention than get angry at us for blowing their ratio.

mmgoodnow (approved):

src/config.template.cjs
ShanaryS: The only difference I see is how ipv6 is handled. If it's available it's `::` instead of `0.0.0.0`. So I think specifying it would be good.

Snake883:
> I also think we should update the config template to `matchMode: "partial"` 

I think I just ran into this problem using the [template](https://raw.githubusercontent.com/cross-seed/cross-seed/refs/heads/master/src/config.template.cjs).  The template has the option "flexible", but that has led to an invalid configuration.

zakkarry:
> > I also think we should update the config template to `matchMode: "partial"`
> 
> I think I just ran into this problem using the [template](https://raw.githubusercontent.com/cross-seed/cross-seed/refs/heads/master/src/config.template.cjs). The template has the option "flexible", but that has led to an invalid configuration.

gen-config will grab from master, not the release, so it will be outdated currently I believe.

@ShanaryS we probably should publish this.

zakkarry:
or push a correction, and remerge the change when we're ready to publish.

mmgoodnow:
gen config pulls from the container

ShanaryS:
I think they were on `:master` but didn't setup `linkDirs`. The goal is to make linking opt-out like `useClientTorrents: true` and `action: inject` since it's better for the vast majority of users. The template config can never work anyways since the users need to add their own urls and paths.

PR #916, opened 2025-02-20
update(outputDir): default to /config/cross-seeds

ShanaryS:
Users frequently misconfigure `outputDir` in a few ways:
- It's to where the client copies .torrent files
- It's used as a watch directory
- Programs such as qbit_manage delete the folder
- rTorrent does not have access to it with the same mount

These are solved by `outputDir: null`. This will be set to `join(appDir(), "cross-seeds")` so it works for npm users and all platforms.

To solve the rTorrent issue, it needed to be decoupled from outputDir. `load.normal` and `load.start` have been replaced with `load.raw` and `load.raw_start`. These are the same methods for injecting, but take a base64 encoded .torrent instead of the path. I will ping the rTorrent role in pre-release for testing, but there should be no issues. This method as been around since 2007 and the arrs use it as well.

This means that we can get rid of the `:/cross-seeds` mount. Now users only need a single mount to use cross-seed, `:/config`, along with other mounts to their data if they are using linking. This will make setup for first time users easier and eliminate all support issues around `outputDir`.

### DIFF
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -369,7 +369,7 @@ export const VALIDATION_SCHEMA = z
 				}
 				return true;
 			}),
-		injectDir: z.string().optional(),
+		injectDir: z.string().nullish(),
 		includeSingleEpisodes: z.boolean(),
 		includeNonVideos: z.boolean(),
 		fuzzySizeThreshold: z


mmgoodnow:
Big fan of this change. The idea of a magic string is a little weird though. What if we just allow it to be unset?

ShanaryS:
Yeah this is a much better way to do this. Since `outputDir` has been required before this, no users can be impacted by this.

zakkarry:
@mmgoodnow I'm not sure if it's still an issue, but didn't we have a problem with loading the .torrent as base64 before, and then as a result had that permissions debacle on the .torrent files, at the time we didn't change to this method, was there a reason or is that resolved now with this?

mmgoodnow:
That was qbittorrent. We've always used load.start for rtorrent, I think I was copying Flood at the time. 

They switched to passing base64 in https://github.com/jesec/flood/commit/622ec7f5156c2875bfd8b257a80c6b08c910d583 which might be after I implemented it. I remember seeing load_raw in the documentation at the time but I think there was a quoted danger of the base64 encoded torrent file blowing the max XMLRPC message size and the call failing. 

Before that commit was this commit which I think was probably on master when I implemented it: https://github.com/jesec/flood/commit/96c754ddeb614b45a565e8307c9985ee85bcb7fa#diff-6611a53b93130f462793029975382b5d05287c42f4b3ceac1e5ce5b2a8dbe519L66-L67

mmgoodnow:
Should be fine given that the max size is configurable and it seems like all the other rtorrent integrators use load_raw now.

PR #933, opened 2025-03-06
update(zod/maxdatadepth): classify greater than 3 depth as warning

zakkarry:
Using a `maxDataDepth` of more than 3 cross-seed will issue an error. This is misleading, as there are certain rare edge-case folder structures that can be valid and necessitate a depth greater than 3.

This reduces the error to a warning.


ShanaryS (approved):
Using `3` still seems to be the maximum optimal value, but it just may not be practical for the user.

PR #939, opened 2025-03-15
enhancement(inject): increase maximum auto resume to 100 MiB

ShanaryS:
A lot of users have asked for an increase to `100 MiB` since a lot of samples top out in the `50-100  MiB` range. I changed the default to this new max since most users would probably want it, though the partial matching tutorial mentions this option anyways.

The main downside is a higher chance of `cross-seed` resuming a partial season pack that's missing an episode, but it's extremely rare for an episode to be less than `100 MiB`.

I think this should be the absolute limit for this setting. @zakkarry mentioned alternative methods for resuming, but I think that would need to be saved for v7.

I would like to reiterate here that these matches still need to respect `fuzzySizeThreshold`. So the minimum torrent size that could use a `100 MiB` auto resume is `1 GiB` or `5 GiB` with the default fuzzy.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -91,7 +91,7 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			"--auto-resume-max-download <number>",
 			"The maximum size in bytes remaining for a torrent to be resumed",
 			parseInt,
-			fallback(fileConfig.autoResumeMaxDownload, 52428800),
+			fallback(fileConfig.autoResumeMaxDownload, 104857600),
 		)
 		.option(
 			"--link-category <cat>",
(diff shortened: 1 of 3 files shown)


zakkarry:
> I think this should be the absolute limit for this setting. @zakkarry mentioned alternative methods for resuming, but I think that would need to be saved for v7.

My method is not a breaking change, I'm not sure why it would need to be saved for v7, I was planning on pushing to a branch later today and opening a PR.

ShanaryS:
I thought you were proposing to replace the current behavior. We can move all discussions for resuming to your PR, depending on the results, this may not be needed as you said.

PR #940, opened 2025-03-16
feat(inject/ignoreFiles): ignore certain non-relevant files for resume behavior with partial matches

zakkarry:
The behavior idea I had in mind involves excluding known non-relevant files as a second measure to check if resuming is desired. This should cover most instances where any sample or miscellaneous file commonly included in some trackers prevents resuming and provides a larger buffer to resume with.

The option to enable this behavior is `ignoreNonRelevantFilesToResume`. Defaults, if not included in the config (onboarding), to false. The previous behavior stays intact with the 50MB limit, but I'm not sure that increasing it as @ShanaryS suggests in #939 is necessary given this option's functionality, and I would contend that that is not the best or proper "total" solution for the job.

### DIFF
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -30,7 +30,7 @@ import {
 	wait,
 } from "../utils.js";
 import {
-	calculateSizeForAutoResume,
+	shouldResumeFromNonRelevantFiles,
 	ClientSearcheeResult,
 	getMaxRemainingBytes,
 	getResumeStopTime,
@@ -336,9 +336,8 @@ export default class Deluge implements TorrentClient {
 				}
 				if (torrentInfo.total_remaining! > maxRemainingBytes) {
 					if (
-						!calculateSizeForAutoResume(
-							meta.files,
-							torrentInfo.total_size!,
+						!shouldResumeFromNonRelevantFiles(
+							meta,
 							torrentInfo.total_remaining!,
 							torrentLog,
 							this.label,
--- a/src/clients/Deluge.ts
+++ b/src/clients/Deluge.ts
@@ -1,6 +1,7 @@
 import { readdirSync } from "fs";
-import { basename } from "path";
 import ms from "ms";
+import { basename } from "path";
+import { inspect } from "util";
 import {
 	DecisionAnyMatch,
 	InjectionResult,
@@ -29,6 +30,7 @@ import {
 	wait,
 } from "../utils.js";
 import {
+	calculateSizeForAutoResume,
 	ClientSearcheeResult,
 	getMaxRemainingBytes,
 	getResumeStopTime,
@@ -37,9 +39,9 @@ import {
 	resumeSleepTime,
 	shouldRecheck,
 	TorrentClient,
+	TorrentExclusionInfo,
 	TorrentMetadataInClient,
 } from "./TorrentClient.js";
-import { inspect } from "util";
 
 interface TorrentInfo {
 	name?: string;
@@ -297,14 +299,19 @@ export default class Deluge implements TorrentClient {
 	/**
 	 * checks the status of an infohash in the client and resumes if/when criteria is met
 	 * @param infoHash string containing the infohash to resume
+	 * @param decision decision by which the newTorrent was matched
+	 * @param options options object for extra flags
 	 * @param options.checkOnce boolean to only check for resuming once
+	 * @param options.meta metafile object containing the torrent data
 	 */
 	async resumeInjection(
 		infoHash: string,
 		decision: DecisionAnyMatch,
-		options: { checkOnce: boolean },
+		options: { checkOnce: boolean; meta: Metafile },
 	): Promise<void> {
 		let sleepTime = resumeSleepTime;
+		const { fuzzySizeThreshold, ignoreNonRelevantFilesToResume } =
+			getRuntimeConfig();
 		const maxRemainingBytes = getMaxRemainingBytes(decision);
 		const stopTime = getResumeStopTime();
 		let stop = false;
@@ -315,6 +322,7 @@ export default class Deluge implements TorrentClient {
 			}
 			await wait(sleepTime);
 			let torrentInfo: TorrentInfo;
+			let exclusionInfo: TorrentExclusionInfo;
 			let torrentLog: string;
 			try {
 				torrentInfo = await this.getTorrentInfo(infoHash);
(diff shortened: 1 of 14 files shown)


stickel:
This feels like a better long term solution to the issue. Raising the limit to feels temporary and would need to be addressed in the future again.

From what I understand of the issue, this should eliminate the root cause (samples, larger extraneous files) in most situations.

ShanaryS:
I have no issues with this feature being added in v6, I was under the impression you wanted to replace the current behavior, which would need to be in v7.

A few weeks ago in [lounge](https://discord.com/channels/880949701845872672/880949702349164556/1344098754843639848) when this was brought up, these were the comments you made:

> I was thinking about changing this to a exempting of certain file types and keywords, not sure if you had considered doing this when implementing it

> for scene, i've found that even for half hour TV shows 50MB does not cover it, i would think 100 would, but i also wonder about whether or not we should just ouright exclude nfo, sample and proof folders and keep it to fuzzysize

I understood these comments to mean replacing `autoResumeMaxDownload`, which I'm not opposed to. But I still think we would need some size limit on these files.

I also feel this option could be controversial as people may have opinions on what is or isn't relevant. I tried to avoid this with `matchMode: "partial"` and `autoResumeMaxDownload`, but it seems it may be needed in some form.

bakerboy448:
> This feels like a better long term solution to the issue. Raising the limit to feels temporary and would need to be addressed in the future again.
> 
> From what I understand of the issue, this should eliminate the root cause (samples, larger extraneous files) in most situations.

concur

src/clients/TorrentClient.ts
```diff
@@ -243,8 +262,58 @@ export function getMaxRemainingBytes(decision: DecisionAnyMatch) {
...
+					? acc.relevantSize
+					: acc.relevantSize + file.length,
+				excludedFileCount: shouldExclude
+					? acc.excludedFileCount + 1
+					: acc.excludedFileCount,
+			};
+		},
+		{ relevantSize: 0, excludedFileCount: 0 },
+	);
+}
```
ShanaryS, line 319: I think doing the check here is best since we don't need any api calls for it.
```ts
export function calculateSizeForAutoResume(files: File[], totalSize: number, remainingSize: number, torrentLog: string, label: string): boolean {
	const { fuzzySizeThreshold, ignoreNonRelevantFilesToResume } = getRuntimeConfig();
	if (!ignoreNonRelevantFilesToResume) return false;
	if (remainingSize > 209715200) return false; // 200 MiB limit

	const { excludedFileCount, relevantSize } = files.reduce(
		(acc, file) => {
			const shouldExclude =
				RESUME_EXCLUDED_KEYWORDS.some(keyword =>
					[file.path, file.name].some(text =>
						text.toLowerCase().includes(keyword.toLowerCase())
					)
				) ||
				RESUME_EXCLUDED_FILETYPES.some(fileType =>
					file.name.toLowerCase().endsWith(fileType.toLowerCase())
				);
			if (shouldExclude) {
				logger.verbose({
					label,
					message: `${torrentLog}: excluding file ${file.path} from auto resume check`,
				})
				return {
					relevantSize: acc.relevantSize,
					excludedFileCount: acc.excludedFileCount + 1,
				};
			}
			return {
				relevantSize: acc.relevantSize + file.length,
				excludedFileCount: acc.excludedFileCount,
			};
		}, { relevantSize: 0, excludedFileCount: 0 }
	);

	if (!excludedFileCount) return false;
	if ((totalSize - remainingSize) * (1 - fuzzySizeThreshold) < relevantSize) {
		return false;
	}
	return true;
}
```

src/clients/Deluge.ts
```diff
@@ -327,6 +335,36 @@ export default class Deluge implements TorrentClient {
...
+						logger.info({
+							label: this.label,
+							message: `Resuming ${torrentLog}: ${humanReadableSize(torrentInfo.total_remaining!, { binary: true })} remaining and ${exclusionInfo.excludedFileCount} files excluded (relevant size: ${humanReadableSize(exclusionInfo.relevantSize, { binary: true })}`,
+						});
+						await this.call<string>("core.resume_torrent", [
+							[infoHash],
+						]);
+						return;
+					}
+				}
```
ShanaryS, line 367: I feel like this block would be simpler if we do this check after checking for `torrentInfo.total_remaining! > maxRemainingBytes` since we will always resume if that's false. We can also move this logic to the `calculateSizeForAutoResume` function and pass the client specific args to it.

```ts
if (torrentInfo.total_remaining! > maxRemainingBytes) {
	if (!calculateSizeForAutoResume(meta.files, torrentInfo.total_size!, torrentInfo.total_remaining!, torrentLog, this.label)) {
		logger.warn({
			label: this.label,
			message: `Will not resume ${torrentLog}: ${humanReadableSize(torrentInfo.total_remaining!, { binary: true })} remaining > ${humanReadableSize(maxRemainingBytes, { binary: true })}`,
		});
		return;
	}
}
```
This goes with my suggestion for the new function.

src/clients/Transmission.ts
```diff
@@ -434,7 +439,7 @@ export default class Transmission implements TorrentClient {
 			const queryResponse = await this.request<TorrentGetResponseArgs>(
 				"torrent-get",
 				{
-					fields: ["leftUntilDone", "name", "status"],
+					fields: ["leftUntilDone", "name", "status", "files"],
```
ShanaryS, line 442: I don't think the `files` are necessary anymore since we are using `Metafile` right?

ShanaryS (commented):
>  Whether or not there should be an interaction between the two resume options seems to be my main "undecided" option, sort of like a hard limit even if the files are deemed irrelevant...anything else please let me know what you think

I think we should try the current method first, then fallback to this if the size is larger. I suggested a hard limit of `200 MiB` for this feature below but I think that's up for debate. I think it's better to start conservative upon release then we can raise it later if no issues arise.

> Also possible other file types and keywords can be added by turning the option in config into a string array or something.

We should keep it hardcoded for now. If anyone has any suggestions we can add them. In the future can always add the config option if needed (probably turn `ignoreNonRelevantFilesToResume` into an array).

src/clients/Transmission.ts
zakkarry: I thought I had removed that, it must have gotten added back when I rebased.

src/clients/Deluge.ts
```diff
@@ -295,14 +297,19 @@ export default class Deluge implements TorrentClient {
...
+	 * @param decision decision by which the newTorrent was matched
+	 * @param options options object for extra flags
 	 * @param options.checkOnce boolean to only check for resuming once
+	 * @param options.meta metafile object containing the torrent data
 	 */
 	async resumeInjection(
 		infoHash: string,
 		decision: DecisionAnyMatch,
-		options: { checkOnce: boolean },
+		options: { checkOnce: boolean; meta: Metafile },
```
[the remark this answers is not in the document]
zakkarry: True. Makes sense, wasn't sure and just wanted to guts of this for review so put it in option.

src/clients/TorrentClient.ts
zakkarry:
```ts
export function calculateSizeForAutoResume(files: File[], totalSize: number, remainingSize: number, torrentLog: string, label: string): boolean {
	const { fuzzySizeThreshold, ignoreNonRelevantFilesToResume } = getRuntimeConfig();
	if (!ignoreNonRelevantFilesToResume) return false;
	if (remainingSize > 209715200) return false; // 200 MiB limit
......
```

I'm not sure we have the same behavior in mind, I'd like to go over this with you on discord and double-check we are thinking this accomplishes the same goals if you don't mind.

src/clients/TorrentClient.ts
```diff
@@ -243,8 +257,69 @@ export function getMaxRemainingBytes(decision: DecisionAnyMatch) {
...
+			}
+			return {
+				relevantSize: acc.relevantSize + file.length,
+				excludedFileCount: acc.excludedFileCount,
+			};
+		},
+		{ relevantSize: 0, excludedFileCount: 0 },
+	);
+
+	if (!excludedFileCount) return false;
```
ShanaryS, line 320: I think we can actually get rid of `excludedFileCount` and use `if (relevantSize === totalSize) return false;`. The verbose logging of the excluded files should be enough to understand what `cross-seed` is doing.

src/clients/Deluge.ts
```diff
@@ -328,11 +333,21 @@ export default class Deluge implements TorrentClient {
...
 				if (torrentInfo.total_remaining! > maxRemainingBytes) {
-					logger.warn({
-						label: this.label,
-						message: `Will not resume ${torrentLog}: ${humanReadableSize(torrentInfo.total_remaining!, { binary: true })} remaining > ${humanReadableSize(maxRemainingBytes, { binary: true })}`,
-					});
-					return;
+					if (
+						!calculateSizeForAutoResume(
+							meta.files,
+							torrentInfo.total_size!,
```
ShanaryS, line 339: I think we should use `meta.length` for `Deluge` and `qBittorrent` to be consistent with the other clients.

ShanaryS (dismissed):

ShanaryS (dismissed):

ShanaryS (dismissed):

ShanaryS (approved):

ShanaryS (approved):

zakkarry:
@mmgoodnow Since this is a new option, @ShanaryS and I were curious if you wanted approval before merging this.

mmgoodnow (approved):

PR #948, opened 2025-04-04
fix(regex): cover edgecases in scene title parsing

ShanaryS:
Covers parsing titles should as `'9-1-1.S01` into `1-1`.


zakkarry:
I will consult with someone about the naming convention we are working 'against' to see if there are any certainties rather than just my assumptions.

zakkarry:
I've found that 2 is possible, however it is exceedingly rare and most likely not a major concern as it seems to be with very old releases. As I stated when I presented this regex initially 3-6 was standard, with 3-5 being the majority, so I think we should be fine.

zakkarry (approved):

PR #960, opened 2025-04-19
feat(server): refactor server.ts to use Fastify

mmgoodnow:
## Summary
- Replace Node.js native HTTP server with Fastify framework for better performance, route management, and middleware support
- Maintain all existing endpoint functionality with same authentication method
- Improve type safety throughout request handling

## Test plan
- Verify all API endpoints function the same as before
- Check that API keys still work via header and query parameter
- Ensure error handling remains robust

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,9 +1,13 @@
 import chalk from "chalk";
 import { existsSync } from "fs";
-import http, { IncomingMessage, ServerResponse } from "http";
 import { parse as qsParse } from "querystring";
 import { inspect } from "util";
 import { z } from "zod";
+import fastify, {
+	FastifyInstance,
+	FastifyRequest,
+	FastifyReply,
+} from "fastify";
 import { checkApiKey } from "./auth.js";
 import {
 	ActionResult,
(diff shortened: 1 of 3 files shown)


zakkarry:
it's a little out of scope of this PR I know, but would any discussion around possibly implementing some form of the status endpoint beyond a mirror of ping be an option? maybe if its just a report of the current jobs as they're reported at startup (lastrun/nextrun/running now) - just to get something on there?

src/server.ts
```diff
@@ -121,108 +113,32 @@ function parseData(data: string) {
...
+		Querystring: { apikey?: string };
+	}>,
+	reply: FastifyReply,
 ): Promise<boolean> {
-	const url = new URL(req.url!, `http://${req.headers.host}`);
 	const apiKey =
-		(req.headers["x-api-key"] as string) ?? url.searchParams.get("apikey");
+		(request.headers["x-api-key"] as string) || request.query.apikey || "";
 	const isAuthorized = await checkApiKey(apiKey);
 	if (!isAuthorized) {
```
ShanaryS, line 124: I think it's better to use a guard clause in cases like these, less indentation.

```ts
if (isAuthorized) return true
...
return false
```

ShanaryS (approved):
Didn't see any changes in behavior but haven't tested it.

mmgoodnow:
I'm not opposed, but won't be adding that to this PR

mmgoodnow:
so far, there's a slight breaking change in the data parser before this didn't check `Content-Type`.

PR #1016, opened 2025-07-16
enhancement(client/injection): prevent subsequent rechecks for disc images on reinjection attempts

Audionut:
Remove disc based content from recheck on reinjection.

Disc based content requires careful consideration and manual intervention, when the initial hash check fails to complete to 100%. autoresume is also disabled for disc based content for essentially the same reason.

Resuming failed hash check disc content will overwrite pieces in the existing torrent, and invalidate the existing torrent.

For this reason, recheck for disc based content on reinjection attempt, will never give a useful result, and will only ever increase drive wear and waste system resources.

### DIFF
--- a/src/inject.ts
+++ b/src/inject.ts
@@ -20,6 +20,7 @@ import {
 	SaveResult,
 	TORRENT_CACHE_FOLDER,
 	UNKNOWN_TRACKER,
+	VIDEO_DISC_EXTENSIONS,
 } from "./constants.js";
 import { assessCandidate } from "./decide.js";
 import { Label, logger } from "./logger.js";
@@ -45,6 +46,7 @@ import {
 	findFallback,
 	formatAsList,
 	getLogString,
+	hasExt,
 	humanReadableDate,
 	isTruthy,
 	Mutex,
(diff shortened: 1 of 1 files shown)


zakkarry:
I would argue that's not really a "solution" - it's more "avoiding" the problem while adding to the workflow. There are plenty of situations where discs can be rechecked to 100% and this just leaves more intervention for the user by requiring them to manually recheck.

As a better option have you considered perhaps only rechecking the discs one time when it is injected rather than repeatedly? I believe this can be accomplished by "when a disc is identified" not saving the file to the outputDir, resulting in it not rechecking hourly. This would seem to solve the issue while also not adding to the workflow/workload of cross-seeding images.

Thoughts?

Audionut:
The better option is what I actually want to do. Distracted doing something else, will get back to this.

Thanks for your time.

zakkarry:
Is there a scenario where you think there isn't a need for at least one recheck performed? I'm not sure I understand your logic, given that to both start and handle (delete mismatched files, for instance) the cross-seeded disc would need to be rechecked to gain information on integrity, or at least have the recheck performed to start the torrent if it's a 100% match.

zakkarry:
I just had a chance to sign onto a computer and look at what you did in the code - your PR description or title doesn't really outline the code you've changed or what effect you have, in fact, it's somewhat misleading.

Either way, I will give it a closer look in the morning.

zakkarry:
I've renamed your PR for you, but since you are more atune with the reasoning behind this PR and disc cross-seeding than I likely am, it would be nice if your PR description better described the problem, reason, and effect the code changes you've made had. Currently, there is no context or reasoning for this change.

You can describe it very briefly, and perhaps even put a link to the latest discussion in the Discord that prompted this PR. That shouldn't take but a few seconds to do. 

Other than that, reviewing your code changes and discussions we've had suggests that this is fine otherwise.

Audionut:
Thanks Zac. I've edited the op. AI would probably describe it better tbh.

zakkarry:
What you've written is fine, @Audionut. I will discuss the next planned release with @ShanaryS and plan for merging, assuming @mmgoodnow has no objections and we've no pending PRs that we want to include in the next release I'd guess it would be sometime this week/weekend. This could easily be a patch/fix release, so either way shouldn't be any problem.

Thanks for the PR.

zakkarry (approved):

zakkarry:
This has been merged and published now @Audionut - just in case you missed it.

PR #1026, opened 2025-08-16
perf(decide): speed up and reduce event loop pressure

ShanaryS:
With some heavy profiling, the `rss` event loop issues should now be resolved. This should reduce the `TypeError: terminated` as well as make the API more responsive while `rss` is running. The main culprit is `Fuse.js` fuzzy lookup, which only happens if the `EPISODE`, `SEASON`, and `MOVIE` regexes is fail. It can take over a second per candidate which blocks the event loop.

The solution is to get rid of `Fuse.js` completely since it's not being used for the main media types #725. In it's place we do the same levenshtein fuzzy on the entire string rather than the regex parsed title. This is not only faster but we can may it more async friendly with the following change.

Added a new function `yieldToEventLoop()` which does a `setTimeout` and then `setImmediate` which should guarantee all pending tasks to run once regardless of event loop phase. This is mainly used in `filterAsyncYield()` which filters with `fromBatches` and calls `yieldToEventLoop()` in between. This is used to break up large otherwise sync processing.

`rss` candidates will also acquire `AsyncSemaphore` to limit concurrent candidate processing to only one at a time. The fact that many sync operations can run back to back is what causes a lot of the responsive issues. All trackers process once candidate from their pool one at a time round robin style. This ensures they all are processed concurrently while still maintaining the the `AsyncGenerator` paging behavior.

Finally, `guidInfoHashMap` can take many seconds to generate and is done for each candidate. Some of that time is spend doing synchronous tasks which also was blocking the event loop. Now we will load this once at startup and modify it with each new assessment. All assessments will share this map saving seconds per candidate as well as memory. The daily `cleanup` job will reset this to be safe.

A side effect of the `guidInfoHashMap` change is it speeds up search as well. Particularly sequential skipped searches now happen all at once rather than each being affected by this delay.

The inject job should now also be faster as we are no longer checking `areMediaTitlesSimilar` on the each searchee and candidate files. This was adding a couple seconds per .torrent for very little again.

### DIFF
--- a/src/dataFiles.ts
+++ b/src/dataFiles.ts
@@ -21,11 +21,11 @@ import {
 	createKeyTitle,
 	exists,
 	filterAsync,
+	filterAsyncYield,
 	flatMapAsync,
 	inBatches,
 	isTruthy,
 	mapAsync,
-	yieldToEventLoop,
 } from "./utils.js";
 import { isOk } from "./Result.js";
 import { distance } from "fastest-levenshtein";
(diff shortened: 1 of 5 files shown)


mmgoodnow:
Nice, glad we're getting rid of Fuse. It caused a lot of long tasks and they were completely synchronous so logs wouldn't even print during that time.

Issue #1031, opened 2025-08-23
Support COPYFILE_FICLONE


hexchen:
### Does your feature request solve a problem you're facing?

ZFS supports reflink copies, including between datasets; but only when using `cp --reflink=auto` or setting `fs.constants.COPYFILE_FICLONE`.

I have stored my media libraries on different datasets, and would like to use reflink copies for cross-seed. As of right now, I can't do that. I would have to place my linkDirs inside of my dataDirs, which is (for obvious reasons) not possible.

### Does an existing issue/feature request already exist?

- [x] I've searched far and wide!

### What would you like for cross-seed to do?

Add a new `linkType` to allow reflink copying with `fs.constants.COPYFILE_FICLONE`. 

### What are you currently doing, if anything, to solve or work around the issue?

Patching `linkFile` in `action.ts to use `fs.constants.COPYFILE_FICLONE` instead of `fs.constants.COPYFILE_FICLONE_ALWAYS`.

[0001-feat-actions-use-COPYFILE_FICLONE.patch](https://github.com/user-attachments/files/21952516/0001-feat-actions-use-COPYFILE_FICLONE.patch)

### Additional comments

N/A

PR #1032, opened 2025-08-23
feat(linking/reflinkOrCopy): allow non-forced reflinks

hexchen:
This allows to use reflinks on ZFS if linkDir and dataDir span multiple datasets. It has the added danger of silently falling back to copying data if a reflink is not successful; this caveat needs to be highlighted in the documentation.

See #1031 for details.

### DIFF
--- a/src/action.ts
+++ b/src/action.ts
@@ -362,7 +362,8 @@ async function getClientAndDestinationDir(
 					await linkFile(
 						srcPath,
 						testPath,
-						linkType === LinkType.REFLINK || linkType === LinkType.REFLINK_OR_COPY
+						linkType === LinkType.REFLINK ||
+							linkType === LinkType.REFLINK_OR_COPY
 							? linkType
 							: LinkType.HARDLINK,
 					);
@@ -721,7 +722,8 @@ async function getLinkDir(pathStr: string): Promise<string | null> {
 				await linkFile(
 					srcFile,
 					testPath,
-					linkType === LinkType.REFLINK || linkType === LinkType.REFLINK_OR_COPY
+					linkType === LinkType.REFLINK ||
+						linkType === LinkType.REFLINK_OR_COPY
 						? linkType
 						: LinkType.HARDLINK,
 				);


ShanaryS:
I would ideally like a way to check this actually linked but that would have to be post linking and very involved.

Your explaination of why this works makes sense but is extremely niche. I know some users will select this without understanding its use. I'm in favor of not updating the comment in the config file and only talking about this option in the reflinks section of the docs which lots of warnings.

The cost of a user copying all their data is too dangerous and they may not even realize it's cross-seed's doing. It could create terabytes of data overnight. Linking as it is already what causes the most support tickets but now it could cause harm as well if misconfigured.

hexchen:
> I would ideally like a way to check this actually linked but that would have to be post linking and very involved.

Regrettably, I don't think there is a filesystem-agnostic way of checking.

> Your explaination of why this works makes sense but is extremely niche. I know some users will select this without understanding its use. I'm in favor of not updating the comment in the config file and only talking about this option in the reflinks section of the docs with lots of warnings.
> 
> The cost of a user copying all their data is too dangerous and they may not even realize it's cross-seed's doing. It could create terabytes of data overnight. Linking as it is already what causes the most support tickets but now it could cause harm as well if misconfigured.

That seems like a very good idea, considering that a warning in the docs can be much more visible than in the config file.

With regards to the other changes needed: I noticed that my current PR is incomplete as I was testing the code in my actual deployment. I've marked the PR as a draft for now, and will mark is as ready once I have verified that it works.

hexchen:
Seems like all that was missing was in fact to change those two occurences of `LinkType.REFLINK` to also have the same behaviour with `reflinkOrCopy`. Tested on my seedbox, works perfectly (reflink copy accross two zfs datasets on the same zpool).

zakkarry:
Please tag me in the documentation if you decide to take on writing it, I want to be absolutely sure that this is prefaced with the proper warnings and caveats before merging this to master.

Thanks.

ShanaryS (approved):

PR #1045, opened 2025-09-12
perf(torrent_cache): reduce disk I/O operations for lookup

ShanaryS:
We will now use a map of `infoHash` to `torrentPath` for working with cached torrents, similar to `guid` to `infoHash`. This makes lookups faster and less I/O intensive and should also reduce memory in case of lots of simulatenous usages.

With this change, the implementation of #1042 has also been updated. We can prune the decision table with every run of the cleanup job since we are no longer need any I/O operations aside from database lookups.


src/decide.ts
```diff
@@ -514,46 +491,27 @@ async function getCachedTorrentFile(
...
+		const torrentPath = path.join(
+			appDir(),
+			TORRENT_CACHE_FOLDER,
+			getCachedTorrentName(meta.infoHash),
 		);
 		await writeFile(torrentPath, new Uint8Array(meta.encode()));
+		if (candidate.indexerId && meta.trackers.length) {
+			await db("indexer")
+				.where({ id: candidate.indexerId })
+				.update({ trackers: JSON.stringify(meta.trackers) });
```
mmgoodnow, line 513: I think we should append rather than replace here

PR #1047, opened 2025-09-19
Allow for ignoring the release group

calumapplepie:
Sometimes, the release group is actively changed by reuploaders moving files around.  Obviously, this makes matching much more difficult.  This is a quick hack to ignore release groups.

I couldn't figure out how to make the cache be ignored when this option is used; and I've thought of an alternate approach to this problem that I think I would like to try as well.  Submitting this PR in case this approach is preferred.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -84,6 +84,11 @@ function createCommandWithSharedOptions(name: string, description: string) {
 				.makeOptionMandatory(),
 		)
 		.option(
+			"--ignore-release-group",
+			"Don't consider the release group during matching",
+			fallback(fileConfig.ignoreReleaseGroup, false)
+		)
+		.option(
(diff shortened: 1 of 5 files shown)


zakkarry:
This could result in snatching behavior that isn't exactly "kind" to the trackers, particularly with WEB releases where the size is often very similar across groups. Have you found this to be a widely spread problem? This generally is only an issue on public trackers as far as I'm aware.

mmgoodnow:
Echoing @zakkarry - this functionality is a non starter. Those reuploaders should stop modifying the release groups.

calumapplepie:
@zakkarry : I suspect you are right, that it is only an issue of public trackers.

@ShanaryS : thanks for the tip! I didn't realize inject would work that way; will keep in mind

PR #1048, opened 2025-09-19
Allow users to filter out undesired words from titles

calumapplepie:
Sometimes a torrent title will include extra words (eg, "- Copy"), or the name of the tracker it is uploaded to.  Other examples include phrases like "epub/mobi" in book torrent titles, or made-up public tracker identities.  While this is obviously bad for uploaders to do, it is a pain to manually work around.

While it could be possible to expand the existing filtering, removing those extra words for all users, doing so risks bugs; for instance, blindly removing "epub" and "mobi" from all book titles would make "Republican mobility problems" into "Rlican lity problems".  Thus, this option lets users test out these changes on their own, and potentially report back with working changes.  If combined with an allowlist mode, users can experiment much more easily; but I'm not going to try and implement such an option.

### DIFF
--- a/src/cmd.ts
+++ b/src/cmd.ts
@@ -268,6 +268,12 @@ function createCommandWithSharedOptions(name: string, description: string) {
 			fallback(fileConfig.blockList, []),
 		)
 		.option(
+			"--title-filter-list <strings...>",
+			"Text which should be ignored if it appears in the torrent title (eg, tracker names)",
+			// @ts-expect-error commander supports non-string defaults
+			fallback(fileConfig.titleFilterList, []),
+		)
+		.option(
 			"--sonarr <urls...>",
 			"Sonarr API URL(s)",
 			// @ts-expect-error commander supports non-string defaults
--- a/src/config.template.cjs
+++ b/src/config.template.cjs
@@ -387,4 +387,10 @@ module.exports = {
 	 * https://www.cross-seed.org/docs/basics/options#blocklist
 	 */
 	blockList: [],
+
+	/**
+	 * Remove these words/substrings from titles before searching.
+	 * Text in brackets is always removed
+	 */
+	titleFilterList: []
 };
--- a/src/configSchema.ts
+++ b/src/configSchema.ts
@@ -522,6 +522,7 @@ export const VALIDATION_SCHEMA = z
 		verbose: z.boolean(),
 		torrents: z.array(z.string()).optional(),
 		blockList: z.array(z.string()).nullish().transform(transformBlocklist),
+		titleFilterList: z.array(z.string()),
 		apiKey: z.string().min(24).nullish(),
 		radarr: z
 			.array(z.string().url())
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -57,6 +57,7 @@ export interface FileConfig {
 	searchTimeout?: string;
 	searchLimit?: number;
 	blockList?: string[];
+	titleFilterList?: string[];
 	apiKey?: string;
 	sonarr?: string[];
 	radarr?: string[];
--- a/src/runtimeConfig.ts
+++ b/src/runtimeConfig.ts
@@ -38,6 +38,7 @@ export interface RuntimeConfig {
 	searchTimeout?: number;
 	searchLimit?: number;
 	blockList: string[];
+	titleFilterList: string[];
 	apiKey?: string;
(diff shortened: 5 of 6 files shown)


mmgoodnow:
Can you provide an example of how you're using this?

zakkarry:
> I have seen these uploads on multiple private trackers.  From what I can tell, many trackers don’t patrol torrent file names seriously.  Removing and forcing re-uploads based on file titles is a significant load on both moderators and tracker users. As for books being unsupported, that seems odd, given that there is dedicated code to improve handling of book files.  If books are entirely unsupported, such that adding new options to handle them better is completely out of scope for the project, then perhaps that existing code should be removed and this limitation documented.

I'm not sure what private trackers you're referring to, but bad uploads (filename and foldernames where applicable) are commonly reasons for trumping/nuking of releases, particularly if " - Copy" - an artifact from windows, or a bad filename was uploaded in the .torrent. I've not seen instances of these releases being kept, and torrent checkers are present on pretty much all reputable private trackers I'm aware of for this very reason.

As @mmgoodnow said, can you give an example? As for the non-supported mediatypes, we have code that ensures that searches are performed properly and we don't assume everything is a movie or tv episode, but that does not mean we support the mediatype outright.

calumapplepie:
> Can you provide an example of how you're using this?

I'm still working on doing a baseline run of my collection before I break out the automated filters.  I don't do  that much in the way of the core supported media types for cross-seed but some examples of how to use this are strings like " by ", "epub+mobi" "{req}" "- Copy".

> I'm not sure what private trackers you're referring to

I'd rather not include a list of specific trackers which I use.  Clearly, they wouldn't be considered 'reputable'; but they serve my needs.  

> As for the non-supported mediatypes, we have code that ensures that searches are performed properly and we don't assume everything is a movie or tv episode, but that does not mean we support the mediatype outright.

The book mediatype includes a specific filter for removing Calibre database indexes; hence why I suspect it is, at least to some extent, supported.  This was added by you in #778.  Are only data-based searches of Calibre libraries supported, then?

In any case, this feature does not impose a significant support burden; it's merely another filter applied to titles.  It has significant utility for users who may wish to transition to more reputable trackers, cross-seeding their existing data, which may be sourced from less reputable sources.  Public trackers are, after all, explicitly supported as a source for data.

calumapplepie:
This PR isn't asking for full ebook support; it's attempting to implement a feature that I was motivated to implement because of working with ebooks.  Is this feature completely undesired?  Or is there some other issue with the PR?

zingn:
i really need this feature. thank you @calumapplepie !

mmgoodnow:
We'll add it to v7.

VannTen:
I had recently a case where extra words broke querying the title of a movie in Radarr for and old torrent, thus preventing id-based searching. (the torrent name was appending `(version.restaurée)` to the title

I fixed it by renaming the torrent file in transmission, but depending on where this filtering would occur, this could have helped for that kind of problem (those torrents are old ones from long closed trackers, this was on initial set up cross-seed with my existing library)

zakkarry:
> I had recently a case where extra words broke querying the title of a movie in Radarr for and old torrent, thus preventing id-based searching. (the torrent name was appending `(version.restaurée)` to the title
> 
> I fixed it by renaming the torrent file in transmission, but depending on where this filtering would occur, this could have helped for that kind of problem (those torrents are old ones from long closed trackers, this was on initial set up cross-seed with my existing library)

There's an entire setting called `blockList` that has multiple triggers, from name, infoHash, path, and cateogry.

you can find the usage in the cross-seed.org page https://www.cross-seed.org/docs/basics/options#blocklist

If you're talking about REMOVING the string from the search, I would highly recommend using ther ID Searching with Sonarr/Radarr integration I wrote to totally make this problem a thing of the past. Otherwise, knowing for sure that the words being removed are the proper words to get the appropriate search terms is just not really feasible.

VannTen:
(i'm not sure why you're mentionning blocklist ? I wasn't looking to block, but to match correctly)

Yeah, I use id-searching. What happened was that cross-seed queried radarr, but the extra string caused radarr to not match on the movie (I tested with "test parsing" in Radarr CF settings, removing the words did match).

That's why I said 

>  but depending on where this filtering would occur,

I'm not sure where this PR would remove the words, obviously if it's only for text based search it would not help my specific case.

PR #1064, opened 2025-10-26
Problems UI

mmgoodnow:
This gives us a good way to surface issues that would have previously intentionally crashed the application.

image


src/problems/linking.ts
```diff
@@ -0,0 +1,115 @@
...
+				path: dir,
+				problem: descriptor,
+				dev: descriptor ? null : (await stat(dir)).dev,
+			};
+		}),
+	);
+
+	for (const { problem } of linkDirStats) {
+		if (problem) problems.push(problem);
+	}
```
mmgoodnow: this feels like it could use flatMap

PR #1152, opened 2026-06-04
Fix Torrent Clients settings save validation

### LINKED ISSUE #1151: Torrent Clients settings can silently fail to save useClientTorrents
## Summary

The Torrent Clients settings page appears to fail silently when saving Torrent Client Options such as `useClientTorrents`, especially for migrated Deluge configurations that still have `torrentDir` set.

This was reported from a Deluge setup that migrated from an older `config.js` shape:

- `delugeRpcUrl` set in file config
- `useClientTorrents: false`
- `torrentDir: "/torrents"`

After migration, attempting to check `useClientTorrents` and save from the WebUI did not show a saved-settings toast, did not show a console error, and did not persist across refresh or daemon restart.

## Current findings

- `delugeRpcUrl` does appear to migrate into `torrentClients` correctly.
- The likely immediate WebUI save bug is that `/settings/clients` validates the Torrent Client Options form with a schema that includes `torrentClients`, even though that form submits only the option fields. Runtime `torrentClients` are serialized strings, while the client editor/table uses UI-shaped objects.
- A local candidate fix exists in commit `f6ee5b60` (`fix client settings form validation`), which omits `torrentClients` from the Torrent Client Options form validation and keeps the client list handled by the separate add/edit/delete flow.

## Follow-up questions

- Should migration normalize older Deluge configs that have both `torrentDir` and a Deluge client configured toward `useClientTorrents` where possible?
- Should the WebUI make it clearer that clearing `torrentDir` means leaving the field blank, rather than entering `null` as users may have done in `config.js`?
- Should the v7 migration guide call out old `delugeRpcUrl` + `torrentDir` setups and explain the expected `torrentClients` / `useClientTorrents` shape?

## Acceptance criteria

- Saving `useClientTorrents` from `/settings/clients` persists and shows the normal success toast.
- Saving an empty `torrentDir` from `/settings/clients` clears the setting as expected.
- Migrated Deluge configurations do not leave users stuck in a state where they cannot enable client torrent indexing from the WebUI.
- Documentation or migration notes clarify the expected path for older Deluge configs if code cannot fully normalize it.

mmgoodnow:
## Summary

Fixes the Torrent Clients settings page so Torrent Client Options can save independently from the client list/table state.

The options form was using `downloaderValidationSchema`, which includes `torrentClients`. That field is not submitted by the options form, and runtime `torrentClients` are serialized strings while the client editor/table works with UI-shaped objects. That mismatch could make saving options such as `useClientTorrents` fail before the settings mutation ran.

This PR omits `torrentClients` from the options form validation and keeps the client list data separate for display/add/edit/delete handling.

Fixes #1151.

## Validation

- `npm -w webui run typecheck`
- `npm -w webui run lint` (passes with existing coverage-file warnings)
- `npm run typecheck`
- `npm run test`
- `npm run build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
 		"proseWrap": "always"
 	},
 	"lint-staged": {
-		"*.{js,ts}": "eslint --cache",
-		"*.{js,json,ts,css,md}": "prettier --write"
+		"*.{js,jsx,ts,tsx}": "eslint --cache",
+		"*.{js,jsx,json,ts,tsx,css,md}": "prettier --write"
 	}
 }
--- a/webui/src/routes/settings/clients.tsx
+++ b/webui/src/routes/settings/clients.tsx
@@ -42,7 +42,7 @@ import { useSaveConfigHook } from "@/hooks/saveFormHook";
 import z from "zod";
 import { RuntimeConfig } from "@cross-seed/shared/configSchema";
 
-type ClientFormData = z.input<typeof downloaderValidationSchema>;
+type ClientOptionsFormData = z.input<typeof downloaderValidationSchema>;
 
 function TorrentClientsSettings() {
   const trpc = useTRPC();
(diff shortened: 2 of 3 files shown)


zakkarry:
@mmgoodnow I tested and found this branch fixes the issue I was facing. It allows modification of the settings from the deprecated `DelugeRpcUrl`, works as expected _with_ those settings, and reflects a proper migration from' DelugeRpcUrl' to' torrentClients' in the database's settings (`settings_json` in the `settings` table). 

I think the only remaining related issue I mentioned to you is details related to each setting and usage in the WebUI, Unraid does this in an [interesting way shown here](https://i.ibb.co/XZtKtj7y/waterfox-Vl-Z4g-Kcz0i.gif) to give a brief context and usage, which previously would have been comments above in config.js. How and what specifically to include is an entirely different conversation and is out of scope for this PR, though.

PR #1203, opened 2026-08-28
Hash-check rTorrent injects after load instead of as a load parameter

xombiemp:
## Summary
- Stop sending `d.check_hash=` as a `load.raw` parameter. Proxies such as ruTorrent httprpc strip unknown load commands, so partial matches (and `skipRecheck: false`) never started a hash check.
- After the torrent is in the client, call `recheckTorrent()` (`d.pause` then `d.check_hash` as their own RPCs) and then `resumeInjection`, matching qBittorrent. A failed hash-check RPC is logged and does not fail the inject.
- Quoted `d.directory_base.set` / `d.custom1.set` on load are unchanged; those still need the ruTorrent quoting sanitizer fix to survive httprpc.

## Test plan
- [ ] Unit tests in `cross-seed/tests/RTorrent.test.ts` (full match uses `load.raw_start` with no hash check; partial / `skipRecheck: false` use `load.raw` then `d.check_hash`; hash-check failure still returns success)
- [ ] Inject a `MATCH_PARTIAL` through ruTorrent httprpc and confirm rtorrent starts hashing (not left paused with no check)
- [ ] Confirm a full `MATCH` with `skipRecheck: true` still uses `load.raw_start` and does not hash-check

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cross-seed/src/clients/RTorrent.ts
+++ b/cross-seed/src/clients/RTorrent.ts
@@ -873,24 +873,13 @@ export default class RTorrent implements TorrentClient {
 		const retries = 5;
 		for (let i = 0; i < retries; i++) {
 			try {
-				await this.methodCallP<void>(
-					loadType,
-					[
-						"",
-						new Metafile(rawWithLibtorrentResume).encode(),
-						`d.directory_base.set="${directoryBase}"`,
-						`d.custom1.set="${TORRENT_TAG}"`,
-						`d.custom.set=addtime,${Math.round(Date.now() / 1000)}`,
-						toRecheck
-							? `d.check_hash=${meta.infoHash.toUpperCase()}`
-							: null,
-					].filter((e) => e !== null),
-				);
-				if (toRecheck) {
-					void this.resumeInjection(meta, decision, {
-						checkOnce: false,
-					});
-				}
+				await this.methodCallP<void>(loadType, [
+					"",
+					new Metafile(rawWithLibtorrentResume).encode(),
+					`d.directory_base.set="${directoryBase}"`,
+					`d.custom1.set="${TORRENT_TAG}"`,
+					`d.custom.set=addtime,${Math.round(Date.now() / 1000)}`,
+				]);
 				break;
 			} catch (e) {
 				logger.verbose({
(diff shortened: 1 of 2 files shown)


zakkarry:
qBittorrent and Deluge both operate the same way, adding the torrent in a paused/stopped state and then issuing a hash check. For Deluge, there's no need to resume it; stopped torrents can be rechecked while paused/stopped. However, qBittorrent differs slightly in its rechecking prerequisites.

If I'm not mistaken, with rTorrent we previously used load, which performed a full hash check on every injected torrent. The injected torrent was loaded by copying it from the torrent_cache to a local temp directory. This was updated at the end of v4 or sometime in early v5 to use the RPC, and I believe it's like Deluge in that you can issue an API command and send a base64-encoded .torrent for injection.

Anyway, in the RPC for rTorrent: [load.normal](https://rtorrent-docs.readthedocs.io/en/latest/cmd-ref.html?highlight=load#term-load-normal) loads the torrent in the same "raw" form but stops it, which we can then use to check the state of skipRecheck and whether it is a data-based match, and issue a d.check_hash on the already paused torrent. If the torrent is meant to skip the rechecking process, then we can instead issue the appropriate `load_raw` vs `load_raw_start` (which it appears there is already a ternary for these two options in the file of the diff of this PR in [`rTorrent.ts`](https://github.com/cross-seed/cross-seed/pull/1203/changes#diff-ffb324bd2b35937391e3eefb72cbd84477d1e0014508f915bf1b2f8f22352a91R871).
<hr>

Based on what I see at the above link to L871, there should be a check/no-recheck-when-appropriate situation already taking place. Do you have example situations of when this did not operate in the manner it should have that you can share? Perhaps some verbose logs showing/strarting at the injection point and going through the checking procedure to the resumeInjection job?

xombiemp:
The `load.raw` / `load.raw_start` split is doing what you describe. `shouldRecheck` still picks `load.raw` when a hash check is required and `load.raw_start` when it should be skipped. `load.raw` only adds the torrent stopped. With `libtorrent_resume` on the payload, rtorrent trusts that resume data, so the check itself was the extra load argument:

```
d.check_hash=<infohash>
```

That argument is valid rtorrent syntax, and a direct XML-RPC connection to rtorrent still runs it. ruTorrent's httprpc proxy, in its default sanitize mode, rebuilds `load.*` and keeps only command arguments on its allowlist (`d.directory_base.set`, `d.custom1.set`, `d.custom.set`, and a few others). `d.check_hash` is not on that list, so the proxy drops it and still adds the torrent. cross-seed sees a successful inject.

`load.normal` takes a path or URL. This path sends the torrent bytes, so it stays on `load.raw` / `load.raw_start`. After the torrent is in the client, the PR calls `d.pause` and `d.check_hash` as their own RPCs, then `resumeInjection`, which is the same sequence qBittorrent uses. httprpc forwards those methods instead of treating them as load arguments. On rtorrent 0.16.21 they are marked untrusted-safe (`rpc.mark_safe` in `command_download.cc`), along with `d.resume`.

A full `MATCH` with `skipRecheck: true` is unaffected, because that path never sent `d.check_hash`. The case this fixes is a `MATCH_PARTIAL`, or any inject with `skipRecheck: false`: the torrent is added stopped and never enters hashing.

PR #1205, opened 2026-08-29
fix(search/parsing): stop EBOOK_AND_MUSIC_RELEASE_REGEX eating the first space of book/audio titles

### LINKED ISSUE #1204: cleanBookAndAudioTitle strips the first space from BOOK/AUDIO search queries, breaking data-based book searching (v6.13.7)
### Describe the bug

`EBOOK_AND_MUSIC_RELEASE_REGEX` (added in 9c574352367dc4db0ef8dbeb0935bcae1ac4d66b, first released in v6.13.7) ends with a bare `[^a-zA-Z0-9]` alternative:

```ts
export const EBOOK_AND_MUSIC_RELEASE_REGEX =
    /(\d+k?\b|…|(Read|Narrated)?\W?By\W\w+\W*\w*\b|[^a-zA-Z0-9])/i;
```

The regex has no `g` flag, so `cleanBookAndAudioTitle()` in `src/utils.ts` replaces only the **leftmost single match**. For any title where none of the earlier alternatives match first, the leftmost match is the *first space in the title*, and exactly that one character is deleted:

- `Jane Doe` → `JaneDoe`
- `The Art of Systems Thinking` → `TheArt of Systems Thinking`
- `Widget Farming A Practical Guide` → `WidgetFarming A Practical Guide`

These mangled strings are sent as the Torznab `q=` parameter for every BOOK/AUDIO searchee on indexers with book-search caps, and typically return zero results. The later alternatives (e.g. stripping `by <author>`) also rarely get a chance to run, since the first space is almost always leftmost.

Net effect: data-based searching of book libraries (e.g. against MyAnonamouse via Prowlarr) finds essentially nothing for any multi-word title. The failure is silent — searches complete normally with 0 candidates, so it looks like an unmatched library.

### Expected behavior

Interior spaces preserved in the query. In v6.13.6 the BOOK path used `cleanTitle()` and queries were well-formed (though 6.13.6 predates the qBittorrent 5.2 login handling, so downgrading isn't a workaround for everyone).

### Verification

Running v6.13.7's actual `cleanBookAndAudioTitle` reproduces the mangling for all examples above. Removing the `|[^a-zA-Z0-9]` alternative fixes it — with that patch applied, a multi-word author query that previously returned 0 candidates (`JaneDoe`-style mangling) returns 55 with the space preserved, including the exact matching release. Still present on `master` and in `v7.0.0-22`.

### Suggested fix

Either drop the bare `[^a-zA-Z0-9]` alternative, or exclude whitespace from it (`[^a-zA-Z0-9\s]`) — and if the intent was to scrub *all* junk characters, the replace also needs the `g` flag with a space (not empty string) replacement so words don't concatenate.

### Environment

cross-seed 6.13.7 (Docker, `ghcr.io/cross-seed/cross-seed:6.13.7`), qBittorrent 5.2.3, Prowlarr Torznab indexers, `dataDirs` with `maxDataDepth: 3`, `matchMode: partial`.

kris-anderson:
Fixes #1204. Corrects book/audio search-query construction. Five commits, each reviewable on its own; this description is self-contained.

## The original bug

`EBOOK_AND_MUSIC_RELEASE_REGEX` ended with a bare `[^a-zA-Z0-9]` alternative, and `cleanBookAndAudioTitle()` applied it with a **non-global** replace substituting an empty string. So exactly one character was removed per title — and when no earlier alternative matched first, that character was the first interior space:

- `Doug Lowe` → `DougLowe`
- `The Art of Systems Thinking` → `TheArt of Systems Thinking`

Those strings went out as the Torznab `q=` for every BOOK/AUDIO searchee on a book-caps indexer, so searches returned almost nothing. It failed silently: a run completes normally with 0 candidates, which is indistinguishable from a library that simply has no cross-seeds.

The benches in #1083 didn't catch it because they ran flags `igm` and substituted a *space*, whereas the shipped code ran `/i` and substituted nothing. Under the bench settings that alternative is harmless.

## What changed

1. **Global replace, substituting a space.** The alternatives now all apply instead of just the leftmost one, and a removal can't fuse two words. The space matters: on MyAnonamouse `Abaddon s Gate` returns the same 28 results as the apostrophe form, while the fused `Abaddons Gate` returns 25.
2. **Anchored numeric alternatives.** `\d+k?\b` matched any digit run, so a global replace ate real title numbers (`Fahrenheit 451` → `Fahrenheit`). Replaced with leading index `^\s*\d+\s*[-.]?(?=\s)`, years, 5+ digit catalog/ISBN runs, and bitrates.
3. **Format tokens given real boundaries.** `[kbps]{2,4}\b` matched any two-to-four characters from {k,b,p,s} at a word ending rather than the token `kbps` — dormant under a single replace, corrupting under a global one (`Books` → `Boo`, `Maps` → `Ma`, `Jobs` → `Jo`). `flac` had no boundaries either (`Flacon` → `on`), and the extensions had a trailing `\b` but no leading one.
4. **Phrase heuristics removed** — `\bNew\b`, the bare `By` clause, and `[a\W]+(novela?|saga|series)`. Globally they eat title text: `A New Hope` → `A Hope`, `Death by Black Hole` → `Death`, `A Series of Unfortunate Events` → `of Unfortunate Events`. Surplus words are harmless because candidates are filtered on size afterwards.
5. **Remaining punctuation → space**, keeping hyphens, which appear inside real titles and which indexers ignore (`Catch-22 Joseph Heller` and `Catch 22 Joseph Heller` both return 47 results). Unicode-aware via `[^\p{L}\p{N}\s-]`, so `Union Café` keeps its accent instead of becoming `Union Caf`.
6. **Never emit an empty query.** A global replace can consume a string outright (`1984.epub` is a year plus an extension), and an empty `q=` returns the indexer's whole feed, so the function falls back to the cleansed title.

Also drops `\[.*?]` and `\(.*?\)` from the regex: `cleanseSeparators()` runs first and already strips square brackets and turns parens into spaces, so neither could ever match. `\{.*?}` stays — braces are untouched by it.

## The pipeline, as cross-seed actually runs it

`cleanBookAndAudioTitle()` is `cleanseSeparators()` → **pass 1** → **pass 2** → collapse whitespace and separator runs → non-empty fallback. Each regex pass has a live demo, with pass 2's test strings being pass 1's output so they read in order:

- **Pass 1** — `EBOOK_AND_MUSIC_RELEASE_REGEX`, flags `gi`, substituting a space: https://regex101.com/r/Ykonie/1
- **Pass 2** — `NON_ALPHANUM_KEEP_HYPHEN_REGEX`, flags `gu`, substituting a space: https://regex101.com/r/WShPBe/1

For reference, the two earlier demos from this thread: the [regex as shipped](https://regex101.com/r/fhLdZZ/1) (`/i`, empty substitution — shows the original single-space deletion) and the [first-commit state](https://regex101.com/r/PfvBph/1).

## Results

Generated by running the built function against `master`:

| input | master | this PR |
| --- | --- | --- |
| `Doug Lowe` | `DougLowe` | `Doug Lowe` |
| `01 - Inkheart - Cornelia Funke.epub` | `- Inkheart - Cornelia Funke epub` | `Inkheart - Cornelia Funke` |
| `(TTC) Jeffrey L. Kasser, Philosophy of Science.PDF` | `TTCJeffrey L Kasser, Philosophy of Science PDF` | `TTC Jeffrey L Kasser Philosophy of Science` |
| `01 Around the World in 80 Days (Read by Patrick Tull).m4b` | `Around the World in 80 Days Read by Patrick Tull m4b` | `Around the World in 80 Days Read by Patrick Tull` |
| `Fahrenheit 451 - Ray Bradbury.epub` | `Fahrenheit451 - Ray Bradbury epub` | `Fahrenheit 451 - Ray Bradbury` |
| `Catch-22 - Joseph Heller.mobi` | `Catch22 - Joseph Heller mobi` | `Catch-22 - Joseph Heller` |
| `A Series of Unfortunate Events - Lemony Snicket.epub` | `of Unfortunate Events - Lemony Snicket epub` | `A Series of Unfortunate Events - Lemony Snicket` |
| `Arcade Fire - Funeral (2004) [CD FLAC] {MRG 255}` | `ArcadeFire - Funeral 2004 {MRG 255}` | `Arcade Fire - Funeral` |
| `A Perfect Circle - 2000 - Mer De Noms (2CD) - FLAC` | `APerfect Circle - 2000 - Mer De Noms 2CD - FLAC` | `A Perfect Circle - Mer De Noms 2CD` |
| `Penguin Cafe Orchestra - Union Café {2017 Remaster} (1993) FLAC` | `PenguinCafe Orchestra - Union Café {2017 Remaster} 1993 FLAC` | `Penguin Cafe Orchestra - Union Café` |
| `Books Maps Steps Jobs Bass` | `Boo Maps Steps Jobs Bass` | `Books Maps Steps Jobs Bass` |
| `1984.epub` | `epub` | `1984 epub` |
## Verification

- Unit tests for `cleanBookAndAudioTitle` — space preservation, token stripping, punctuation-to-space, hyphen retention, non-ASCII, the truncation class, separator collapse, and the empty-query fallback. Suite passes (56), typecheck and lint clean.
- The built function was run over the full test benches from #1083 (142 music + 3085 ebook strings): no empty results, no residual punctuation, no truncated words.
- Verified live against a Prowlarr/MyAnonamouse setup: a multi-word author query that returned 0 candidates when mangled returns 55 with the space preserved.

## Known limits

- A numeric-only title (`1984`) is indistinguishable from a year and gets stripped; the fallback keeps the query non-empty rather than sending an empty one.
- Short or alphanumeric catalog groups can still leave noise (`VJCP-68559` → `VJCP`, `2CD` survives). Harmless given size filtering, but worth knowing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zakkarry:
Discussion also had in the original issue @kris-anderson opened, and for historical record's sake you can view that at #1204 

While I'm sure you're aware kris, the master branch is now pre-release v7, but your post mentions 6.13.7 - I'm not sure if this is worthy of a backport, although it is rather small and can be seen as a bug, for those who continue to use 6.x it may be beneficial, although for the sake of posterity I must reiterate that we don't fully support music and ebooks and I added this regex and the "support" for the searching merely based on users attempting to cross-seed unsupported formats regardless - and because of that spamming trackers with essentially useless and erroneous searches for minimal (if any) results.

Will be waiting on @mmgoodnow for his decision on whether this is worth backporting to 6.x and when we will be merging to the master branch and potentially cutting a pre-release as well.

zakkarry (dismissed):
looks good to me, but keep in mind this is still an unsupported media type that i just happen to have been testing to build my MaM account up for fun - didn't really need it just was coding it up for fun and to practice my regex, not to get recruitment stuff...but I ended up speaking to admins, and I know OPS has a pretty strict policy and "Download Factor" that also affects your account and can be inflated rather large with cross-seeding....so I put this together the way I though would work best, clearly I missed something with my test cases that you found.

Thank you for that.

The tests are also very nice to have, I think sometime down the road we might make more robust testing, but for the time being this and the regex101 seems to cover what we need.
<hr>
I did notice that rather than your regex stripping the numericals in front of the title, along with publishers and some other things it would be a bit more strictly author/title previously. Is there something that can be done so we search for just the author and title moreso?

kris-anderson:
Answering the author/title question — it turned out to have the same root cause as the original bug.

**Why numerals and publishers aren't being stripped today:** the regex in the code isn't global, so `cleanBookAndAudioTitle()`'s `.replace()` only ever removes the single leftmost match. Your benches ran `igm` with a space substitution, so there every alternative applied — the same flags mismatch that hid the space bug. Concretely, `01 - Inkheart - Cornelia Funke.epub` currently cleans to `- Inkheart - Cornelia Funke epub`: one token gone, everything else intact.

So yes — making the replace global is most of what you were going for. One trap: `\d+k?\b` matches any digit run, so a naive `/g` eats real title numbers (`Fahrenheit 451` → `Fahrenheit`, `Catch-22` → `Catch-`, `80 Days` → `Days`). Anchoring the numeric cases keeps them — leading index `^\s*\d+\s*[-.]?(?=\s)`, years, 5+ digit catalog/ISBN runs, bitrates `\d+k\b`.

Two related notes: `\[.*?]` and `\(.*?\)` are unreachable in this path, since `cleanseSeparators()` strips square brackets and turns parens into spaces before the regex runs — `\{.*?}` is *not*, so it stays (it's what removes `{MRG 255}`). And a global replace can consume a string outright (`1984.epub` is a year plus an extension), which would send an empty `q=` and pull the indexer's whole feed, so the function now falls back to the cleansed title.

Pushed as a second commit so the minimal space fix and this enhancement stay separately reviewable, with the OP updated and a third regex101 for the new version: https://regex101.com/r/a90oim/1. Results on your own bench strings:

- `01 - Inkheart - Cornelia Funke.epub` → `Inkheart - Cornelia Funke`
- `Arcade Fire - Funeral (2004) [CD FLAC] {MRG 255}` → `Arcade Fire - Funeral`
- `1993 - Radiohead - Pablo Honey (0777 7 81409 2 4)` → `Radiohead - Pablo Honey 0777 7 2 4`
- `01 Around the World in 80 Days (Read by Patrick Tull).m4b` → `Around the World in 80 Days`

Ran the built function over both full benches (142 music + 3085 ebook cases): no empty results, no orphaned separators, no double spaces. Known limits — a numeric-only title like `1984` is indistinguishable from a year, and short or alphanumeric catalog groups still leave noise (`VJCP-68559` → `VJCP`).

For context on how I got here: MAM is my main use case, and pointing cross-seed's data-based matching at an existing ebook/audiobook library is what surfaced this. I've recently started using RED as well, and I'm waiting for OPS to open registration again. I'm aware of fertilizer for the RED/OPS side, and for music specifically that's likely the better-suited tool — so I'm not asking you to invest further in music support on my behalf. This one just looked like a plain bug in the path that already exists.

On the backport: don't bother on my account — I'm on 6.13.7 with a local patch, and v7 is the right target. Happy to tune the anchoring differently if you'd prefer.

zakkarry:
> On the backport: don't bother on my account — I'm on 6.13.7 with a local patch, and v7 is the right target. Happy to tune the anchoring differently if you'd prefer.

I think backporting wouldn't be a big deal anyway, and v7 is still a decent ways off.

There are certain characters and strings that should be removed, you basically want just the author and title - even a hyphen is good to remove.

See the following for examples of what should be removed, the spaces before and after can be truncated fairly easy with a helper function in utils to just remove trailing and preceeding spaces after the capturing from the regex.

https://i.ibb.co/0Vr7B964/image.png

kris-anderson:
Pushed as a third commit. One thing worth flagging first, since it affects what's actually left to fix: the screenshot is the regex101 substitution panel, which shows the regex on its own — but the collapse-and-trim happens in `cleanBookAndAudioTitle()`, so three of the four boxes were already handled in the shipped function:

| input | function output before this commit |
| --- | --- |
| `01 - Inkheart - Cornelia Funke.epub` | `Inkheart - Cornelia Funke` |
| `01 Around the World in 80 Days (Read by Patrick Tull).m4b` | `Around the World in 80 Days` |
| `0310283205 Myths About Christianity.pdf` | `Myths About Christianity` |
| `(TTC) Jeffrey L. Kasser, Philosophy of Science.PDF` | `TTC Jeffrey L Kasser, Philosophy of Science` ← the real gap |

`(TTC)` survived because `cleanseSeparators()` turns parens into spaces before the regex runs, so a source tag becomes an ordinary word and nothing can distinguish it. Fixed by stripping parenthesized groups up front with the existing `ALL_PARENTHESES_REGEX` — the same treatment the `noParentheses` variant already applies further down `utils.ts`. That also removes catalog numbers and narrator credits that live in parens.

Hyphens now go too, separator and intra-word alike. I checked this doesn't cost recall on MAM before committing to it: `Catch-22 Joseph Heller` and `Catch 22 Joseph Heller` both return 47 results, `All-in-One For Dummies` and `All in One For Dummies` both return 100 — their search ignores them.

And there's now a `collapseSpacing()` helper for the leading/trailing whitespace token removal leaves behind, as you suggested. The same `.replace(ALL_SPACES_REGEX, " ").trim()` pattern appears at three other call sites in `utils.ts`; I left those alone to keep the diff scoped, but happy to fold them in if you'd like.

Results now:

- `(TTC) Jeffrey L. Kasser, Philosophy of Science.PDF` → `Jeffrey L Kasser, Philosophy of Science`
- `01 - Inkheart - Cornelia Funke.epub` → `Inkheart Cornelia Funke`
- `1993 - Radiohead - Pablo Honey (0777 7 81409 2 4)` → `Radiohead Pablo Honey`
- `A Perfect Circle - 2000 - Mer De Noms (2CD) - FLAC` → `A Perfect Circle Mer De Noms`
- Title numbers still intact: `Fahrenheit 451 Ray Bradbury`, `Around the World in 80 Days`

Across both full benches (142 music + 3085 ebook): no empty results, no residual hyphens, no spacing artifacts.

One judgement call to put in front of you rather than bury: stripping all parens also drops series and edition markers — `(Expanse vol. 3)`, `(8th Edition)`. I think that's right for cross-seed, since the query only needs to surface candidates and the size/file comparison does the actual filtering, but it's a deliberate widening and easy to restrict to a known tag list instead if you'd rather.

On the backport — if it's genuinely low-cost on your side, I'll take you up on it and open the equivalent PR against `v6`. Selfishly it lets me drop the local patch I'm running, and it's the same small change.

kris-anderson:
Heads-up on something the global replace exposed — one part fixed in a fourth commit, two that need your call since they're design decisions rather than clear bugs.

**Fixed: `[kbps]{2,4}\b` was truncating ordinary words.** It matches any two-to-four characters drawn from {k,b,p,s} at a word ending, rather than the token `kbps`. Non-globally it fired at most once per title and usually lost the leftmost-match race, so it was effectively dormant. Global, it fires everywhere:

| word | becomes |
| --- | --- |
| Books | Boo |
| Maps | Ma |
| Jobs | Jo |
| Steps | Ste |
| Bass | Ba |

`Books` losing its tail on a books tracker is the worst case. `flac` had no boundaries either (`Flacon` → `on`), and the extension tokens had a trailing `\b` but no leading one. Fixed with `\b\d+\s?kbps\b|\bkbps\b` (which also swallows a preceding bitrate) plus boundaries on `flac` and every extension, with regression tests. Verified across both benches from #1083: no truncations, no empties.

**Two more that are yours to decide, because they're intentional heuristics rather than mistakes:**

`\bNew\b` and `(Read|Narrated)?\W?By\W\w+\W*\w*\b` were tolerable when they fired once. Globally they eat real title text:

| title | query |
| --- | --- |
| `A New Hope - George Lucas` | `A Hope George Lucas` |
| `The New Jim Crow - Michelle Alexander` | `The Jim Crow Michelle Alexander` |
| `Death by Black Hole - Neil deGrasse Tyson` | `Death` |
| `Gone by Midnight - Candice Fox` | `Gone Fox` |

On my own library that's about 2% of searchees affected, so it's not catastrophic, but `Death by Black Hole` collapsing to `Death` is a bad query.

The cleanest fix I can see keeps your intent intact: **split into two regexes** — one global for the unambiguous tokens (extensions, `kbps`, years, catalog numbers, braces), and one single-pass or end-anchored for the heuristics (`By` credits, `New`, `novel/saga/series`). Author credits and release junk almost always trail, so anchoring the `By` clause with `\s*$` would keep it doing its job without reaching into titles. Alternatively `\bNew\b` could require adjacent release junk rather than standing alone.

Happy to implement whichever you prefer, or to drop the `/g` change entirely if you'd rather keep the single-pass semantics and accept that most tokens survive — that's a legitimate trade and it's your call. Right now the PR is strictly better than `master` either way, since `master` has both the space bug and these latent flaws.

zakkarry:
I would get rid of any non-alphanumeric characters - including commas, periods, and parentheses...searches for the cross-seeds won't need any of that (exception being some hyphens, but you can probably just keep all the hyphens unless (if possible) it is the first couple of characters like `"01 -"`. Even the `"(Narrated/Read by John Doe)"` would be fine without any non-alphanumeric characters like parentheses around it, as there would be matches across the entire title and no potentially erroneous or non-matching "punctuation"  to trip absolutely anything up.

When you search for a book on MaM, for example, do you put all the parentheses, initials' periods, and hyphens that reside outside of words in the title or author's last name type of thing? Because it returns a larger number of results - yes, but that's not an issue and cross-seed can quickly filter through those and match them based on size, as you mentioned.

kris-anderson:
Done — implemented as a fifth commit, with two measured refinements to your spec.

**Punctuation now maps to a space, not to nothing.** That distinction turned out to matter: on MAM, `Abaddon's Gate` and `Abaddon s Gate` both return 28 results, but the fused `Abaddons Gate` returns 25. Deleting rather than spacing is the same failure mode that started this thread, so the replacement is a space.

**The character class is Unicode, not ASCII.** `[^a-zA-Z0-9]` would turn `Union Café` into `Union Caf` and gut non-Latin titles entirely. Using `[^\p{L}\p{N}\s-]` keeps them — consistent with `NON_UNICODE_ALPHANUM_REGEX`, which already uses `\p{L}\p{N}` elsewhere in the file. About 1% of the bench titles are affected, so it's quiet but real.

**Hyphens kept, leading index still stripped** — verified they don't cost recall: `Catch-22 Joseph Heller` and `Catch 22 Joseph Heller` both return 47 results, `All-in-One For Dummies` and `All in One For Dummies` both return 100.

**And I took your "surplus words are harmless" reasoning to its conclusion** and removed the phrase heuristics entirely, rather than tightening them — `\bNew\b`, the bare `By` clause, and `[a\W]+(novela?|saga|series)`. Under a global replace all three eat title text:

| title | before |
| --- | --- |
| `A New Hope` | `A Hope` |
| `Death by Black Hole` | `Death` |
| `A Series of Unfortunate Events` | `of Unfortunate Events` |

Since candidates get filtered on size anyway, leaving `Read by John Doe` as plain words is strictly safer than trying to detect and remove it — which is what you were saying. Say the word if you'd rather keep any of them.

Results now:

- `01 - Inkheart - Cornelia Funke.epub` → `Inkheart - Cornelia Funke`
- `(TTC) Jeffrey L. Kasser, Philosophy of Science.PDF` → `TTC Jeffrey L Kasser Philosophy of Science`
- `Penguin Cafe Orchestra - Union Café {2017 Remaster} (1993) FLAC` → `Penguin Cafe Orchestra - Union Café`
- `Death by Black Hole - Neil deGrasse Tyson.epub` → `Death by Black Hole - Neil deGrasse Tyson`

Full benches from #1083 (142 music + 3085 ebook): no empty results, no residual punctuation. 56 unit tests, typecheck and lint clean.

Worth noting where this landed: your original `[^a-zA-Z0-9]` alternative was the right instinct all along. It just needed the `g` flag, a space replacement, and to spare whitespace and hyphens — the whole thread came out of that one alternative deleting a single space.

zakkarry:
Just for the sake of posterity, I'm going to post my "tell all" about the creation of and testing with my REGEX and the results/experience I had. I wont go into detail here, but my reddit post tells all.

https://www.reddit.com/r/trackers/comments/1pp27h7/crossseed_for_books_and_music/nutzsks/

@kris-anderson  if you could check it out, I think you'll get an idea of what the goal and workflow was - I got a lot of results.

kris-anderson:
Read the whole thing — thanks for writing it up. It answers the question I'd been sitting on, which was why your results and mine diverged so sharply.

Mine is the organic case, and it's probably a useful counterpoint for the record. I pointed data-based matching at an existing personal library rather than a corpus built for overlap: ~1,400 searchees across ebooks, audiobooks and kids' books, MAM as the only book-capable indexer, `matchMode: partial`, 120s delay. One full run got through 200 queries and 45 candidate inspections. **Zero cross-seeds.** The rejection breakdown explains it:

- **Calibre-managed ebooks** — Calibre rewrites files, so total sizes drift and the file-level comparison fails. This was the bulk of the near-misses.
- **Audiobooks** — Audible/Libation output, which exists on no tracker in that form.
- **Comics** — mine are single usenet issues; MAM stocks complete series collections, so candidates came back 88–252× larger. I dropped that directory once I saw the ratios.
- **45 hits were `infoHash` matches** — MAM torrents I already seed. Nothing to gain, but proof the queries were landing on the right books.

Put next to yours it reads almost like a controlled comparison: engineered overlap — scraping ~6,000 IPT uploads from people who also upload to MAM — produced ~3,000 matches; an organic, tool-managed library produced ~0. Which is your point about exact naming searches being where the results actually are, and the broad title path being marginal.

I took the practical warnings on board too. Given the "not a suitable target for cross-seeding at scale" position and the satisfied-seeding cap — I'm User class, so 50 rather than your VIP 150 — I'm keeping this deliberately small: one indexer, 200 queries a day, 120s delay, no plans to grow it. Your point about a few extra BP not being worth losing API access is well taken.

On music I'm not planning to point cross-seed at it at all. If OPS ever reopens registration I'd set up fertilizer, which is built for that pair — with only RED there's no counterpart tracker for it to work against anyway. nemorosa is new to me, so thanks to whoever surfaced it in the thread.

One suggestion that might save others the trip: a line in the data-based matching docs saying results come mainly when your files came from the same release as the tracker's copy, and pointing RED/OPS users at fertilizer or nemorosa. I'd have calibrated my expectations very differently had I read that first, and MAM would have seen a few hundred fewer queries from me. Happy to open a small docs PR if that's useful.

None of this changes the code, I think: the first-space bug and the truncation bugs were real regardless of how marginal the feature is, and the tests should keep them from recurring. Nothing further from me unless you'd rather keep any of the heuristics I pulled.

zakkarry:
> On music I'm not planning to point cross-seed at it at all. If OPS ever reopens registration I'd set up fertilizer, which is built for that pair — with only RED there's no counterpart tracker for it to work against anyway. nemorosa is new to me, so thanks to whoever surfaced it in the thread.
> 
> One suggestion that might save others the trip: a line in the data-based matching docs saying results come mainly when your files came from the same release as the tracker's copy, and pointing RED/OPS users at fertilizer or nemorosa. I'd have calibrated my expectations very differently had I read that first, and MAM would have seen a few hundred fewer queries from me. Happy to open a small docs PR if that's useful.

I've taken over fertilizer and am the only active person in the organization for its repo, although i don't have a whole lot of time for another few weeks or maybe a month or 2 - dont know yet. cross-seed has always been my main priority next to writing documentation for things like libtorrent and deluge support/plugins/bugfixes.

I'm going to try and get some of those open and bordering on stale PR's on fertilizer going, not sure if you're at all versed in python and whether you have used fertilizer enough to have glanced at its codebase, but I think there's some room for improvement there too - but would love to see v7 through before really diving in.

<hr>

As far as this regex/parsing fix goes, are you satisfied with the search strings? Can you update the regex101's for testing parsing as cross-seed would and cover the changes made so the OP is enough for someone to read and know whats going on (if you already have, then feel free to ignore me). 

I talked to Q and backporting seems like a wise decision, considering v7 is still pre-release and a slow work in progress over this summer. If you want to target the branch v6 and implement the same regex/logic/etc, there's no tests in v6 really over parsing IIRC, so you might not have to bother with that, but we can cut a release for v6.13.8, which might be the last v6 we release, but for those who want to keep using v6 during pre-release and early versions of v7, then v6 should at least have worthwhile and accurate parsing for searching.

kris-anderson:
Both done. The OP is rewritten to stand on its own — the bug, each change and why, the pipeline, a before/after table generated by running the built function against `master`, and the known limits. The regex101s are now two demos matching what the code actually does: [pass 1](https://regex101.com/r/Ykonie/1) is the token regex (`gi`, substituting a space), [pass 2](https://regex101.com/r/WShPBe/1) is the punctuation pass (`gu`), and pass 2's test strings are pass 1's real output so they read in order. The earlier links are still in the OP for the history.

On whether I'm satisfied with the search strings — yes, with two limits I've documented rather than papered over. A numeric-only title like `1984` is indistinguishable from a year and gets stripped, with the fallback keeping the query non-empty; and short or alphanumeric catalog groups can still leave noise (`VJCP-68559` → `VJCP`, `2CD` survives). Both feel acceptable now that surplus words are explicitly fine.

Backport is up as #1213, targeting `v6`, squashed into one commit. I did add the parsing tests — v6 already has the vitest harness, just no coverage there — and the full suite is green under the Node version in `.nvmrc` (31 tests).

On fertilizer — I'm comfortable in Python and would genuinely like to help, but I don't think I'd be much use until I'm actually using it. Reading a codebase without running the thing tends to produce shallow patches. So let me circle back once I've got it going and see whether you still want another pair of hands then.

Worth being upfront about where I'm coming from, though: this is all quite new to me. I've been on usenet since the 90s but only started with torrents about a month ago. Books were the entry point, then it became clear music would benefit too, which is why I joined RED. I've no plans to use torrents for movies or TV, so I doubt I'll be much long-term help on cross-seed itself — but fertilizer sits squarely in the part I do care about, so that's where I'd expect to be useful, once I'm on OPS.

zakkarry:
> Both done. The OP is rewritten to stand on its own......The earlier links are still in the OP for the history.
> 
> On whether I'm satisfied with the search strings — yes.....Both feel acceptable now that surplus words are explicitly fine.
> 
> Backport is up as #1213, targeting `v6`, squashed into one commit. I did add the parsing tests — v6 already has the vitest /harness, just no coverage there — and the full suite is green under the Node version in `.nvmrc` (31 tests).

Yes, molecularkayak from fertilizer and myself added vitest with the expectation it would be used in the future, especially when we bump major versions. As you can see, `v6` tests are pretty bare by comparison to `master`

Since you've already written them, I don't see why they shouldn't be included.
<hr>

> On fertilizer ....... let me circle back once I've got it going and see whether you still want another pair of hands then.

No problem at all, I was just putting feelers out there since you seem active and competent, but open source development (for me at least) is a hobby/practice that I do for fun (for lack of a better word), so feel free to spend your time and energy wherever you choose, I just prefer to work with someone than pick the project back up after this Summer vacation and be the sole active developer making all the decisions (and reviewing my own code :P)
<hr>

> Worth being upfront about where I'm coming from, though: this is all quite new to me. I've been on usenet since the 90s but only started with torrents about a month ago. Books were the entry point, then it became clear music would benefit too, which is why I joined RED. I've no plans to use torrents for movies or TV, so I doubt I'll be much long-term help on cross-seed itself — but fertilizer sits squarely in the part I do care about, so that's where I'd expect to be useful, once I'm on OPS.

All that is perfectly fine; while we have the primary target of movies and TV with cross-seed, as that was where it all started and has the widest reach. The additions for music and ebooks were (as I mentioned) to prevent users who adamantly attempted ebook/audiobook/music searches via cross-seed whilst disregarding our warnings in the documentation that it was not officially supported by us and could be consequential for user accounts if done at scale. 

At the time, I was still using crops (which forked into fertilizer) but wanted those using cross-seed for that purpose to at least produce relatively "sane" strings from parsing filenames, with some library software support via the calibre "stripping". 
<hr>

Regex is one of those tricky beasts. Capturing all the naming schemes, which vary drastically with the personal preferences of users, indexers, trackers, and the few release groups that exist for books but definitely exist more primarily in music with older scene CD/concert rips, and that doesn't even take into account the naming of the books and albums themselves across foreign languages and character sets. Producing coverage for all of those, and combinations of them, was sort of challenging for me. I did the best I could at the time, and haven't really thought to revisit it until you and I crossed paths here.

<hr>

I truly appreciate your work here, the details and coverage tests, and working through this with you throughout this last week was a great experience, despite it being entirely aimed at you fixing my oversights and mistakes :)


I think for now, I'm ready to give the OK to merge. However, I'd like to give @mmgoodnow or maybe @ShanaryS some time to return and share any thoughts they may have, if they have any.

PR #1212, opened 2026-09-03
fix(client/rtorrent) start injected torrents instead of only resuming them

jaredledvina:
While working on https://github.com/cross-seed/cross-seed/pull/1211 I kept facing this issue as well. In my setup, I have recheck turned on at the moment. Which causes the injected torrents to go into a stopped state after checking. As such, the existing resume logic doesn't work, they need to be started. So, this PR adds d.start support.

This PR overlaps #1201, which switches the same check to `d.state`. That change is
needed but not sufficient on its own: without `d.start` the torrent is retried
until `getResumeStopTime()` and still never starts.

In addition to the included tests, I'm also running this patch and so far it's working perfectly.

### DIFF
--- a/cross-seed/src/clients/RTorrent.ts
+++ b/cross-seed/src/clients/RTorrent.ts
@@ -203,6 +203,7 @@ export default class RTorrent implements TorrentClient {
 				bytesLeft: number;
 				hashing: 0 | 1 | 2 | 3;
 				isMultiFile: boolean;
+				isStarted: boolean;
 				isActive: boolean;
 			},
 			"FAILURE" | "TORRENT_NOT_COMPLETE" | "NOT_FOUND"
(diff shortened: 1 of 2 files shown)


zakkarry:
I've noticed that I, as well, have torrents that get injected into my rTorrent instance and end up at 100% after several runs of the inject job but never get started.

I periodically find only a few and start them manually, but this seems like a more appropriate process. 

Unfortunately, though, on our team we have a sort of distribution of clients per developer, since each of us uses, has written, and has worked on different clients specifically. 

@mmgoodnow - wrote this originally about 5 years ago now, with rTorrent support, and still uses rTorrent exclusively (as far as I know) - so he'd need to review this and have the final say on both the implementation and the codebase.

Myself and @ShanaryS are the Deluge and qBittorent guys, respectively (although we don't have anyone who is in charge of the lesser-used transmission client, but @mmgoodnow also has final say via review/approval prior to merging for our client changes as well, being the head/original author.)

jaredledvina:
Okay cool, sounds good to me! Happy to help improve things here best I can. I've been running both PR's for a few days now and so far it's been good. Thanks a lot for the project too, very handy!

